PR #90, opened 2021-06-22
transaction builder

collinc97:


rust/account/src/private_key.rs
```diff
@@ -24,9 +24,9 @@ use snarkvm_dpc::{
 use rand::{CryptoRng, Rng};
 use std::{fmt, str::FromStr};
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
howardwu, line 27: Is this necessary?

rust/transaction/src/transaction_builder.rs
```diff
@@ -0,0 +1,215 @@
...
+                .push(TransactionError::DuplicateArgument(transaction_kernel_string));
+        }
+        self
+    }
+
+    ///
+    /// Returns a new transaction builder with the added vector of old_death_program_proofs
+    /// Otherwise, logs a `TransactionError`.
+    ///
+    pub fn add_old_death_program_proofs(self, old_death_program_proofs: Vec<PrivateProgramInput>) -> Self {
```
howardwu, line 78: The `add_*` part of each of the below methods is likely not needed (it is implied)

rust/account/src/private_key.rs
collinc97: removed

rust/transaction/src/transaction_builder.rs
collinc97: renamed old_death_program_proof -> input_proof
renamed new_birth_program_proof -> output_proof

Issue #75, opened 2021-05-25
[Proposal] Testnet2 Record Epic


collinc97:
## 💥 Proposal

Requirements to add testnet2 record support to aleo sdk.

Tracking issue checklist:
- [x] implement datatypes
  - [x] record
  - [x] record ciphertext
- [x] implement wasm_bindgen methods
  - [x] record
  - [x] record ciphertext
- [ ] implement CLI support
  - [ ] record
  - [ ] record ciphertext
- [x] call JS methods in website frontend
  - [x] record
  - [x] record ciphertext

collinc97:
Closed by #144

PR #144, opened 2021-10-19
Add RecordCiphertext to wasm

amousa11:
Add RecordCiphertext to wasm, enabling encryption/decryption of records with a view key string


collinc97:
adds progress to #75

PR #164, opened 2021-11-02
update wasm tests with bech32 encoded objects

collinc97:


wasm/src/transaction/transaction.rs
```diff
@@ -57,25 +57,19 @@ impl Transaction {
     pub fn transaction_id(&self) -> String {
         self.transaction.transaction_id().to_string()
     }
-
```
amousa11, line 60: Do we still expose all the properties of transaction as a string here? We might be missing some since this hasn't been updated in a while.

amousa11 (approved):

wasm/src/transaction/transaction.rs
collinc97: This is a good idea. Implementing in a separate pr.

PR #173, opened 2021-11-05
Add record ciphertext decryption to cli

collinc97:
To decrypt a record ciphertext and view its contents run:
```
aleo record from --ciphertext <ciphertext> --viewkey <viewkey
```


collinc97:
deprecated by testnet3

PR #270, opened 2022-04-15
Deterministic account derivation with seed

piotr-roslaniec:
## Motivation

With these changes, WASM-bindings users can take advantage of RNG bundled with SDK to generate accounts deterministically. 

## Test Plan

Tested by modifying the existing test suite.

## Related PRs

(No related PRs)


wasm/src/account/account.rs
```diff
@@ -47,6 +47,18 @@ impl Account {
         }
     }
 
+    #[wasm_bindgen]
+    pub fn from_seed(seed: &[u8]) -> Self {
```
collinc97, line 51: This method can be shortened. In addition, you should use `ChaChaRng`.
```rust
AccountNative::new(&mut ChaChaRng::seed_from_u64(seed))
```
Please see our implementation in the CLI of this repository for an example.

collinc97 (changes requested):
Thanks for the PR!
It is close to what we want but with a few changes.

howardwu:
@piotr-roslaniec checking in here. Is this PR still relevant?

piotr-roslaniec:
@howardwu Yes, I still need a way to generate `Account` (or `PrivateKey`) in a deterministic way.
I will attempt to rebase this PR and adjust changes to `testnet3`.

PR #304, opened 2022-08-16
approve and unapprove functionality added

evanmarshall:
## Motivation

The ERC-20 standard supports the idea of "approvals" or allowances. Approvals enable another address to spend within a provided limit. This will be used for swap type programs that enable a program to spend a balance. 

Swaps in ETH work by an owner of a token to approve a program to spend its balance. The swap then spends the balance of two accounts it's been approved to spend from.

## Test Plan

Example usage:
1. Setup: `aleo run mint [address1] 420u64` (Ensure that the package.json reflects [address1])
1. `aleo run approve_public [address1] [address2] 69u64`
1. Change the package.json to use the private/public key for address2
1. `aleo run transfer_from_public [address1] [address3] 69u64`


examples/token/main.aleo
```diff
@@ -17,6 +21,76 @@ record token:
...
+
+finalize approve_public:
+    input r0 as address.public; // approver
+    input r1 as address.public; // spender
+    input r2 as u64.public; // increase in amount spender is allowed to withdraw from approver
+
+    hash.ped64 r0 into r3; // hash approver into field
+    hash.ped64 r1 into r4; // hash spender into field
+
+    mul r3 r4 into r5; // multiply approver f by spender f
```
evanmarshall, line 41: Is this a decent way of combining fields into a unique hash? We need the mapping of allowances to be derived from both the owner and spender addresses. Our approach is to hash each one using `hash.ped64` and then multiply the fields together.

examples/token/main.aleo
```diff
@@ -17,6 +21,76 @@ record token:
...
+    input r2 as u64.public; // decrease in amount spender is allowed to withdraw from approver
+
+    hash.ped64 r0 into r3; // hash approver into field
+    hash.ped64 r1 into r4; // hash spender into field
+
+    mul r3 r4 into r5; // multiply approver field by spender field
+
+    // if approvals for approver field times spender field exists, the approved amount is decreased.
+    // otherwise, the approved allowance is created.
+    decrement approvals[r5] by r2;
```
evanmarshall, line 68: Can we access the value of a mapping? It would be nice if the decrease to the approvals didn't underflow and instead was reset to 0 in the case that we're trying to decrease it by too much.

We tried:
```
gt approvals[r5] r2 into r6;
ternary r6 r2 approvals[r5] into r7;
decrement approvals[r5] by r7;
```

examples/token/main.aleo
```diff
@@ -17,6 +21,76 @@ record token:
...
+
+/* Transfer From */
+
+function transfer_from_public:
+    input r0 as address.public; // from the approver
+    input r1 as address.public; // to the receiver
+    input r2 as u64.public; // amount to transfer
+
+    finalize self.caller r0 r1 r2;
```
evanmarshall, line 77: When another program calls this, will `self.caller` reflect the `program_id` or the signer address in the `package.json`

examples/token/main.aleo
fulltimemike, line 68: One more thing -- when we tried decrementing `approvals` by more than the amount allotted, no error was thrown and the transaction executed successfully. Without a way to check what the value was before and after, it's hard to say what happened. However, we'd expect a transaction that caused `approvals` to underflow to fail.

examples/token/main.aleo
fulltimemike: Something we realized as we were implementing -- this approach won't work in the real world because field multiplication is associative. The addresses 1 & 2 are hashed into fields 1 & 2. Field 1 x Field 2 is equivalent to Field 2 x Field 1. In this case, address 1 approving address 2 would be the same entry as address 2 approving address 1, leading to an easy attack vector for setting allowances for any address.

We still need some way to derive a hash from both addresses in a specified order. We could add Field 1 to the product result, but any guidance on hashing with the addresses would be appreciated.

acoglio: Perhaps you could try defining an `interface` type that consists of two addresses (approver and spender), and then form and hash values of that type. I think that the components of an interface type are serialized in a fixed order before being hashed, and so they shouldn't "commute" as field multiplication does.

fulltimemike: @acoglio, I've updated the hashing to use the interface type -- it built and the functions executed 👍 Thanks for the suggestion

iamalwaysuncomfortable:
It makes me sad to close this, but it will live on in the history

PR #341, opened 2022-09-02
Print transaction ID of the deployment

IAvecilla:
## Motivation

The user of the Aleo CLI tool needs to know the transaction ID associated with the deployment of their programs. This PR includes a print of the transaction id of the deployment given by the deployment response.


IAvecilla:
These changes are deprecated with the new versions of the Aleo CLI, snarkOS, and snarkVM. If necessary, these changes will be introduced again in the future.

PR #361, opened 2022-09-15
Generate private key deterministically from seed

evanmarshall:
## Motivation

Wallets will need to be able to generate private keys deterministically from a given seed. Deterministic key generation is necessary for mnemonics to generate many keys from a single seed phrase. This PR is meant to replace/update [this existing PR ](https://github.com/AleoHQ/aleo/pull/270)

## Test Plan

1. `wasm-pack test` runs & passes.
2. This has been published as an [npm package](https://www.npmjs.com/package/@demox-labs/aleo-sdk). I'm currently using it to build a chrome extension wallet. 

## Related PRs

This should replace [this existing PR ](https://github.com/AleoHQ/aleo/pull/270)

### DIFF
--- a/wasm/src/account/private_key.rs
+++ b/wasm/src/account/private_key.rs
@@ -17,7 +17,10 @@
 use aleo_account::{Address, PrivateKey as PrivateKeyNative, ViewKey};
 
 use rand::{rngs::StdRng, SeedableRng};
-use std::{convert::TryFrom, str::FromStr};
+use std::{
+    convert::{TryFrom, TryInto},
+    str::FromStr,
+};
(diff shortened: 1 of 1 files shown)


wasm/src/account/private_key.rs
```diff
@@ -44,6 +47,17 @@ impl PrivateKey {
         Self { private_key }
     }
 
+    #[wasm_bindgen]
+    pub fn from_seed(seed: &[u8]) -> Self {
```
evanmarshall, line 51: In #270 , there was a comment about turning this into a u64. That wouldn't be secure. This way provides up to 32 bytes of entropy for the seed (instead of 8).

Issue #364, opened 2022-09-19
[Proposal] Deploy command refactor


jrchatruc:
## 💥 Proposal

The `aleo deploy` command currently delegates most of its behaviour to `snarkVM`, including broadcasting the transaction to the blockchain. This is not ideal, as the VM should only be concerned with the crafting of the transaction (i.e. the execution/proof generation), but not the broadcasting aspect; instead, it should return the `Deployment`, which the CLI then converts into a transaction and broadcasts.

This also allow us to remove the custom `deploy` node endpoint, which currently just (almost) duplicates the `broadcast` logic.

iamalwaysuncomfortable:
Thanks for this suggestion! An `aleo deploy` command has since been added which references the Aleo `rust sdk`. Thank you for the forethought in this matter.

PR #373, opened 2022-10-04
[Feature] Program Deploy Command

ilitteri:
## Motivation

This PR addresses #364.

## Additional Changes

- The Transaction ID is printed.
- The record commitment is printed if there is one (or more) as output.


howardwu:
This PR needs rebasing

Issue #356, opened 2022-09-08
[Feature] Execute program command


jrchatruc:
## 🚀 Feature

Add an `aleo execute` command and equivalent WASM functionality to execute a program's function and broadcast the resulting transaction on-chain.

## Motivation

Both Aleo developers and users need a way to execute their programs on-chain. For wallets and any other applications running on top of Aleo to work, they need a way to execute functions and publish them on the blockchain.

This is also necessary for the explorer to give users the ability to execute program functions.

## Implementation

As opposed to the `run` feature, this will require accessing on-chain state (i.e. records). When a program receives a record as input, the `execute` command (or wasm function) will expect an on-chain record *commitment*, which it will fetch from a node, decrypt using the user's view key and then use for the execution.

As an example, if there's a `hello` function with one record as input, and the CLI user has a record on chain with commitment `3892363209129015695237595996230595282638607490815872341685075479616088416491field`, they would run (on the CLI)

```
aleo execute hello 3892363209129015695237595996230595282638607490815872341685075479616088416491field
```

iamalwaysuncomfortable:
This has been implemented! `aleo execute` now executes programs and posts them on chain, thank you for this detailed issue!

PR #374, opened 2022-10-04
[Feature] Program Execute Command

ilitteri:
## Motivation

This PR addresses #356


cli/commands/execute.rs
```diff
@@ -0,0 +1,136 @@
...
+
+        // Derive the view key from the private key
+        let view_key = ViewKey::try_from(private_key)?;
+
+        // TODO: Find a better way to filter commitments from the program execution inputs.
+        let mut inputs: Vec<Value<Network>> = Vec::new();
+        self.inputs.into_iter().try_for_each(|input| {
+            if input.to_string().ends_with("field") {
+                let ciphertext: Record<Network, Ciphertext<Network>> =
+                    ureq::get(&format!("http://localhost/testnet3/ciphertext/{input}"))
```
howardwu: Can we change these URLs from `localhost` to an opinionated hosted endpoint?

howardwu:
This PR needs rebasing

cli/commands/execute.rs
ilitteri: #338 fixes this issue for deployments when it's merged I'll make the same changes for executions.

cli/commands/execute.rs
```diff
@@ -0,0 +1,127 @@
...
+                let record = ciphertext.decrypt(&view_key)?;
+                inputs.push(Value::Record(record));
+            } else {
+                inputs.push(input);
+            }
+            Ok::<(), Error>(())
+        })?;
+
+        // Find an unspent ciphertext to use later for the fee.
+        let ciphertext = ureq::post("http://localhost/testnet3/ciphertexts/unspent")
```
evanmarshall, line 75: I think this should be: `/records/unspent`. Should it also be more careful to ensure that we're spending the correct records? Like if we have a record representing other data, we may not necessarily want to spend it for this transaction. We'd want to find a record with no other data fields.

PR #377, opened 2022-10-06
Update snarkVM to revision `ea14990`

howardwu:
## Motivation

Update snarkVM to revision `ea14990`


raychu86 (approved):

howardwu:
Clippy failure is from a known nightly Rust bug (fix is already staged for next nightly released)

PR #389, opened 2022-10-19
Record wasm compatible

IAvecilla:
## Motivation

To avoid creating the same struct in the wallet and decoding and encoding a record multiple times, we need to make the Record struct available to use with WASM.


IAvecilla:
Close this PR momentaneously to test it properly in our side and include it here in the future

PR #467, opened 2023-02-07
[Feature] Add serial number function to aleo wasm

fulltimemike:
## Motivation

As a wallet developer, I would like to be able to determine which aleo records have been spent, in order to show customers a comprehensive view of their aleo network activity. Once a customer has collected and decrypted all of their records, we need to be able to determine which of those records have been spent. We can do this by computing the serial numbers of those records, and then scanning the blockchain for them. In order for the wallet and customers do this, the `serial_number` function has to be exposed in the wasm.

## Test Plan

I added a wasm test for computing the serial number given a private key string, a program id string ("token.aleo"), and a record name string ("token"). Additionally, I manually verified in the console that calling the serial number function returned the an expected serial number value.

```
import {useAleoWASM} from "../../aleo-wasm-hook";

export const example = () => {
  const aleo = useAleoWASM();
  const sampleRecord = '{\
    owner: aleo1jzmty94us0aahq9qlwzxs7zrnfaudnsfxnc57z87fdygynwsh5ysejwe7g.private,\
    gates: 1099999999999989u64.private,\
    _nonce: 2771765477105483056707957153205874478873922763747149066952146036032579289594group.public\
  }';
  const samplePK = 'APrivateKey1zkpCmsBYqVXXCgk4j1UPw4Jeyva8fRsdS8LBiJuQt1QxQoV';

  const record = aleo.RecordPlaintext.fromString(sampleRecord);
  console.log(record.serialNumber(samplePK, "credits.aleo", "credits");
}
```

## Related PRs

(Link your related PRs here)


wasm/src/record/record_plaintext.rs
```diff
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with the Aleo library. If not, see <https://www.gnu.org/licenses/>.
 
-use crate::types::RecordPlaintextNative;
+use crate::types::{IdentiferNative, PrivateKeyNative, ProgramIDNative, RecordPlaintextNative};
```
iamalwaysuncomfortable:
```suggestion
use crate::{account::PrivateKey, types::{IdentiferNative, ProgramIDNative, RecordPlaintextNative}};
```

wasm/src/record/record_plaintext.rs
```diff
@@ -43,6 +43,17 @@ impl RecordPlaintext {
     pub fn gates(&self) -> u64 {
         ***self.0.gates()
     }
+
+    #[wasm_bindgen(js_name = serialNumber)]
+    pub fn serial_number(&self, private_key: &str, program_id: &str, record_name: &str) -> String {
```
iamalwaysuncomfortable: The private key optimally shouldn't be passed in as a string as this implies that on the javascript side, the private key is stored in JS memory as a string which is high-risk if we're assuming systems might be compromised. I'd suggest passing this in as the wasm-type. 

```suggestion
    pub fn serial_number_string(&self, private_key: PrivateKey, program_id: &str, record_name: &str) -> String {
```

iamalwaysuncomfortable: Ultimately this should likely have a path for private key in ciphertext form wherein the private key ciphertext is decrypted within the function, but we can address this as further iteration.

wasm/src/record/record_plaintext.rs
```diff
@@ -43,6 +43,17 @@ impl RecordPlaintext {
...
+        let parsed_private_key = PrivateKeyNative::from_str(private_key).unwrap();
+        let parsed_program_id = ProgramIDNative::from_str(program_id).unwrap();
+        let record_identifier = IdentiferNative::from_str(record_name).unwrap();
+        let commitment = self.to_commitment(&parsed_program_id, &record_identifier).unwrap();
+
+        let serial_number = RecordPlaintextNative::serial_number(parsed_private_key, commitment).unwrap();
+        serial_number.to_string()
```
iamalwaysuncomfortable:
```suggestion
        let parsed_program_id = ProgramIDNative::from_str(program_id).unwrap();
        let record_identifier = IdentiferNative::from_str(record_name).unwrap();
        let commitment = self.to_commitment(&parsed_program_id, &record_identifier).unwrap();

        let serial_number = RecordPlaintextNative::serial_number(private_key, commitment).unwrap();
        serial_number.to_string()
```

iamalwaysuncomfortable: Also a question - how is your application identifying spent records on the javascript/typescript side? Are you string matching against data being scanned from the API?

iamalwaysuncomfortable (changes requested):
The private key optimally shouldn't be passed in as a string, otherwise this is workable.

wasm/src/record/record_plaintext.rs
fulltimemike: Great question -- yes, we will need match serial number strings that we generate against serial number strings contained in the transaction outputs on the blockchain. We'll do that through our RPC's API, and we can either request all serial numbers on chain and then string match directly in the application, or we can send a list of specific serial number strings to the RPC to scan on our behalf.

iamalwaysuncomfortable: That is a reasonable way to do it given the block explorer exports those elements as plaintext. I am still thinking over whether or not this should have a newtype abstraction over it in the API. 

I think, for the purposes of future-proofing, this should probably be renamed to serialNumberString or a related name that indicates this is a string output instead of its own object.

wasm/src/record/record_plaintext.rs
```diff
@@ -43,6 +43,17 @@ impl RecordPlaintext {
     pub fn gates(&self) -> u64 {
         ***self.0.gates()
     }
+
+    #[wasm_bindgen(js_name = serialNumber)]
```
iamalwaysuncomfortable:
```suggestion
    #[wasm_bindgen(js_name = serialNumberString)]
```

wasm/src/record/record_plaintext.rs
```diff
@@ -92,6 +103,16 @@ mod tests {
...
     }
 
+    #[wasm_bindgen_test]
+    fn test_serial_number() {
+        let pk = "APrivateKey1zkpDeRpuKmEtLNPdv57aFruPepeH1aGvTkEjBo8bqTzNUhE";
+        let record = RecordPlaintext::from_string(RECORD).unwrap();
+        let program_id = "token.aleo";
+        let record_name = "token";
+        let expected_sn = "4564977995400415519058823909143155627601970323571971278914520967771079582104field";
+        assert_eq!(expected_sn, record.serial_number(pk, program_id, record_name))
```
iamalwaysuncomfortable:
```suggestion
        assert_eq!(expected_sn, record.serial_number_string(pk, program_id, record_name))
```

iamalwaysuncomfortable (commented):
Some minor further changes suggested - feel free to reach out to discuss directly.

fulltimemike:
@iamalwaysuncomfortable, thanks for the excellent feedback, on github and offline. I'll recap the discussion below for anyone else who wants to look at the PR.

I ended up implementing the `into` trait for the wasm private key, since the native `serial_number` function takes the native private key instead of the wasm privateKey class. I also needed to pass the wasm privateKey into the function as a reference, since I wasn't able to use the same privateKey multiple times for creating creating serial numbers for multiple plaintext records. We could instead clone the privateKey before passing it to the function instead of passing a reference, maybe by using the [`getter_with_clone`](https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-rust-exports/getter_with_clone.html) wasm-bindgen attribute? I'll defer to whatever you think is better.

One bug I ran into that's not specific to Aleo, just using wasm in general -- loading the wasm data in nodejs wasn't working properly. The classes defined in the wasm were exported fine, but the actual wasm data was undefined. It turns out there was [a wasm-pack update](https://github.com/rustwasm/wasm-pack/pull/1224) that broke how the wasm is initiated on import, which isn't a problem when loading asynchronously in a react hook (like in the aleo sample website), but is a problem when loading in our wallet. Until the wasm-pack devs fix it, there's a small hacky fix that seems like a good workaround. After generating the wasm with `wasm-pack build`, we can modify the `aleo_wasm.js` file that's produced:
```
import * as wasm from "./aleo_wasm_bg.wasm";
import { __wbg_set_wasm } from "./aleo_wasm_bg.js";
__wbg_set_wasm(wasm);
export * from "./aleo_wasm_bg.js";
```
Remove lines:
```
import { __wbg_set_wasm } from "./aleo_wasm_bg.js";
__wbg_set_wasm(wasm);
```
Then modify the aleo_wasm_bg.js file:
```
let wasm;
export function __wbg_set_wasm(val) {
    wasm = val;
}
```

Replace those lines with:
```
import * as wasm from './aleo_wasm_bg.wasm';
```
This fixes the issue for loading outside of react hooks.

wasm/src/record/record_plaintext.rs
```diff
@@ -43,6 +46,16 @@ impl RecordPlaintext {
...
+
+    #[wasm_bindgen(js_name = serialNumberString)]
+    pub fn serial_number_string(&self, private_key: &PrivateKey, program_id: &str, record_name: &str) -> String {
+        let parsed_program_id = ProgramIDNative::from_str(program_id).unwrap();
+        let record_identifier = IdentiferNative::from_str(record_name).unwrap();
+        let commitment = self.to_commitment(&parsed_program_id, &record_identifier).unwrap();
+
+        let serial_number = RecordPlaintextNative::serial_number(private_key.clone().into(), commitment).unwrap();
+        serial_number.to_string()
+    }
```
iamalwaysuncomfortable: One final nit, we generally want to avoid `unwrap()` within Rust in general when it's possible the code will error. In the `wasm` context, unwrapping will cause a panic in the `wasm` environment which cannot be handled in the javascript environment other than a try/catch statement. Even if you do this in javascript the errors are indecipherable in a general wasm panic. Creating a Result Enum to handle errors allows the developer to debug their code better.

Specifically the `Result` types in SnarkVM are of type `anyhow::Result` which is a `Result` type from a crate called `anyhow`. Wasm doesn't directly support this so we use an error transformation covnenience method to transform this type to a Result<String, String> which the wasm ffi supports.

Also secondly, any type that implements the `Copy` trait does not need to be cloned as rust takes care of copying the memory directly. So you can pass by value. Or in the case that you have a reference like this function you've defined you can copy by using the `deref` operator.

Thus you have
```rust
#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
pub struct PrivateKey<N: Network> {
```

And you can `deref` like so
```rust
*private_key
// Or in the case you're using it
*private_key.into()
```


```suggestion
    pub fn serial_number_string(&self, private_key: &PrivateKey, program_id: &str, record_name: &str) -> Result<String, String> {
        let parsed_program_id = ProgramIDNative::from_str(program_id).map_err(|_| "Invalid ProgramID specified".to_string());
        let record_identifier = IdentiferNative::from_str(record_name).map_err(|_| "Invalid Identifier specified for record".to_string());
        let commitment = self.to_commitment(&parsed_program_id, &record_identifier).map_err(|_| "Record does not have a commitment and is likely either invalid or unspent".to_string())

        let serial_number = RecordPlaintextNative::serial_number(*private_key.into(), commitment).map_err(|_| "Serial number derivation failed".to_string());
        serial_number.to_string()
    }
```

wasm/src/record/record_plaintext.rs
```diff
@@ -92,6 +105,27 @@ mod tests {
         assert_eq!(record.gates(), 99);
     }
 
+    #[wasm_bindgen_test]
+    fn test_serial_number() {
+        let pk = PrivateKey::from_string("APrivateKey1zkpDeRpuKmEtLNPdv57aFruPepeH1aGvTkEjBo8bqTzNUhE").unwrap();
```
iamalwaysuncomfortable, line 121: Unwraps however are fine in tests, and actually "mostly" desirable because you want the tests to fail if you're expecting success (that is, unless you're testing for specific Error variants - but that's a convo for another time)

wasm/src/record/record_plaintext.rs
```diff
@@ -43,6 +46,16 @@ impl RecordPlaintext {
     pub fn gates(&self) -> u64 {
         ***self.0.gates()
     }
+
```
iamalwaysuncomfortable, line 49: Also always try to comment public `functions` `methods` and `structs` as they're meant to be part of the outward facing api. `Rustdoc` comments are specified with three slashes like `///` whereas regular comments which won't show up in the documation are two slashes `//`

```suggestion
/// Attempt to get the serial number of a record to determine whether or not is has been spent
```

iamalwaysuncomfortable (commented):
A couple further nits on error handling, copy vs clone, and doc comments.

Other than that this is great, thanks a ton for doing this.

Let me know if you need to discuss anything

iamalwaysuncomfortable:
> I ended up implementing the `into` trait for the wasm private key, since the native `serial_number` function takes the native private key instead of the wasm privateKey class.

The `into` method is the right way to do this, however in rust it's definitely more idiomatic to use the `From` trait to achieve this functionality. Also privateKey implements copy in rust so you can use the deref operator stated the comments above to copy it.

> One bug I ran into that's not specific to Aleo, just using wasm in general -- loading the wasm data in nodejs wasn't working properly

Good catch on this - it's unfortunate that change was made, but thank you posting for the workaround

wasm/src/account/private_key.rs
```diff
@@ -114,6 +114,12 @@ impl From<PrivateKeyNative> for PrivateKey {
     }
 }
 
+impl Into<PrivateKeyNative> for PrivateKey {
+    fn into(self) -> PrivateKeyNative {
+        self.0
+    }
+}
+
```
iamalwaysuncomfortable: The `From` trait is always preferred in Rust over `Into`. It will still allow you to use the `into()` function.

```suggestion
impl From<PrivateKey> for PrivateKeyNative {
    fn from(private_key: &PrivateKey) -> Self {
        private_key.0
    }
}

impl From<&PrivateKey> for PrivateKeyNative {
    fn from(private_key: &PrivateKey) -> Self {
        *private_key.0
    }
}
```

fulltimemike: That makes a ton of sense, thanks. I saw that in the Rust docs, but I was hesitant to change the underlying native private key in snarkVM -- adding the trait in the wasm code is very clean!

wasm/src/record/record_plaintext.rs
fulltimemike: Great suggestion, this will make it easier for wasm-clients (like me!)

wasm/src/record/record_plaintext.rs
fulltimemike: Thanks for the explanation of why we want to use `Result` instead of just `unwrap`; mapping the errors like you've outlined will be a much better experience for javascript clients. The copy instead of clone suggestion makes sense to me too. Only thing I would change is the error message on the `to_commitment` function -- since this is just a hash of the program id, the record name, and the data of the record itself, this should only error if hashing or converting the plaintext record object to bits fails. We should be able to compute the commitment even if the record is unspent. So maybe a better message is `"A commitment for this record and program could not be computed, likely because the record data is invalid"`?

wasm/src/record/record_plaintext.rs
```diff
@@ -92,6 +116,49 @@ mod tests {
...
+    #[wasm_bindgen_test]
+    fn test_serial_number_invalid_record_name_returns_err_string() {
+        let pk = PrivateKey::from_string("APrivateKey1zkpDeRpuKmEtLNPdv57aFruPepeH1aGvTkEjBo8bqTzNUhE").unwrap();
+        let record = RecordPlaintext::from_string(RECORD).unwrap();
+        let program_id = "token.aleo";
+        let record_name = "not a real record name";
+        let expected_value = "Invalid Identifier specified for record".to_string();
+        assert_eq!(record.serial_number_string(&pk, program_id, record_name).err(), Some(expected_value.into()));
+    }
```
fulltimemike, line 160: These tests might be overkill, but I added to them to demonstrate the behavior. Wasn't sure how to do an equivalent test for a commitment failure though, as that's not as trivial as passing in an incorrect string.

iamalwaysuncomfortable: More test coverage is always better - happy to have further tests. Thank you!

iamalwaysuncomfortable (approved):
Great job!

PR #472, opened 2023-02-15
Add proof generation to wasm:

evanmarshall:
1. Add ability to create transfers from wasm
2. Update website to provide example

## Motivation

Light clients like the wallet should have the ability to generate proofs.

## Test Plan

The easiest way to test is by running the tests or through the Website UI.
Current proof generation takes ~400 seconds on a 2017 MacBook Pro.
Deserialization of the ~200MB prover key file takes ~10 seconds.

## Related PRs

This won't be ready to land until https://github.com/AleoHQ/snarkVM/pull/1367 lands

Screenshot 2023-02-14 at 5 02 54 PM

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,9 +21,19 @@ edition = "2018"
 members = ["rust", "wasm" ]
 
 [workspace.dependencies.snarkvm]
+path="../snarkVM"
 version = "0.9.11"
 
 [workspace.dependencies.snarkvm-wasm]
+path="../snarkVM/wasm"
+version = "0.9.11"
+
+[workspace.dependencies.snarkvm-synthesizer]
+path="../snarkVM/synthesizer"
+version = "0.9.11"
+
+[workspace.dependencies.snarkvm-console]
+path="../snarkVM/console"
 version = "0.9.11"
 
 [lib]
@@ -40,62 +50,62 @@ wasm = [ "aleo-wasm" ]
 [dependencies]
 num-format = "0.4.4"
 
-  [dependencies.aleo-rust]
-  path = "rust"
-  version = "0.3.3"
-  optional = true
+[dependencies.aleo-rust]
+path = "rust"
+version = "0.3.3"
+optional = true
 
-  [dependencies.aleo-wasm]
-  path = "wasm"
-  version = "0.3.3"
-  optional = true
+[dependencies.aleo-wasm]
+path = "wasm"
+version = "0.3.3"
+optional = true
 
-  [dependencies.anyhow]
-  version = "1.0"
+[dependencies.anyhow]
+version = "1.0"
 
-  [dependencies.clap]
-  version = "3.2"
-  features = [ "derive" ]
+[dependencies.clap]
+version = "3.2"
+features = [ "derive" ]
 
-  [dependencies.colored]
-  version = "2"
+[dependencies.colored]
+version = "2"
 
-  [dependencies.parking_lot]
-  version = "0.12"
+[dependencies.parking_lot]
+version = "0.12"
 
-  [dependencies.rand]
-  version = "0.8"
-  default-features = false
+[dependencies.rand]
+version = "0.8"
+default-features = false
 
-  [dependencies.rand_chacha]
-  version = "0.3.0"
-  default-features = false
+[dependencies.rand_chacha]
+version = "0.3.0"
+default-features = false
 
-  [dependencies.rpassword]
-  version = "7.2.0"
+[dependencies.rpassword]
+version = "7.2.0"
 
-  [dependencies.self_update]
-  version = "0.32"
+[dependencies.self_update]
+version = "0.32"
 
-  [dependencies.serde]
-  version = "1"
+[dependencies.serde]
+version = "1"
 
-  [dependencies.serde_json]
-  version = "1"
+[dependencies.serde_json]
+version = "1"
 
-  [dependencies.snarkvm]
-  workspace = true
-  features = [ "aleo-cli", "circuit", "console", "parallel", "snarkvm-synthesizer"]
+[dependencies.snarkvm]
+workspace = true
+features = [ "aleo-cli", "circuit", "console", "snarkvm-synthesizer"]
 
-  [dependencies.thiserror]
-  version = "1.0"
+[dependencies.thiserror]
+version = "1.0"
 
-  [dependencies.tokio]
-  version = "1.22"
-  features = [ "rt" ]
+[dependencies.tokio]
+version = "1.22"
+features = [ "rt" ]
 
-  [dependencies.warp]
-  version = "0.3"
+[dependencies.warp]
+version = "0.3"
 
 [dev-dependencies.rusty-hook]
 version = "0.11.2"
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -53,13 +53,22 @@ version = "1.0.91"
 
 [dependencies.snarkvm]
 version = "0.9.11"
-features = [ "parallel" ]
+path="../../snarkVM"
 optional = true
 
 [dependencies.snarkvm-wasm]
 version = "0.9.11"
-features = [ "parallel" ]
+path="../../snarkVM/wasm"
 optional = true
+
+[dependencies.snarkvm-synthesizer]
+path = "../../snarkVM/synthesizer"
+default-features = false
+features = [ "wasm" ]
+
+[dependencies.snarkvm-console]
+path = "../../snarkVM/console"
+default-features = false
 
 [dev-dependencies.bencher]
 version = "0.1.5"
--- a/wasm/src/program/proving_key.rs
+++ b/wasm/src/program/proving_key.rs
@@ -50,9 +50,9 @@ impl Deref for ProvingKey {
     }
 }
 
-impl Into<ProvingKeyNative> for ProvingKey {
-    fn into(self) -> ProvingKeyNative {
-        self.0
+impl From<ProvingKey> for ProvingKeyNative {
+    fn from(proving_key: ProvingKey) -> ProvingKeyNative {
+        proving_key.0
     }
 }
 
--- a/wasm/src/program/proving_key.rs
+++ b/wasm/src/program/proving_key.rs
@@ -1,0 +1,73 @@
+// Copyright (C) 2019-2023 Aleo Systems Inc.
+// This file is part of the Aleo library.
+
+// The Aleo library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// The Aleo library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with the Aleo library. If not, see <https://www.gnu.org/licenses/>.
+
+use crate::{
+  types::{ProvingKeyNative, ToBytes, FromBytes},
+};
+
+use std::{ops::Deref};
+
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+#[derive(Clone, Debug)]
+pub struct ProvingKey(ProvingKeyNative);
+
+#[wasm_bindgen]
+impl ProvingKey {
+    #[wasm_bindgen]
+    pub fn from_bytes(bytes: Vec<u8>) -> ProvingKey {
+        console_error_panic_hook::set_once();
+        let proving_key = ProvingKeyNative::from_bytes_le(&bytes);
+        Self(proving_key.unwrap())
+    }
+
+    #[wasm_bindgen]
+    pub fn to_bytes(&self) -> Result<Vec<u8>, String> {
+        console_error_panic_hook::set_once();
+        self.0.to_bytes_le().map_err(|_| "Failed to serialize prover key".to_string())
+    }
+}
+
+impl Deref for ProvingKey {
+    type Target = ProvingKeyNative;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Into<ProvingKeyNative> for ProvingKey {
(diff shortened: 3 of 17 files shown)


Cargo.toml
```diff
@@ -21,9 +21,19 @@ edition = "2018"
 members = ["rust", "wasm" ]
 
 [workspace.dependencies.snarkvm]
+path="../snarkVM"
```
evanmarshall, line 24: I'll update these when the associated snarkVM changes land: https://github.com/AleoHQ/snarkVM/pull/1367

wasm/src/program/proving_key.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use wasm_bindgen_test::*;
+
+    #[wasm_bindgen_test]
+    #[ignore]
```
evanmarshall, line 66: I'm not sure if there's a better way to test this. It requires a rather large file and may not be suitable for CI.

wasm/src/program/transaction.rs
```diff
@@ -0,0 +1,76 @@
...
+  owner: aleo184vuwr5u7u0ha5f5k44067dd2uaqewxx6pe5ltha5pv99wvhfqxqv339h4.private,
+  gates: 1159017656332810u64.private,
+  _nonce: 1635890755607797813652478911794003479783620859881520791852904112255813473142group.public
+}";
+
+    const ALEO_PRIVATE_KEY: &str = "APrivateKey1zkp3dQx4WASWYQVWKkq14v3RoQDfY2kbLssUj7iifi1VUQ6";
+
+    #[test]
+    #[ignore]
+    fn test_build_transaction() {
```
evanmarshall, line 73: Similarly, this test requires a large file and also takes a long time. Any suggestions for testing are appreciated.

website/src/db/index.js
```diff
@@ -0,0 +1,115 @@
...
+  //   name: 'MintProver',
+  //   url: 'https://aleo-public.s3.us-west-2.amazonaws.com/testnet3/mint.prover.11fa6f2'
+  // },
+  // {
+  //   name: 'MintVerifier',
+  //   url: 'https://aleo-public.s3.us-west-2.amazonaws.com/testnet3/mint.verifier.e1ae0c8'
+  // },
+  {
+    name: 'TransferProver',
+    url: 'https://aleo-public.s3.us-west-2.amazonaws.com/testnet3/transfer.prover.837ad21'
```
evanmarshall, line 20: Right now, we're hosting these files. AleoHQ's CORS policy on the s3 bucket prevents this from working.

rust/Cargo.toml
```diff
@@ -53,14 +53,23 @@ version = "1.0.91"
 
 [dependencies.snarkvm]
 version = "0.9.11"
-features = [ "parallel" ]
+path="../../snarkVM"
 optional = true
 
 [dependencies.snarkvm-wasm]
 version = "0.9.11"
-features = [ "parallel" ]
```
iamalwaysuncomfortable, line 61: It's likely we'll need to sort the issue of the parallel feature out more in the [Make Marlin key types public PR ](https://github.com/demox-labs/snarkVM/pull/2)because I believe they'll want those to be slightly more additive than core behavior changing, we'll make this a priority.

wasm/src/program/proving_key.rs
```diff
@@ -0,0 +1,73 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Into<ProvingKeyNative> for ProvingKey {
+    fn into(self) -> ProvingKeyNative {
+        self.0
+    }
+}
```
iamalwaysuncomfortable: From is always idiomatically preferred over Into.

```suggestion
impl From<ProvingKeyNative> for ProvingKey {
    fn from(self) -> ProvingKeyNative {
        self.0
    }
}
```

wasm/src/program/proving_key.rs
iamalwaysuncomfortable: How large roughly?

wasm/src/program/transaction.rs
```diff
@@ -0,0 +1,76 @@
...
+
+#[wasm_bindgen]
+impl TransactionBuilder {
+    /// Creates an execute transaction from a full proof of execution
+    pub fn build_transfer_full(
+        private_key: PrivateKey,
+        proving_key: ProvingKey,
+        address: Address,
+        amount: u64,
+        record: RecordPlaintext,
```
iamalwaysuncomfortable, line 29: I think this at its core is the right abstraction, passing the proving key. We will probably also need the inclusion proof data here too.

Below suggest just passing the PrivateKey as you can derive the record from it and renaming amount to "gates" to disambiguate them from credits.
```suggestion
        private_key: PrivateKey,
        proving_key: ProvingKey,
        gates: u64,
        record: RecordPlaintext,
```

wasm/src/program/transaction.rs
```diff
@@ -0,0 +1,76 @@
...
+    ) -> String {
```
iamalwaysuncomfortable: This should be a result enum so a more informative error can be piped up to js.

```suggestion
    ) -> Result<String, String> {
```

wasm/src/program/transaction.rs
```diff
@@ -0,0 +1,76 @@
...
+        console_error_panic_hook::set_once();
+
+        let process = Process::load_wasm().unwrap();
+        let credits_program = Program::credits().unwrap();
```
iamalwaysuncomfortable: In general we want to avoid using unwraps here. Usually we use anyhow's error handling https://docs.rs/anyhow/latest/anyhow/ which we transform to a generic Result enum here.

```suggestion
        let process = Process::load_wasm().map_err(|_| "Could not initialize wasm".to_string())?;
        let credits_program = Program::credits().map_err(|_| "Could not access credits program".to_string())?;
```

wasm/src/program/transaction.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub fn build_transfer_full(
+        private_key: PrivateKey,
+        proving_key: ProvingKey,
+        address: Address,
+        amount: u64,
+        record: RecordPlaintext,
+    ) -> String {
+        console_error_panic_hook::set_once();
+
+        let process = Process::load_wasm().unwrap();
```
iamalwaysuncomfortable: This is also the right abstraction I think, but we should audit the load_wasm method as we may need some edits to the underlying storage (i.e. we don't necessarily need to initialize all the powers of tau) abstractions and feature flags.

wasm/src/program/transaction.rs
```diff
@@ -0,0 +1,76 @@
...
+        let credits_program = Program::credits().unwrap();
+
+        process.insert_proving_key(credits_program.id(), &Identifier::from_str("transfer").unwrap(), proving_key.into()).unwrap();
+
+        let mut amount_str = amount.to_string();
+        amount_str.push_str("u64");
+        let inputs = [record.to_string(), address.to_string(), amount_str];
+        let rng = &mut rand::thread_rng();
```
iamalwaysuncomfortable, line 43: I have some nits about these, but let's solve the bigger questions about powers of tau downloads, the structure of the SnarkVM edits and inclusion proofs prior to addressing them.

website/src/db/index.js
iamalwaysuncomfortable: We may opt to put up a separate static endpoint for transfers specifically, we'll need to chat with the team on it.

website/src/db/index.js
```diff
@@ -0,0 +1,115 @@
...
+    name: 'InclusionVerifier',
+    url: 'https://aleo-public.s3.us-west-2.amazonaws.com/testnet3/inclusion.verifier.3f4d6b7'
+  }
+];
+
+export async function getAllSavedFiles() {
+  return await db.table("parameters").toArray();
+}
+
+async function downloadFile(url) {
```
iamalwaysuncomfortable, line 72: A few nits on this function too for after we solve the other issues, but I like the approach. Question - were there any issues with creating an IndexDB and storing the files there - (i.e. like they can only be used once per site?)?

website/src/tabs/transaction/SendCredits.js
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    const buildTransaction = async () => {
+      try {
+        let startTime = performance.now();
+        await downloadAndStoreFiles();
+        console.log(`Download Completed: ${performance.now() - startTime} ms`);
+        startTime = performance.now();
+
+        const files = await getAllSavedFiles();
```
iamalwaysuncomfortable: Maybe these can be appropriately filtered before js?

website/src/tabs/transaction/SendCredits.js
```diff
@@ -0,0 +1,91 @@
...
+        console.log(`Fetching Transfer Prover from IndexedDb Completed: ${performance.now() - startTime} ms`);
+        startTime = performance.now();
+
+        const pK = aleo.PrivateKey.from_string(privateKey);
+        const provingKey = aleo.ProvingKey.from_bytes(transferProver.bytes);
+        console.log(`Deserialized proving key Completed: ${performance.now() - startTime} ms`);
+        startTime = performance.now();
+        const add = aleo.Address.from_string(toAddress);
+        const rec = aleo.RecordPlaintext.fromString(plaintext);
+        const transaction = await aleo.TransactionBuilder.build_transfer_full(pK, provingKey, add, BigInt(amount), rec);
```
iamalwaysuncomfortable: Even though this takes a while to complete, I think we should aim for getting this feature done prior to optimization

iamalwaysuncomfortable (changes requested):
Added some commentary on style and nits to get this up to a ready state. In general I believe we need to get the SnarkVM changes in [Make Marlin key types public](https://github.com/demox-labs/snarkVM/pull/2) reviewed/edited as necessary/and approved prior to addressing those nits.

The main things we need to address in the [Make Marlin key types public](https://github.com/demox-labs/snarkVM/pull/2) are the following. We have spoken about this offline, but writing it here for community visibility:
* How to properly handle the `parallel` feature issue
* Powers of tau storage/access via the load_wasm() method
* Some of the general edits in SnarkVM to existing methods
* How to manage inclusion proofs

rust/Cargo.toml
evanmarshall: Yes. Btw this is the correct PR now to reference: https://github.com/AleoHQ/snarkVM/pull/1367

wasm/src/program/proving_key.rs
evanmarshall: It's a 200mb file. We probably also have to ensure it's downloaded for testing

wasm/src/program/transaction.rs
evanmarshall: I'm not quite sure what you're suggesting here. Are you suggesting we pass in an `RecordCiphertext` instead of a `RecordPlaintext` b/c we also have to use the private key?

wasm/src/program/proving_key.rs
evanmarshall: Are you saying it should be?
```
impl From<ProvingKey> for ProvingKeyNative {
    fn from(proving_key: ProvingKey) -> ProvingKeyNative {
        proving_key.0
    }
}
```
and then we'll call it: `ProvingKeyNative::from(proving_key)`

wasm/src/program/transaction.rs
evanmarshall: Ideally, no storage will be necessary and all data can be passed in, even if it includes powers of tau. I think the client should be responsible for passing in any required data and the wasm should be stateless.

website/src/db/index.js
evanmarshall: IndexedDB is per site so they will need to be downloaded once per site. I don't actually expect them to be used that way as the wallet (and all chrome extensions) have their own IndexedDB instance so you also only have to download these once per extension.

evanmarshall: Another thought, if a site wants to use these files and doesn't want to download them, we can update the [aleo-wallet-adapter](https://github.com/demox-labs/aleo-wallet-adapter) to serve these files so as long as they have a wallet installed, a website can get the files from there instead of adding another download. 

I don't think it makes sense to add a dependency on the aleo-wallet-adapter here though.

website/src/tabs/transaction/SendCredits.js
evanmarshall: Yeah, this was lazy. I'll update it.

website/src/tabs/transaction/SendCredits.js
evanmarshall: We've started the optimization work and I think it can be done mostly in parallel as it will be mostly snarkVM changes. This of course has to land first. 

I'd rather not maintain aleo & snarkVM forks 😄

wasm/src/program/proving_key.rs
iamalwaysuncomfortable: Understood - we'll think about this.

wasm/src/program/transaction.rs
iamalwaysuncomfortable: Oh yes, I'm sorry, I meant 'gates' not records!

wasm/src/program/transaction.rs
iamalwaysuncomfortable: Actually, I was adderssing a separate concern, but this function isn't the one that brings it up, so I'll resolve this comment.

wasm/src/program/transaction.rs
iamalwaysuncomfortable: 200 mb in ci isn't terrible, but we might be able to get away with something like a mock key

website/src/db/index.js
iamalwaysuncomfortable: I'm not as familiar with chrome extensions, but if you say the extension can store these proving keys in the chrome's extension that's probably good too.

That's an interesting idea to have the wallet serve them, agreed it doesn't make sense to do that here.

website/src/tabs/transaction/SendCredits.js
iamalwaysuncomfortable: So the parallelization is potentially working? If so, and we can keep the parallel feature flags in SnarkVM, that would be optimal.

iamalwaysuncomfortable:
Thanks to our work together we've successfully added this!

PR #482, opened 2023-03-01
Improve rest api call ui

collinc97:
## Problem Overview
Previously, the REST API form fields would make a new api request every time a character within the form field was edited.

## Fixes
- [x] A new REST API call is only made when the user submits the search bar input (by hitting enter or clicking the magnifying glass)
- [x] Upon success, the search bar turns green and the output is displayed below (block data, program bytecode, etc...)
- [x] Upon failure, the search bar turns red 
- [x] The search bar can be cleared and the border color turns back to default grey.

### DIFF
--- a/website/src/tabs/rest/GetProgram.js
+++ b/website/src/tabs/rest/GetProgram.js
@@ -5,35 +5,67 @@ import { CopyButton } from "../../components/CopyButton";
 
 export const GetProgram = () => {
     const [program, setProgram] = useState(null);
-    const [programName, setProgramName] = useState(null);
+    const [programID, setProgramID] = useState(null);
+    const [programExists, setProgramExists] = useState(false);
 
+    // Returns the program id if the user changes it or the "Demo" button is clicked.
     const onChange = (event) => {
+        if (event.target.value !== null) {
+            setProgramID(event.target.value);
+        }
+        return programID;
+    }
+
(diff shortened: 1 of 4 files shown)


website/src/tabs/rest/GetProgram.js
```diff
@@ -5,45 +5,77 @@ import { CopyButton } from "../../components/CopyButton";
 
 export const GetProgram = () => {
     const [program, setProgram] = useState(null);
-    const [programName, setProgramName] = useState(null);
+    const [programID, setProgramID] = useState(null);
+    const [status, setStatus] = useState("");
```
iamalwaysuncomfortable, line 9: Each of the states in these `useState` vars modify the same var if they're named the same way. These are all named status. Is this by design? I wonder if there's a way state could get out of order. Seems really low risk honestly though, and I'm not %100 sure what the best state management practice is in a lightweight web app like this so this is probably fine.

Just something to note.

iamalwaysuncomfortable (approved):
Quick question about state, but otherwise really beautiful changes!

website/src/tabs/rest/GetProgram.js
collinc97: I'm not sure what the best React practice is here. When clicking on the webpage I do not not notice any strange state sharing between `status` variables. I will do some research and add any idiomatic changes in a follow up PR.

PR #515, opened 2023-03-28
Update README.md

ghost:
So these links are broken, which I assume has to do with the "default" branch being changed to "testnet3". For now, this will solve the problem, but we probably want to create a more scalable way to do this or set up a process to check that these are updated when/if the default branch changes.

For /wasm/pkg I just put the same URL as wasm, since the actual npm package is linked.

## Motivation

Dead links suck 🤣

## Test Plan

N/A

## Related PRs

N/A


iamalwaysuncomfortable (approved):
Thanks for catching this!

PR #512, opened 2023-03-24
Aleo Development Server

iamalwaysuncomfortable:
## Motivation

Most developers working building applications on Aleo are working with javascript or typescript environments. Unfortunately Aleo tooling mostly consists of rust libraries or cli tools which force developers to build somewhat painful bridges such as OS pipes within cloud containers in order to interact with the Aleo Blockchain.

The Aleo development server is meant to provide a simple REST API for use in local development or cloud-based applications.

## Test Plan
* Integration tests will be written against the server once its running that validate its creating successful executions and deployments.

PR #529, opened 2023-04-11
bump snarkvm v0.9.16

collinc97:


rust/Cargo.toml
```diff
@@ -49,13 +49,11 @@ version = "2.6.2"
...
 [dependencies.snarkvm-console]
-features = [ "parallel" ]
-version = "0.9.15"
+version = "0.9.16"
 
 [dependencies.snarkvm-synthesizer]
-features = [ "parallel" ]
+version = "0.9.16"
 optional = true
-version = "0.9.14"
```
howardwu: These are duplicate deps to `snarkVM` (see the dep below). Please import them directly from `snarkvm::synthesizer` and `snarkvm::console`.

rust/src/program/execute.rs
```diff
@@ -148,7 +150,7 @@ impl<N: Network> ProgramManager<N> {
 mod tests {
     use super::*;
     use crate::{random_program_id, AleoAPIClient, RECORD_5_GATES};
-    use snarkvm_console::network::Testnet3;
+    use snarkvm::console::network::Testnet3;
```
iamalwaysuncomfortable, line 153: Currently we can't remove this, because even without default features, the SnarkVM crate imports lead the wasm unable to compile, and the snarkvm_wasm crate currently induces the broken parameter download path from the parameter crate.

iamalwaysuncomfortable: Also having it blows up the size of the `wasm` crate because it imports all the stuff in the SnarkVM base crate

collinc97:
This PR's changes are included in #512

rust/src/program/execute.rs
howardwu: No worries, thanks. For now we can use `snarkvm_console` as its needed. For `synth` we can remove its import.

PR #536, opened 2023-04-16
Ref function

itmilos:
## Motivation

The type of the SECRET OnceCell was changed to hold a fixed-size array of bytes [u8; 16] instead of a dynamic vector Vec<u8>. 

This is because a fixed-size array can be directly used as a reference to a slice &[u8], which is what encode and decode functions expect.

The return type of the function was changed to &'static [u8] to match the type of the value held by the SECRET OnceCell.
The closure passed to get_or_init was simplified to just generate a random byte array using thread_rng().gen(), which is more concise than generating a seed array and converting it to a vector. 

The as_ref method is used to return a reference to the byte array held by the OnceCell

## Test Plan

`#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_jwt_secret() {
        let secret1 = jwt_secret();
        let secret2 = jwt_secret();
        assert_eq!(secret1, secret2);
    }
}`

This test checks that the jwt_secret function returns the same secret every time it is called. 

The OnceCell type ensures that the secret is only generated once, and then reused on subsequent calls. 
The thread_rng function generates a new random number generator, which is used to generate a random 16-byte seed. 
This seed is then used as the secret for the JSON Web Token.


onetrickwolf:
Closing seems to be out of date/stale but thanks for the contribution!

onetrickwolf:
This may make sense for https://github.com/AleoHQ/aleo-rust now?

PR #589, opened 2023-06-05
Add wasm integration tests

iamalwaysuncomfortable:
## Motivation

The tests in the wasm crate currently fall under 2 categories. Those that only run in the browser and those that run in nodejs and the browser.

This PR separates the tests that can only be run in a browser into a separate tests directory that can be run with the invocation:
`wasm-pack test --chrome --tests offchain`


collinc97 (approved):
LGTM thank you for breaking up the tests into separate PRs.

PR #614, opened 2023-06-18
Create README.ua.md

Mistmorn0:
Add a Ukrainian translation of README.md


onetrickwolf:
Hey thanks for the contribution! I think this may be fairly out of date now though and we haven't really defined a way to maintain translations at this point. We may reach out in the future though, thank you anyway!

PR #615, opened 2023-06-19
Create Account fromDerivePath

daoauth:
## Motivation

The purpose of using the [slip44(bip44)](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) protocol is to create an account.

```
const path = 'm/44'/41440'/0'/0'/0'';
```

## Test Plan

I have conducted tests using a seed with the BIP44 protocol.

## Related PRs

https://github.com/AleoHQ/sdk/issues/611

### DIFF
--- a/sdk/package.json
+++ b/sdk/package.json
@@ -36,8 +36,11 @@
   "homepage": "https://github.com/AleoHQ/sdk#readme",
   "dependencies": {
     "@aleohq/wasm": "0.4.2",
+    "@noble/hashes": "^1.3.1",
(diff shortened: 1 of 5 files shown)


iamalwaysuncomfortable:
Give me a few days to do a proper cryptographic review on this and I'll comment

onetrickwolf:
@daoauth sorry for delay we may close this for now as I think if we are to implement this we may do it from the rust side of things, but thanks for the contribution definitely helped at least get our minds thinking about this!

Issue #598, opened 2023-06-12
[Proposal] Auto-generate input fields from program source code on aleo.tools


iamalwaysuncomfortable:
## 💥 Proposal - Auto-generate input fields from program source code on aleo.tools

The `aleo-wasm` toolkit provides the ability to get a javascript object of the inputs for a program.

When a user enters a program on aleo.tools - a webform should be created that has fields for all the inputs to that program. Users should then be able to enter their inputs to the program and submit it to execute the program. The form should allow the user to choose which function they want to execute

The form should also:
* Validate the inputs are valid and of the correct type prior to submission
* Display a useful error message if the inputs are incorrect
* Allow users to clearly select the function within the program they want to execute

PR #618, opened 2023-06-21
[Feature] Improve HMR and build performance of aleo.tools

### LINKED ISSUE #617: [Proposal] Improve HMR performance of aleo.tools
## 💥 Proposal

Improve https://aleo.tools compile time and hot module reload time.

Currently takes me over 3 seconds to compile and hot reload the aleo.tools dev environment but we should be able to get that down to ~100ms by doing the following:

- Switching to [Vite](https://vitejs.dev/) 
- Upgrading to React 18
- Integrating [SWC](https://swc.rs/)

onetrickwolf:
## Motivation

Resolves #617 

## Test Plan

Functionality should be identical but compile time and hot reload should be significantly faster (roughly 50x performance in my tests).


onetrickwolf:
Currently have new build in `website-swc` am keeping old `website` folder separate to easily compare and bench. This branch was also coming off of a branch for #598 will be removing that code before coming out of draft.

onetrickwolf:
Unfortunately, a bug in vite means we still have to use Webpack for production builds, but this PR is still worth merging considering how much it speeds up the dev server spinup and HMR. 

Tracking vite bug here: #628; I will remove the web pack once this bug is resolved.

onetrickwolf:
Some issues with how this deploys to netlify, making a few small changes.

collinc97 (approved):
LGTM

iamalwaysuncomfortable (approved):
This looks good, I'd say pending the merge of the new fixes upstream, this looks fine to me.

onetrickwolf:
Okay fixed the vite build.

Had to do a small patch for `wasm-bindgen-rayon` here: https://github.com/GoogleChromeLabs/wasm-bindgen-rayon/compare/main...onetrickwolf:wasm-bindgen-rayon:main

Patch implemented via root (patches do not work in workspaces apparently) `Cargo.toml`: https://github.com/AleoHQ/sdk/pull/618/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R23-R25

I attempted to find a solution that would work for both webpack and vite so I could submit it as a PR to `wasm-bindgen-rayon` but was unable to find a simple one. I may make a PR to them for a feature flag for `non-webpack` bundlers or something though so we can get off this patch eventually or maybe implement the solution in vite.

onetrickwolf:
Okay this should be good to go. Decided to go back to the webpack build for now after some more research I am pretty convinced this is actually best solved with a vite patch rather than patching `wasm-bindgen-rayon`.

Currently deployed to https://staging.aleo.tools/ seems to be fine!

There's a few non-ideal configurations from the hybrid approach but nothing major, still a huge improvement. Hopefully can be cleared up in the near future. In staging will throw that it is missing `index.jsx` due to a difference in how vite and webpack do entry points but will not affect anything.

PR #655, opened 2023-07-11
[Feature] Add offline program execution option

iamalwaysuncomfortable:
## Motivation

Currently there is no mechanic for execution programs offline in the Rust SDK. This PR introduces the ability to execute programs online and create an object with facilities to verify the proof which can be shared with 3rd parties. Currently there's no wire serialization format included in this, but this will be added in future work.

### DIFF
--- a/rust/src/program/execute.rs
+++ b/rust/src/program/execute.rs
@@ -15,8 +15,59 @@
 // along with the Aleo SDK library. If not, see <https://www.gnu.org/licenses/>.
 
 use super::*;
+use snarkvm::prelude::AleoID;
 
 impl<N: Network> ProgramManager<N> {
+    /// Create an offline execution of a program to share with a third party.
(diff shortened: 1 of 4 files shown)


onetrickwolf (approved):

rust/src/program/execute.rs
```diff
@@ -15,8 +15,59 @@
 // along with the Aleo SDK library. If not, see <https://www.gnu.org/licenses/>.
 
 use super::*;
+use snarkvm::prelude::AleoID;
 
 impl<N: Network> ProgramManager<N> {
+    /// Create an offline execution of a program to share with a third party.
```
collinc97, line 21: add note on programs that use online public state

iamalwaysuncomfortable: Done. Let me know what you think of it.

collinc97 (approved):
LGTM

PR #670, opened 2023-07-19
[Feature] Add multiple transfer type options to Aleo.tools

iamalwaysuncomfortable:
## Motivation

There are 4 different types of transfers currently possible
`transfer_private` - Transfer microcredits privately and create another record in the target address with a microcredits balance
`transfer_private_to_public` - Transfer a microcredits privately to a public microcredits mapping in credits.aleo for the target address
`transfer_public` - Transfer microcredits publicly from one microcredits mapping to another
`transfer_public_to_private` - Transfer microcredits from a public mapping to a private record

Aleo.tools currently only supports the ability to execute `transfer_private` - this PR the ability execute all 4 types of transfers on aleo.tools

### DIFF
--- a/wasm/src/programs/manager/transfer.rs
+++ b/wasm/src/programs/manager/transfer.rs
@@ -89,20 +89,20 @@ impl ProgramManager {
 
         let (transfer_type, inputs) = match transfer_type {
             "private" => {
-                let inputs = Array::new_with_length(3);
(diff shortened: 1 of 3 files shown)


onetrickwolf (approved):
Thanks for adding this so fast looks good! 🚀

PR #682, opened 2023-07-27
[Feature] Enumerate registers for top level inputs & parameter names for nested structs

### LINKED ISSUE #679: [Bug] `getFunctionInputs` needs to give register for nested inputs need
```
program tictactoe.aleo;

struct Row:
    c1 as u8;
    c2 as u8;
    c3 as u8;

struct Board:
    d1 as Row;
    d2 as Row;
    d3 as Row;


function make_move:
    input r0 as u8.private;
    input r1 as u8.private;
    input r2 as u8.private;
    input r3 as Board.private;
    output r0 as u8.private;
```

The above program requires an input in this format:

```
"1u8" "2u8" "3u8" "{d1:{c1:1u8,c2:1u8,c3:1u8},d2:{c1:1u8,c2:1u8,c3:1u8},d3:{c1:1u8,c2:1u8,c3:1u8}}"
```

You need to denote the register as `d1`, `d2`, and `d3`, but the Aleo WASM `getFunctionInputs` (e.g. `aleoWASM.Program.fromString(value).getFunctionInputs()`) does not give the register names when nested. This is the result of just the `Board` from `getFunctionInputs` but you can see if gives the child registers but not the `d` registers within the Board struct.

```json
{
    "type": "struct",
    "name": "Board",
    "members": [
        {
            "type": "struct",
            "name": "Row",
            "members": [
                {
                    "name": "c1",
                    "type": "u8"
                },
                {
                    "name": "c2",
                    "type": "u8"
                },
                {
                    "name": "c3",
                    "type": "u8"
                }
            ]
        },
        {
            "type": "struct",
            "name": "Row",
            "members": [
                {
                    "name": "c1",
                    "type": "u8"
                },
                {
                    "name": "c2",
                    "type": "u8"
                },
                {
                    "name": "c3",
                    "type": "u8"
                }
            ]
        },
        {
            "type": "struct",
            "name": "Row",
            "members": [
                {
                    "name": "c1",
                    "type": "u8"
                },
                {
                    "name": "c2",
                    "type": "u8"
                },
                {
                    "name": "c3",
                    "type": "u8"
                }
            ]
        }
    ],
    "visibility": "private"
}
```

I would propose possibly structuring it like this for consistency. Here's an example of one of the Rows updated to include the `registry` and also replacing `name` with `registry` reserving `name` for the named data types (which aren't needed for execution but could be useful for UIs).

```json
{
    "register": "d1",
    "type": "struct",
    "name": "Row",
    "members": [
        {
            "register": "c1",
            "type": "u8"
        },
        {
            "register": "c2",
            "type": "u8"
        },
        {
            "register": "c3",
            "type": "u8"
        }
    ]
}
```

### LINKED ISSUE #683: [Feature] Refactor Program Inputs
## 🚀 Feature

Currently in the `Program.getProgramInputs` method, if an input is a nested struct, the `name` field lists the concrete type instead of the parameter name of the struct. This should be changed to the parameter name. 

Additionally it should also list register names for top level inputs.

## Motivation

The current aim is to autogenerate a form from program inputs. When taking the inputs from these forms, the inputs then need to be constructed in text string that can be deserialized into a "Value" object in rust, so the naming MUST be exact.

## Implementation

The following PR outlines the implementation. 
#682

iamalwaysuncomfortable:
## Motivation

For both a `Struct` or `Record` type, its concrete type is its name. However, if a struct is a parameter of a record or another struct, its parameter name is defined in the enclosing struct or record.

In order to create automatically generated Input forms, the input name of the variable needs to be included. This PR changes the "name" field of nested structs from the struct type to the parameter "name". It also adds a field to identify the concrete struct.

## Test Plan
Tests of the correct serialization accompany this struct.

### DIFF
--- a/wasm/src/programs/program.rs
+++ b/wasm/src/programs/program.rs
@@ -69,30 +69,44 @@ impl Program {
             .ok_or_else(|| format!("function {} not found in {}", function_name, self.0.id()))?;
         let function_inputs = Array::new_with_length(function.inputs().len() as u32);
         for (index, input) in function.inputs().iter().enumerate() {
+            let register = JsValue::from_str(&input.register().to_string());
             match input.value_type() {
                 ValueType::Constant(plaintext) => {
-                    function_inputs.set(
-                        index as u32,
-                        self.get_plaintext_input(plaintext, Some("constant".to_string()), None)?.into(),
-                    );
+                    function_inputs.set(index as u32, {
+                        let input = self.get_plaintext_input(plaintext, Some("constant".to_string()), None)?;
+                        Reflect::set(&input, &"register".into(), &register).map_err(|_| "Failed to set property")?;
+                        input.into()
+                    });
                 }
                 ValueType::Public(plaintext) => {
-                    function_inputs.set(
-                        index as u32,
-                        self.get_plaintext_input(plaintext, Some("public".to_string()), None)?.into(),
-                    );
+                    function_inputs.set(index as u32, {
+                        let input = self.get_plaintext_input(plaintext, Some("public".to_string()), None)?;
+                        Reflect::set(&input, &"register".into(), &register).map_err(|_| "Failed to set property")?;
+                        input.into()
+                    });
                 }
                 ValueType::Private(plaintext) => {
-                    function_inputs.set(
-                        index as u32,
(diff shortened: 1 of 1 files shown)


onetrickwolf (approved):
Tested it in my branch, looks like it is returning what we need 👍 🎉

PR #694, opened 2023-07-27
[Feature] Program execution in Aleo Typescript & Javascript SDKs

### LINKED ISSUE #687: Add program management to the Aleo SDK
The Aleo javascript SDK currently does not have the ability to do Program Management (i.e. import a program and get its inputs/records, etc.). Add the ability manage programs within the Aleo SDK using the @aleohq/wasm package.

### LINKED ISSUE #688: Add program deployment to the Aleo SDK
Currently program deployment only exists in the Aleo Wasm SDK. Add this ability within the Aleo Javascript SDK

iamalwaysuncomfortable:
## Motivation

Currently the aleo-wasm crate compiles to javascript bindings to webassembly which allows program execution to be done directly from javascript (available at [@aleohq/wasm](https://www.npmjs.com/package/@aleohq/wasm) on npm). However, this is very much a collection of APIs and not a fully fledged web software development kit. This PR adds the full featured capability of program execution and deployment to the Aleo SDK.

This includes the following features: 
* Program Execution & Deployment in the JS/TS Sdk
* Aleo API methods for submitting transactions, downloading function proving keys, and import resolution
* Re-exports of key methods from the wasm SDK

Part of the future goals of this SDK upgrade are to provide developers of Aleo app a canonical toolkit for building zero knowledge web applications using Aleo cryptography

## Test Plan
* Create integration tests of all new api functionality & program execution functionality


sdk/src/aleo_network_client.ts
```diff
@@ -250,6 +262,88 @@ export class AleoNetworkClient {
...
   }
 
+  /**
+   * Returns the proving and verifying keys for a specified function in an Aleo Program given the url of the prover and
+   * verifier keys
+   *
+   * @param {string} url of the proving key
+   * @param {string} url of the verifying key
+   *
+   * @returns {Promise<[ProvingKey, VerifyingKey] | Error>} Proving and verifying keys for the specified program
```
collinc97: JSDoc complains when you use brackets or curly braces.
Do you mind escaping them in the new docs?
```suggestion
   * @returns {Promise<\[ProvingKey, VerifyingKey\] | Error>} Proving and verifying keys for the specified program
```

collinc97: Actually escaping via the suggestion I made does not work either hmm...

onetrickwolf: @collinc97 not sure if something like this would be worth it but [TypeDoc](https://typedoc.org) might be useful to solve issues like this. It'll automatically pull the types which prevents kind of redundant work and the possibility of the docs getting out of sync with the code.

iamalwaysuncomfortable: I could potentially decouple these methods, however since the verifying key is so much smaller than the proving key and I'm trying to encourage always using them in pairs (rather than encouraging disparate keys accidentally being used) I'm inclined to keep them this way. However, perhaps we could update the method signature

iamalwaysuncomfortable: I just created a type Signature for it, and that solved the issue

onetrickwolf (approved):

Issue #693, opened 2023-07-27
[Feature] Add webworker management to the Aleo SDK


iamalwaysuncomfortable:
## 🚀 Feature

Current instantiating webworkers is a process being handled directly in react. Optimally though, creating a webworker should be encapsulated behind an object which provides the ability to configure inputs, webworker lifetimes, worker sharing, memory management, and error handling.

## Motivation

Web workers are essential for running proofs at a reasonable speed within a WebAssembly environment in the browser. Thus being able to deploy and management them efficiently is essential.

iamalwaysuncomfortable:
Finished by PR

#734

Issue #596, opened 2023-06-12
[Proposal] Create web workers in an top-level component in aleo.tools so they can be shared


iamalwaysuncomfortable:
## 💥 Proposal - Create web workers in an top-level component in aleo.tools so they can be shared

Webworkers are currently started on each react component which does Program Execution (currently the Execute, Deploy, Transfer, Split, and Join components) 

This has the following disadvantages
* A wasteful amount of web workers are started & stopped on each page
* Resources aren’t shared between the components which mean keys synthesized or downloaded can’t be shared between the workers
* It creates a situation where a heavy load on browser’s host memory can be generated

Optimally the web workers should be moved up to a parent component and be shared between the react components which perform program execution

jaketarnow:
Closing this issue as its stale and was iterated upon in https://github.com/ProvableHQ/sdk/pull/700.

PR #700, opened 2023-08-01
[Fix] Improved loading

### LINKED ISSUE #681: [Bug] Aleo.tools fails to load on mobile or slow connections sometimes
Aleo.tools won't render until the full wasm is downloaded. This file is pretty large (~8MB) so on slower connections or mobile it may appear the site is broken (just a white screen) until everything is done loading. Furthermore, if the wasm is downloaded slow enough, the site seems to have an error where it will not render at all; this forces the user to refresh which will cause the site to usually render properly as the wasm is now being pulled instantly from the cache.

Proposed fixes:
- Have some earlier first paint indicating the site is loading
- Fix condition where sites fails to render is wasm is not loaded fast enough (make sure we are properly waiting until it is loaded to try to initiate)
- Potentially looks into faster edge CDNs for the larger files

Replication:
- Clear cache on any browser and reload (sometimes occurs)
- Clear cache on any browser and reload with speed limiting (consistently occurs)
- Try hard reloading on iOS Safari (consistently occurs)

onetrickwolf:
## Motivation

Resolves #681

Sets some ground work for #693: context provider added to ensure app doesn't load until worker is done initializing; this context can eventually be used for worker management (and maybe eventually moved to the TS SDK).

This issue will be iterated on in #693 and #596. This PR is just meant to be a quick fix to resolve the app not loading in some conditions.

## Test Plan

Functionality should be identical but app shouldn't need a refresh on mobile or slow connections. Should see spinner while initial assets are loading.

Preview link: https://improved-loading--aleo-tools-3481ab.netlify.app/


website/src/workers/worker.js
```diff
@@ -656,3 +670,32 @@ self.addEventListener("message", (ev) => {
...
+                if (!imports[key]) {
+                    imports[key] = nestedImports[key];
+                }
+            }
+            imports[import_id] = importedProgram;
+        }
+    }
+    return imports;
+}
+async function getProgram(name) {
```
iamalwaysuncomfortable:
```suggestion
}

async function getProgram(name) {
```

iamalwaysuncomfortable (approved):
One small nit, but otherwise great. Quick question though: does this remove wasm loading until after the react components load?

onetrickwolf:
> One small nit, but otherwise great. Quick question though: does this remove wasm loading until after the react components load?

@iamalwaysuncomfortable No it doesn't the worker still triggers the wasm download before the react components mount so it's no slower to download. The main difference here is just that we are waiting until the worker is ready (i.e. for the wasm to download) before we allow the app to render the main route.

It's kind of a quick hack but it should fix the slow connection white screen which seems to be caused by the app rendering before the wasm is done initiating. In future issues/as part of some of the SDK improvements I have some ideas for properly managing the worker and the wasm to make it easier to navigate and optimal for render speed.

onetrickwolf:
Also want to mention for reference we had an issue where the wasm was being initiated and downloaded twice before: 
image

After:

image

In simulated slow connections this seems to double the wasm load speed.

Throttle profile for reference: 
image

Issue #692, opened 2023-07-27
Create an Aleo 101 app the runs through basics of using Aleo


iamalwaysuncomfortable:
The first app in the create aleo app should be a tutorial on how to use Aleo systems and do things like

* Walk through creating an account and getting credits
* Walk through making a value transfer
* Walk through a program execution

jaketarnow:
Closing this issue as its stale and has been addressed through recent updates to create-leo-app examples

Issue #691, opened 2023-07-27
Create a Readme with create-aleo-app that explains how to get started building a basic web app


iamalwaysuncomfortable:
After the `npm create` command is run, a Readme should be present explaining how to create a web app with Aleo

Issue #591, opened 2023-06-07
[Proposal] Create-Aleo-App


iamalwaysuncomfortable:
## 🔮 Proposal - Create-Aleo-App
Privacy preservation through zero Knowledge computing is an ideal that Aleo aspires to enable, but the concept in reality is often difficult to grasp. The create-aleo-app strives to present an example of how to bring zero knowledge computing to every-day web applications.

## Description
The create-aleo-app provides a template for creating zero knowledge apps via Aleo cryptography.

Users who use create-aleo-app will be able to receive templates to build the following
* Create and manage Aleo accounts
* Simple wallet using Aleo
* Simple zero knowledge applications 

## Concrete Tasks
- [x] #687
- [x] #688
- [x] #689
- [x] #690
- [x] #691

## Implementation
* Create-Aleo-App should be published to NPM and be able to be run via NPX

PR #704, opened 2023-08-04
[Feature] Create Aleo App alpha

### LINKED ISSUE #690: Create a package that can be created via `npm create` to create a basic web app using Aleo that has a similar structure to Aleo.tools
As part of the create-aleo-app, people should be able to use an `npm create` invocation to install the example versions of the Aleo web apps.

onetrickwolf:
## Motivation

Addresses #692, #691, #591

Resolves #690 

Based on vite's create package, modifying to also bring in some aspects of CRA since we need to bootstrap webpack for the workers for now. Probably limit this to one template until after we wrap the wasm with the SDK since a lot will probably change.

## Test Plan

```bash
cd create-aleo-app
npm run build
# in a new directory...
npx path/to/create-aleo-app
```

Go through prompts (only React template works right now) should create new and working Aleo web app.


onetrickwolf:
Prompts working, setting up working Aleo template still.

onetrickwolf:
Closing in favor of https://github.com/AleoHQ/sdk/pull/714

PR #705, opened 2023-08-05
[Spike] Conditional browser and node building

### LINKED ISSUE #689: Create conditional feature selection in the Aleo SDK that determines Node vs. Web environment

onetrickwolf:
## Motivation

Resolves #689 

Suggesting pattern for handling browser and node sdk builds in single package.

## Test Plan

`npm run build`

Should be able to see dist output created node and browser modules.


onetrickwolf:
I think there may be a better way to do this.

PR #733, opened 2023-09-05
[Feature] improving loading time by moving wasm load to after first paint

onetrickwolf:
## Motivation

WASM loading caused long loading spinner on slower connections and slower devices. This reverts to the dynamic import and adds a message and boolean to the hook allowing users to still use the site while the wasm loads in the background.

Should probably iterate on this to make the buttons actually disabled until the wasm is done loading but this is similar to what we had before but a little bit better since we have a loading indicator.

https://github.com/AleoHQ/sdk/assets/13836477/3a842ba4-6fb1-4cd8-9985-3f9fecf85a93

## Test Plan

Can throttle speed on browser to see loading message and confirm page is intractable before wasm is done loading.


website/src/components/WasmLoadingMessage.jsx
```diff
@@ -0,0 +1,28 @@
...
+export function WasmLoadingMessage() {
+    const { message, notification } = App.useApp();
+    const [_, aleoLoading] = useAleoWASM();
+
+    useEffect(() => {
+        if(aleoLoading){
+            message.open({
+                key: "wasmLoading",
+                type: 'loading',
+                content: 'Loading WASM...some functionality may not be available yet...',
```
iamalwaysuncomfortable:
```suggestion
                content: 'Loading... some functionality may not be available yet...',
```

website/src/components/WasmLoadingMessage.jsx
```diff
@@ -0,0 +1,28 @@
...
+                duration: 0,
+            });
+        } else {
+            message.open({
+                key: "wasmLoading",
+                type: 'success',
+                content: 'WASM Loaded!',
```
iamalwaysuncomfortable:
```suggestion
                content: 'Aleo.tools Loaded!',
```

website/src/components/WasmLoadingMessage.jsx
iamalwaysuncomfortable: I think the message about Wasm should be encapsulated, but otherwise great!

website/src/components/WasmLoadingMessage.jsx
iamalwaysuncomfortable: Same nit here

iamalwaysuncomfortable (approved):
Overall great - there's intermittent times where some components don't load - but this should improve the overall experience for anyone in a low bandwidth situation

collinc97 (approved):

onetrickwolf:
@iamalwaysuncomfortable Just pushed a commit that should fix intermittent loads!

PR #734, opened 2023-09-05
[Feature] SDK Worker rework

onetrickwolf:
## Motivation

Currently the website and create-aleo-app utilize the Aleo SDK / WASM via a worker instance and in an instance in the main thread. This PR aims to simplify the developer experience and optimize it a bit by having everything go through the worker and by having the worker managed by the SDK.

This PR also aims to do various house keeping by cleaning up some scripts and solidifying the build pipeline as well:

- [x] SDK bundles rather than building into source
- [x] Removing some old development preinstall scripts
- [x] ~Updating testing and build configuration to work automatically without having to comment lines out~ 
Some complications here but significantly simplified will iterate

#### 1. ~~All calls to the SDK will have to be asynchronous~~

~~This is already largely the case due to how most existing calls work and top level awaits are widely supported so should not be an issue.~~

#### 2. ~~Will require polyfill in node environments~~

~~We may do a dual build at first but we do have a polyfill now that should work for us once we remove some other browser specific dependencies from the WASM. This polyfill can be included in the SDK and won't affect developer experience.~~

When playing around with comlink and piping everything through the worker, calling constructors ended up being less ergonomic than just dealing with two instances of the WebAssembly. An example of the managed worker has been left in but will likely iterate on this.


onetrickwolf:
This isn't exactly where I wanted to go with this PR but I think it has some good examples we can iterate on and also cleans up quite a bit. Also shouldn't impact anything existing mostly just new additions. Should be good to review and then we can iterate 👍

sdk/src/obj_worker.ts
```diff
@@ -0,0 +1,10 @@
+import { expose } from 'comlink';
+
+const obj = {
+    counter: 0,
+    inc() {
+        this.counter++;
+    },
+};
+
+expose(obj);
```
iamalwaysuncomfortable: This is an alternative worker correct for illustrative purposes to show the usage of the workers? Per our last conversation - maybe we have a `workers` directory with the other worker file + one other worker focused maybe on record scanning and make it abundantly clear in the worker module docs that these are base patterns for using the worker as an api.

sdk/src/worker.ts
```diff
@@ -0,0 +1,145 @@
...
+keyProvider.useCache(true);
+
+// self.postMessage({
+//     type: "ALEO_WORKER_READY",
+// });
+
+export interface WorkerAPI {
+    additionalFunction1: () => Promise<string>;
+    additionalFunction2: (data: any) => Promise<any>;
+}
```
iamalwaysuncomfortable: Does this conflict with the definition immediately following/do we need this one?

sdk/src/worker.ts
```diff
@@ -0,0 +1,145 @@
...
+        privateKey: string
+    ) => Promise<{ outputs: any; execution: string } | string>;
+
+    getPrivateKey: () => Promise<PrivateKey>;
+}
+async function executeOffline(
+    localProgram: string,
+    aleoFunction: string,
+    inputs: string[],
+    privateKey: string
```
iamalwaysuncomfortable:
```suggestion
    privateKey: string
    proveExecution: boolean = false
```

We should have the prove execution here, default false and propagate that below

sdk/src/worker.ts
```diff
@@ -0,0 +1,145 @@
...
+        const keyParams = new AleoKeyProviderParams({
+            cacheKey: cacheKey,
+        });
+
+        // Execute the function locally
+        let response = await programManager.executeOffline(
+            localProgram,
+            aleoFunction,
+            inputs,
+            false,
```
iamalwaysuncomfortable: This should just be changed to proveExecution to provide the option to use it.

```suggestion
            proveExecution,
```

sdk/src/index.ts
```diff
@@ -33,22 +90,55 @@ import { Output } from "./models/output";
...
-import { Address, ExecutionResponse, PrivateKey, PrivateKeyCiphertext, Program, ProvingKey, RecordCiphertext, RecordPlaintext, Signature, Transaction as WasmTransaction, ViewKey, VerifyingKey, verifyFunctionExecution } from '@aleohq/nodejs';
-export { Account, Address, AleoKeyProvider, AleoKeyProviderParams, AleoKeyProviderInitParams, AleoNetworkClient, Block, BlockHeightSearch, CachedKeyPair, DevServerClient, Execution, ExecutionResponse, FunctionKeyPair, FunctionKeyProvider, Input, KeySearchParams, NetworkRecordProvider, PrivateKey, PrivateKeyCiphertext, Program, ProgramImports, ProvingKey, Output, RecordCiphertext, RecordPlaintext, RecordProvider, RecordSearchParams, Signature, Transaction, Transition, VerifyingKey, ViewKey, WasmTransaction, CREDITS_PROGRAM_KEYS, KEY_STORE, PRIVATE_TRANSFER, PRIVATE_TO_PUBLIC_TRANSFER, PRIVATE_TRANSFER_TYPES, PUBLIC_TRANSFER, PUBLIC_TO_PRIVATE_TRANSFER, VALID_TRANSFER_TYPES, logAndThrow, verifyFunctionExecution}
+// The following imports and exports are for a NodeJS context - if using the SDK in a browser context, delete or comment out this line
+export * from './node';
```
iamalwaysuncomfortable, line 144: Thanks for this change!

iamalwaysuncomfortable (changes requested):
A few very small nits + let's publish a new aleo.tools demo site with most PRs just so that we can quickly sanity check everytime. This pattern is a big step forward towards usability.

sdk/src/obj_worker.ts
onetrickwolf: Oops left this in by accident was me debugging. Will remove.

iamalwaysuncomfortable (approved):
PR Concerns addressed in the call - good to merge once integration testing passes

sdk/src/worker.ts
onetrickwolf: Oops sorry left this by accident too, removed.

PR #738, opened 2023-09-11
[Feature] Add Leo example to templates, other minor cleanup

onetrickwolf:
## Motivation

Adding a new template that shows using Leo and compiling an Aleo program directly in your Aleo app. Some other cleanup, simplifying the worker in this new template and various build improvements.

The new template shows a pattern where users could manage their own worker with comlink which simplifies things a bit compared to previous patterns but also is more flexible than the managed worker being included in the SDK.

Everything is forced through the worker to prevent two wasm instances but comes with some slight annoyances where any constructor has to be [proxied manually](https://github.com/AleoHQ/sdk/pull/738/files#diff-372749cb6a8a0894ae14b85a66b699361edec6e4c4ae040e92db3e42e60bf44aR41).

## Test Plan

`npm create aleo-app@latest`

Test new Leo template.

Can also install in the template directory directly.


iamalwaysuncomfortable (approved):
Looks great - one suggestion though is before merging this, add a little explanation in the create-aleo-app REadme.md about the different templates and how to install this new tempalte!

PR #753, opened 2023-09-26
[Fix] adding nonce to getFunctionInputs for records

### LINKED ISSUE #742: [Bug] Aleo.tools `execute` tab does not provide the "nonce" field for records
## 🐛 Bug Report

Aleo.tools' execute tab provides autogenerated inputs but it is missing the nonce field which is required for records. Add a nonce field for aleo.tools inputs so that records can be properly used via aleo.tools.

## Steps to Reproduce

1. Add credits.aleo to aleo.tools/develop under the execute tab
2. Open transfer_private
3. Notice there's no nonce

onetrickwolf:
## Motivation

Resolves #742 
## Test Plan

Load credits.aleo in aleo.tools and notice `_nonce` now shows up for functions with records (e.g., `transfer_private`)

### DIFF
--- a/wasm/src/programs/program.rs
+++ b/wasm/src/programs/program.rs
@@ -683,12 +683,12 @@ function add_and_double:
     fn test_get_inputs() {
         let credits = Program::from(ProgramNative::credits().unwrap());
         let inputs = credits.get_function_inputs("transfer_private".to_string()).unwrap();
-        let expected = r#"Array { obj: Object { obj: JsValue([Object({"type":"record","record":"credits","members":[{"name":"microcredits","type":"u64","visibility":"private"}],"register":"r0"}), Object({"type":"address","visibility":"private","register":"r1"}), Object({"type":"u64","visibility":"private","register":"r2"})]) } }"#.to_string();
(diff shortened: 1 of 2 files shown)


wasm/src/programs/program.rs
```diff
@@ -313,6 +313,15 @@ impl Program {
...
         Reflect::set(&input, &"members".into(), &record_members).map_err(|_| "Failed to set property")?;
+
+        // Adding _nonce object to record
+        let _nonce = Object::new();
+        Reflect::set(&_nonce, &"name".into(), &"_nonce".into()).map_err(|_| "Failed to set property")?;
+        Reflect::set(&_nonce, &"type".into(), &"group".into()).map_err(|_| "Failed to set property")?;
+        Reflect::set(&_nonce, &"visibility".into(), &"public".into()).map_err(|_| "Failed to set property")?;
+
+        Reflect::set(&input, &"_nonce".into(), &_nonce).map_err(|_| "Failed to set property")?;
+
```
iamalwaysuncomfortable: Per our discussion offline I think that you're right it would probably make more sense for this to be part of the members

iamalwaysuncomfortable (approved):
Hey this looks great. However, before merging could you publish a netifly build of it so we can check out that it works? Also, this is a critical enough of a change where we'll want to release a new version when it's finished.

onetrickwolf:
@iamalwaysuncomfortable added a test deploy as requested: https://nonce-fix--aleo-tools-3481ab.netlify.app/develop

And yeah I can do a version bump and release after the merge for the wasm and SDK if you want

PR #756, opened 2023-09-30
Converting to use rollup-plugin-rust instead of wasm-pack

Pauan:
## `@aleohq/wasm`

* The `@aleohq/wasm` package no longer uses `wasm-pack`, instead it now uses [`rollup-plugin-rust`](https://github.com/wasm-tool/rollup-plugin-rust/) in order to build the Wasm. This has a few advantages:

   * It is no longer necessary to manually install wasm-pack.
   * The build steps are now automated with Rollup, no more manual building.
   * The prepublish build step is no longer needed, snippets are handled automatically.
   * It gives more control over the `package.json`, the TypeScript types, and the public API.
   * It now exports a `worker.js` file which gives explicit control over the multi-threading.

* It no longer uses `wasm-bindgen-rayon` (which is fiddly and a bit buggy), instead it implements a custom thread-pool in the `wasm/src/thread_pool/mod.rs` file.

   This fixes various issues and also gives a lot more control over the multi-threading.

* The various Rust flags are now specified in the `wasm/rust-toolchain.toml` and `wasm/rollup.config.js` files.

* Publishing a new version is as easy as just doing `npm publish`, which builds the package and then publishes the new version.

## `@aleohq/nodejs`

* The `@aleohq/nodejs` package is no longer needed, instead we can use `@aleohq/wasm` for both the browser and NodeJS.

## `@aleohq/sdk`

* The SDK now works seamlessly in both the browser and NodeJS. The user API is identical in both environments.

   It only needs a single build to support both environments, any differences are fixed with polyfills (in the `sdk/polyfill` folder).

* The `initializeWasm` function no longer needs to be called, instead the Wasm is automatically initialized.

* Multi-threading (using the `initThreadPool` function) now works identically in both the browser and NodeJS. This gives a very significant performance boost in NodeJS.

* There is a `create-aleo-app/template-node` example for running the Aleo SDK directly in NodeJS.

* Publishing a new version is as easy as just doing `npm publish`, which builds the package and then publishes the new version.


create-aleo-app/template-node/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "node-starter",
+  "private": true,
+  "version": "0.0.0",
```
iamalwaysuncomfortable, line 4:
```suggestion
  "version": "0.1.0",
```

onetrickwolf (approved):
Will wait on @iamalwaysuncomfortable's review to merge since he's going deeper into the rust, but everything looks good from my perspective! Awesome job 👍

wasm/src/thread_pool/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+                window.navigator().hardware_concurrency() as usize
+            }),
+        )
+    }
+
+    // TODO this should cleanup the receiver when the ThreadPool is dropped
+    pub fn build_local(self) -> impl Future<Output = Result<rayon::ThreadPool, JsValue>> {
+        let (url, num_threads) = self.defaults();
+        spawn_local_thread_pool(url, num_threads)
+    }
```
iamalwaysuncomfortable, line 166: I like this approach. When this is cleaned up I assume the workers are also dropped?

wasm/js/worker.js
```diff
@@ -0,0 +1,31 @@
+import wasm from "../Cargo.toml";
+
+async function initializeWorker(wasm) {
```
iamalwaysuncomfortable, line 3: How does this one differ from the one defined inline in the Rust file below?

wasm/src/thread_pool/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+            }, {
+                capture: true,
+                once: true,
+            });
+
+            worker.postMessage({
+                module,
+                memory,
+                address,
+            });
```
iamalwaysuncomfortable, line 43: Who's the receiver of this post message again?

wasm/src/thread_pool/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+        });
+    }
+"###)]
+extern "C" {
+    #[wasm_bindgen(js_name = spawnWorker)]
+    fn spawn_worker(
+        url: &web_sys::Url,
+        module: &JsValue,
+        memory: &JsValue,
+        address: *const Receiver<ThreadBuilder>,
```
iamalwaysuncomfortable, line 53: This is the address of the pointer right?

sdk/src/node-polyfill.ts
```diff
@@ -0,0 +1,8 @@
+import "./polyfill/crypto";
+import "./polyfill/fetch";
+import "./polyfill/xmlhttprequest";
+import "./polyfill/worker";
+
+if (!globalThis.self) {
+    (globalThis as any).self = globalThis;
+}
```
iamalwaysuncomfortable, line 8: These polyfills are only ever bundled when the node version is built right?

sdk/rollup.config.js
```diff
@@ -0,0 +1,34 @@
+import typescript from "rollup-plugin-typescript2";
```
iamalwaysuncomfortable, line 1: Would this ever effect this library being embedded with bundlers like WebPack or Vite?

iamalwaysuncomfortable (commented):
Would love to merge this soon, couple of questions in the meantime

iamalwaysuncomfortable (approved):

wasm/src/thread_pool/mod.rs
Pauan: Currently no, but I have a TODO to fix that.

wasm/src/thread_pool/mod.rs
Pauan: The receiver is the `wasm/js/worker.js` file.

wasm/js/worker.js
Pauan: The `initializeWorker` Rust function accepts a pointer to the Rayon thread. It dereferences that pointer, and then runs the Rayon thread.

The `spawnWorker` JS function spawns the `wasm/js/worker.js` file in a new Worker, and then uses `postMessage` to send the pointer/memory/module:

https://github.com/AleoHQ/sdk/blob/eb8ef4ac36320015088256290457fab502246342/wasm/src/thread_pool/mod.rs#L38-L42

The `initializeWorker` JS function waits for the pointer/memory/module to be received:

https://github.com/AleoHQ/sdk/blob/eb8ef4ac36320015088256290457fab502246342/wasm/js/worker.js#L15-L18

And then after it receives those things it runs the Wasm module inside of the Worker, and calls the Rust `initializeWorker` with the pointer:

https://github.com/AleoHQ/sdk/blob/eb8ef4ac36320015088256290457fab502246342/wasm/js/worker.js#L26

So the Rust `initializeWorker` only runs the Rayon thread, whereas the JS `initializeWorker` manages the Worker, runs the Wasm, and then runs the Rayon thread.

Because it needs to run the Wasm, `wasm/js/worker.js` must be written in JS, it can't be written in Rust.

wasm/src/thread_pool/mod.rs
Pauan: Yes, it's a pointer to the Rayon thread. We can't send Rust objects to another Worker, but we can send a pointer (integer) to another Worker. So it sends the pointer to the Worker, so that way the Worker can dereference the pointer and access the Rayon thread.

sdk/src/node-polyfill.ts
Pauan: Yes, it's only used for Node, they aren't included at all for the browser.

sdk/rollup.config.js
Pauan: No, it's only used while building, the compiled code should work identically in every bundler (Webpack, Snowpack, Rollup, Parcel, etc.)

PR #764, opened 2023-10-05
[Feature] Bundle upgrades

### LINKED ISSUE #689: Create conditional feature selection in the Aleo SDK that determines Node vs. Web environment

onetrickwolf:
## Motivation

- [x] Fix wasm-pack CICD tests by disabling `"type": "module"` in package.json
- [x] Added template-node as cicd test
- [x] Bumped and release wasm and sdk to `0.6.0`
- [x] Update website to use latest

## Test Plan

WIP


onetrickwolf:
Changes cherry picked into https://github.com/AleoHQ/sdk/pull/765

PR #765, opened 2023-10-07
Adding in npm workspace

### LINKED ISSUE #689: Create conditional feature selection in the Aleo SDK that determines Node vs. Web environment

Pauan:
Sets up npm workspaces, which has several advantages:

* It's now very easy to test local changes without needing to manually `npm link` or `yarn link`.
* It's now possible to publish all of the npm packages with a single `npm publish` command in the root folder.
* It's now possible to use `npm test` in the root folder to run the unit tests for everything.
* It's now possible to use `npm run build:wasm` / `npm run build:sdk` / `npm run build:all` in the root folder to build the sub-packages.

In addition this PR has the following changes:

* Fixed running unit tests for the `@aleohq/wasm` package.
* Fixed CI to test the local `@aleohq/sdk` and `@aleohq/wasm` packages.
* Fixed the Node version of `@aleohq/sdk` so that it properly exits when the main thread is finished.
* The website is now built with the local version of `@aleohq/sdk` (instead of the published version).
* Updating the starter examples (and removing the unneeded React example).
* Adding in CI test for the `template-node` example.

Fixes #689


onetrickwolf (approved):
Great work! LGTM!

PR #772, opened 2023-10-12
Wasm Binary Size & Multi-Threading Optimization

iamalwaysuncomfortable:
## Motivation

The Wasm binary is currently large because of two main reasons:
1. Parameters are currently included as hardcoded in SnarkVM
2. Lots of inlining & generics within SnarkVM. In particular the `execution` pathway within SnarkVM involves a deep call stack and thus includes a very large amount of inlined monorphized functions.

Addittionally the `ProgramManager` struct included an optional process object allowing users to store keys in Wasm. However this can lead to unnecessary memory pressure and prevented any methods from being easily multi-threaded.

This PR 
1. Changes the `synthesizeKey` method to one that uses the existing execution_function paths
2. Removes the `wasm` cache from the `ProgramManager` to enable ProgramManager functions to be used concurrently within a Rayon threadpool (via WebWorkers)
3. Allows PrivateKeys to be passed by reference & removes `String` parameters in favor of `&str` parameters

Preview: https://wasm-optimization--aleo-dot-tools-latest.netlify.app/

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,4 @@
-{
+ {
   "name": "sdk-root",
   "private": true,
   "version": "0.0.0",
@@ -13,6 +13,9 @@
     "build:wasm": "cd wasm && yarn build",
     "build:sdk": "cd sdk && yarn build",
     "build:all": "yarn build:wasm && yarn build:sdk",
+    "build:website": "yarn && yarn build:all",
+
+    "start:website": "cd website && yarn dev",
 
     "test:wasm": "cd wasm && yarn test",
     "test:sdk": "cd sdk && yarn test",
--- a/package.json
+++ b/package.json
@@ -13,7 +13,6 @@
     "build:wasm": "cd wasm && yarn build",
     "build:sdk": "cd sdk && yarn build",
     "build:all": "yarn build:wasm && yarn build:sdk",
-    "build:website": "yarn && yarn build:all",
 
     "start:website": "cd website && yarn dev",
 
--- a/sdk/src/program-manager.ts
+++ b/sdk/src/program-manager.ts
@@ -515,7 +515,7 @@ class ProgramManager {
             );
             return [<VerifyingKey>keyPair.provingKey(), <ProvingKey>keyPair.verifyingKey()];
         } catch (e) {
-            throw logAndThrow("Could not synthesize keys - error {e}. Please ensure the program is valid and the inputs are correct.");
+            throw logAndThrow(`Could not synthesize keys - error ${e}. Please ensure the program is valid and the inputs are correct.`);
         }
     }
 
--- a/sdk/src/program-manager.ts
+++ b/sdk/src/program-manager.ts
@@ -31,7 +31,6 @@ class ProgramManager {
     host: string;
     networkClient: AleoNetworkClient;
     recordProvider: RecordProvider | undefined;
-    executionEngine: WasmProgramManager;
 
     /** Create a new instance of the ProgramManager
      *
@@ -54,7 +53,6 @@ class ProgramManager {
             this.keyProvider = keyProvider;
         }
 
-        this.executionEngine = new WasmProgramManager();
         this.recordProvider = recordProvider;
     }
 
@@ -186,7 +184,7 @@ class ProgramManager {
         }
 
         // Build a deployment transaction and submit it to the network
-        const tx = await this.executionEngine.buildDeploymentTransaction(deploymentPrivateKey, program, fee, feeRecord, this.host, false, imports, feeProvingKey, feeVerifyingKey);
+        const tx = await WasmProgramManager.buildDeploymentTransaction(deploymentPrivateKey, program, fee, feeRecord, this.host, imports, feeProvingKey, feeVerifyingKey);
         return await this.networkClient.submitTransaction(tx);
     }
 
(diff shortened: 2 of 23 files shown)


onetrickwolf (changes requested):
`website` execution is broken think you just need to update the worker to not use `executionEngine` like you did with the node template but otherwise looks good 👍

sdk/src/program-manager.ts
```diff
@@ -357,7 +355,7 @@ class ProgramManager {
         console.log("Running program offline")
         console.log("Proving key: ", provingKey);
         console.log("Verifying key: ", verifyingKey);
-        return this.executionEngine.executeFunctionOffline(executionPrivateKey, program, function_name, inputs, proveExecution,false, imports, provingKey, verifyingKey);
+        return WasmProgramManager.executeFunctionOffline(executionPrivateKey, program, function_name, inputs, proveExecution, false, imports, provingKey, verifyingKey);
```
Pauan, line 358: Shouldn't this remove the `false` cache argument as well?

wasm/src/programs/manager/execute.rs
```diff
@@ -44,11 +44,10 @@ impl ProgramManager {
     /// @param {boolean} prove_execution If true, the execution will be proven and an execution object
     /// containing the proof and the encrypted inputs and outputs needed to verify the proof offline
     /// will be returned.
-    /// @param {boolean} cache Cache the proving and verifying keys in the ProgramManager's memory.
-    /// If this is set to 'true' the keys synthesized (or passed in as optional parameters via the
-    /// `proving_key` and `verifying_key` arguments) will be stored in the ProgramManager's memory
-    /// and used for subsequent transactions. If this is set to 'false' the proving and verifying
-    /// keys will be deallocated from memory after the transaction is executed.
+    /// @param {boolean} cache Cache the proving and verifying keys in the Execution response.
```
Pauan, line 47: Shouldn't this also remove the cache?

wasm/src/programs/manager/execute.rs
```diff
@@ -57,10 +56,9 @@ impl ProgramManager {
...
-        &mut self,
-        private_key: PrivateKey,
-        program: String,
-        function: String,
+        private_key: &PrivateKey,
+        program: &str,
+        function: &str,
         inputs: Array,
         prove_execution: bool,
         cache: bool,
```
Pauan, line 64: And this should be removed too.

Pauan (changes requested):

sdk/src/program-manager.ts
iamalwaysuncomfortable: The old `synthesize_keypair` argument calling the `synthesize_kepair` pathway in SnarkVM created an a complicated monormrophization of the `execution_function` path which added ~1MB to the wasm binary size. The new pathway uses the execute_function_offline function which returns an optional object with the synthesized key in it within the `ExecutionResponse`

wasm/src/programs/manager/execute.rs
iamalwaysuncomfortable: No - reason described above ^^

wasm/src/programs/manager/execute.rs
iamalwaysuncomfortable: The cache is kept around in this function & this function only - reason described above.

sdk/src/program-manager.ts
Pauan: How does that affect monomorphization? The cache doesn't change the static type, and it's always passing `false` anyways.

iamalwaysuncomfortable: The cache within the `ProgramManager` had no effect on the `wasm` binary size - it was removed because it was ultimately a poor caching strategy. 

What had an effect on it was calling invoking the `synthesize_keypair` method from `SnarkVM` which imported a lot of generics we didn't use directly except for that function which had a huge effect on the size. In total removing it took 1MB off the size of the `wasm` binary.

However, SDK users would optimally still like to pre-synthesize keys for storage, so the work is delegated now to the offline execution function, and if `cache` is set to true, then they can get the keys from the function.

package.json
```diff
@@ -13,6 +13,9 @@
     "build:wasm": "cd wasm && yarn build",
     "build:sdk": "cd sdk && yarn build",
     "build:all": "yarn build:wasm && yarn build:sdk",
+    "build:website": "yarn && yarn build:all",
```
onetrickwolf: This wouldn't build the website I think you maybe wanted:

```cd website && yarn build```

**Review by onetrickwolf — APPROVED** [review]
LGTM 👍

File: sdk/src/program-manager.ts (hunk 1 above)
Comment (Pauan) [reply sdk/src/program-manager.ts:358]: I see, so you're saying that we *will* use the cache eventually, even though it's currently not being used?

File: package.json (hunk 4 above)
Comment (iamalwaysuncomfortable) [reply package.json]: I was mainly just thinking to have a command to pre-build the website since it might be difficult for newcomers to know the exact steps. We can update this in the documentation though.

File: sdk/src/program-manager.ts (hunk 1 above)
Comment (iamalwaysuncomfortable) [reply sdk/src/program-manager.ts:358]: Basically it puts the caching onto the implementor. It is being used now in the `synthesize_key` function.

File: package.json (hunk 4 above)
Comment (iamalwaysuncomfortable) [inline package.json]:
```suggestion
```

iamalwaysuncomfortable: Removed!

sdk/src/program-manager.ts
Pauan: Ahhh, I see, I had missed that.

package.json
Pauan: Brent is saying that it needs to be changed to this:

```json
"build:website": "cd website && yarn build",
```

Your script doesn't actually build the website at all, it only builds the SDK.

Pauan (approved):
LGTM

PR #774, opened 2023-10-15
fix: JsValue fetching error

hiromasa0629:
## Motivation

I was using @aleohq/sdk 0.6.2, trying to use `progamManager.execute` on a testdeploy program, but it wouldn't succeed. It would always show `JsValue fetching error...` after `executing fee`. Then I found that the error comes from the `globalThis.oldFetch`. Then I tried changing to the normal js fetch function, then it works.

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)

### DIFF
--- a/sdk/src/polyfill/fetch.ts
+++ b/sdk/src/polyfill/fetch.ts
@@ -28,6 +28,10 @@ const oldFetch = globalThis.fetch;
         });
 
     } else {
-        return await oldFetch(request);
+        try {
+            return await oldFetch(request);
+        } catch (e) {
+            return await fetch(request);
+        }
     }
 };


onetrickwolf:
@hiromasa0629 do you have an example of code that can replicate this error?

`globalThis.oldFetch` should be "normal fetch" in this case so your fix is basically just retrying the fetch after a failure. It's possible a network call is being made too quickly before the data is available on chain and the retry is resolving that.

Definitely something we should fix though thanks for reporting it! If you could give us some code to replicate the issue I'd be happy to get you a fix ASAP.

hiromasa0629:
@onetrickwolf Hi, sorry if the initial pr is too brief as I was actually in the hackathon

Below is the error 
```
Executing function: mint on-chain
Check program imports are valid and add them to the process
Executing program
parsing inputs
Loading program
Loading function
Adding program to the process
Creating authorization
Executing program
Preparing inclusion proofs for execution
Proving execution
Executing fee
Inserting externally provided fee proving and verifying keys
Authorizing Fee
Executing fee
error sending request: JsValue(TypeError: fetch failed
TypeError: fetch failed
    at Object.fetch (node:internal/deps/undici/undici:11118:11)
    at async globalThis.fetch (file:///Users/hiromasa/744-pr/node_modules/@aleohq/sdk/dist/node-polyfill.js:29:16))
```
And the code to replicate this
```js
import {
  Account,
  AleoKeyProvider,
  AleoNetworkClient,
  initThreadPool,
  NetworkRecordProvider,
  PrivateKey,
  ProgramManager,
} from "@aleohq/sdk";

await initThreadPool();

async function main() {
  const myAccount = new Account({
    privateKey: "Pirvate Key here",
  });

  const keyProvider = new AleoKeyProvider();
  keyProvider.useCache = true;

  const networkClient = new AleoNetworkClient("https://vm.aleo.org/api");
  const recordProvider = new NetworkRecordProvider(myAccount, networkClient);

  const programName = "testdeploy19923.aleo";
  const programManager = new ProgramManager(
    "https://vm.aleo.org/api",
    keyProvider,
    recordProvider
  );
  programManager.setAccount(myAccount);

  try {
    const tx_id = await programManager.execute(programName, "mint", 2, false, [
      "aleo public address here",
      "2000u64",
    ]);

    console.log(tx_id);
  } catch (e) {
    console.error(e);
  }
}

main();

```
The request object that is passed into oldFetch(request)
```
Request {
  [Symbol(realm)]: {
    settingsObject: { baseUrl: undefined, origin: [Getter], policyContainer: [Object] }
  },
  [Symbol(state)]: {
    method: 'GET',
    localURLsOnly: false,
    unsafeRequest: false,
    body: null,
    client: { baseUrl: undefined, origin: [Getter], policyContainer: [Object] },
    reservedClient: null,
    replacesClientId: '',
    window: 'client',
    keepalive: false,
    serviceWorkers: 'all',
    initiator: '',
    destination: '',
    priority: null,
    origin: 'client',
    policyContainer: 'client',
    referrer: 'client',
    referrerPolicy: '',
    mode: 'cors',
    useCORSPreflightFlag: false,
    credentials: 'same-origin',
    useCredentials: false,
    cache: 'default',
    redirect: 'follow',
    integrity: '',
    cryptoGraphicsNonceMetadata: '',
    parserMetadata: '',
    reloadNavigation: false,
    historyNavigation: false,
    userActivation: false,
    taintedOrigin: false,
    redirectCount: 0,
    responseTainting: 'basic',
    preventNoCacheCacheControlHeaderModification: false,
    done: false,
    timingAllowFailed: false,
    headersList: HeadersList {
      cookies: null,
      [Symbol(headers map)]: Map(0) {},
      [Symbol(headers map sorted)]: null
    },
    urlList: [ URL {} ],
    url: URL {
      href: 'http://vm.aleo.org/api/testnet3/latest/stateRoot',
      origin: 'http://vm.aleo.org',
      protocol: 'http:',
      username: '',
      password: '',
      host: 'vm.aleo.org',
      hostname: 'vm.aleo.org',
      port: '',
      pathname: '/api/testnet3/latest/stateRoot',
      search: '',
      searchParams: URLSearchParams {},
      hash: ''
    }
  },
  [Symbol(signal)]: AbortSignal { aborted: false },
  [Symbol(abortController)]: AbortController { signal: AbortSignal { aborted: false } },
  [Symbol(headers)]: HeadersList {
    cookies: null,
    [Symbol(headers map)]: Map(0) {},
    [Symbol(headers map sorted)]: null
  }
}
```

onetrickwolf:
@hiromasa0629 yeah so just confirming I think your code is fixing the issue because it's basically just retrying the call until it succeeds. I am guessing this either might be due to API instability or just the SDK trying to act on data faster than the network can reflect it. 

I think this is going to need a bit of a deeper fix where we are catching proper error messages and also backing off properly to ensure we aren't just infinitely retrying or we may just expose a helper method so users can implement their own retry at various network steps.

We'll hopefully be tackling this soon thank you so much for the detailed report and replication, I made an issue here to track: https://github.com/AleoHQ/sdk/issues/781

PR #802, opened 2023-11-05
Updating extension template to use the V3 manifest

Pauan:
The old extension template was designed for the V2 manifest. But V2 extensions are [deprecated](https://developer.chrome.com/docs/extensions/migrating/mv2-sunset/).

So this updates the extension template to the newer (and far worse) V3 manifest.

It is necessary to use an offscreen document for 2 reasons:

1. Service workers do not support top-level await.
2. Service workers cannot spawn Workers.

These are fundamental limitations of Chrome extensions, so there is no way for us to fix it.


create-aleo-app/template-extension/src/service_worker.js
```diff
@@ -0,0 +1,20 @@
...
+    });
+
+    if (existingContexts.length > 0) {
+        return;
+    }
+
+    await chrome.offscreen.createDocument({
+        url: offscreenUrl,
+        reasons: ["WORKERS"],
+        justification: "Top-level await and Workers cannot be used in service workers, the V3 extension system is horrendously awful.",
```
onetrickwolf: lol but maybe something a little more neutral?

```suggestion
        justification: "Enable usage of the Aleo WASM multi-threading via Workers.",
```

onetrickwolf (approved):
🥇

iamalwaysuncomfortable (approved):

PR #822, opened 2023-12-03
Template for React Aleo Hooks library

akalmykov:
## Motivation

We propose to add a new `create-aleo-app` template with Aleo Hooks library. 

React Hooks are functional components that isolate the reusable part from app business logic. Hooks make it easy to reuse stateful logic between components and develop complex dApps. 

Aleo Hooks is a library that offers a set of hooks for common operations, e.g. connecting a wallet, sending a transaction, fetching records. Aleo Hooks aims to be a wallet agnostic library that will simplify developing Aleo React dApps. In the first version, Leo wallet is supported. Support for the Puzzle wallet is planned next.

## Test Plan

There are two templates - for TS and JS. 
Exectuing:

```sh
npm create aleo-app@latest react-aleo-hooks-project -- --template=react-aleo-hooks
npm create aleo-app@latest react-aleo-hooks-ts-project -- --template=react-aleo-hooks-ts
```

should create two projects with sample apps that can be compiled and run.


iamalwaysuncomfortable (commented):
We're currently in network stability mode so don't have the allocation to give this the honest review it deserves, but we will try to prioritize this as soon as we can.

onetrickwolf:
@akalmykov very cool template! But this might be better as its own repo and linked in https://github.com/howardwu/awesome-aleo

Would be happy to link it in our documentation too but there might be easier to maintain it in a repo you control. If you think it should be here though open to hearing reasons!

PR #842, opened 2024-01-12
refactor: change inverted conditions

xxxigm:
Instead of using inverted conditions, you should strive to write code that is clear and readable by using original (non-negated) conditions and providing meaningful variable and function names. This helps readers of the code better understand the purpose and behavior of the code.
if (skip) {
// Do something when skip
} else {
// Do something when not skip
}


Arbuzer3000:
gm

PR #858, opened 2024-02-25
chore: use relative links

enjoyooor:
## Motivation

I feel relative links are way cleaner, just in case the org/repo changes which will impact the full URL.

Also fixed a few broken links to `create-aleo-app`

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,10 +30,10 @@ The official Aleo SDK providing Javascript/Typescript tools for creating zero kn
 
 ### ⚡ Build your own app
 
-Start here with the [Aleo SDK Readme](https://github.com/AleoHQ/sdk/tree/testnet3/sdk#readme) to get started building your 
+Start here with the [Aleo SDK Readme](/sdk/README.md) to get started building your 
 first zero knowledge web app.
 
-#### Source: [`sdk/sdk`](https://github.com/AleoHQ/sdk/tree/testnet3/sdk)
+#### Source: [`sdk/sdk`](/sdk)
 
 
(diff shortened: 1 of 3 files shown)


ultratwo:
Nice airdrop farming you got going there

Brand new account (post Starknet airdrop), contributing only to anticipated projects with airdrops : Aleo, L0, Kroma, Rabby etc...

Put your profile in private so we wouldn't spot it, but your repositories are public.

Please do not merge. Also I'll monitor every single of your future commits enjoyooor, no matter the repository

image

enjoyooor:
Yikes, people go to such lengths to deprive others instead of spending time building their skillset or contributing themselves. While your spending time on this, you should also comment nice things on my PRs that aren't as simple as these two that you've commented on 🤣 💀

Oh btw, you do realize that Aleo has already closed their contributions program, right? That's the case with half of the projects I've contributed to, they are just projects that I have interacted with in the past, so I'd like to contribute in any way possible however small it may seem. I start off with low hanging fruits/easy to fix bugs but if I can't find, I do simple patches. You should consider doing so as well instead of being salty and toxic, my friend :) The teams can be judge of my PRs and see if they are worth spending 15s reviewing or passing (spoiler alert: most are actually glad to take in these changes)

Lol4ikek:
hi

iamalwaysuncomfortable:
The readme has been updated since this, but appreciate the effort

PR #868, opened 2024-03-08
[Feature] Add address getter to Program

MaciejNadolski98:
`address()` returns the `Address` object associated with the `Program`

## Motivation

Currently there's no simple way to get the unique address of a program exposed to JS


iamalwaysuncomfortable (approved):
This is approved. Thanks for the community contribution.

PR #876, opened 2024-04-08
Adding in new X-Aleo-SDK-Version header

Pauan:
When the SDK makes a request to the server, this automatically adds in an `X-Aleo-SDK-Version` header, so the server can identify whether the request came from the SDK or not.

Fixes https://github.com/AleoHQ/explorer-monorepo/issues/14

### DIFF
--- a/sdk/package.json
+++ b/sdk/package.json
@@ -48,6 +48,7 @@
     "sync-request": "^6.1.0"
   },
   "devDependencies": {
+    "@rollup/plugin-replace": "^5.0.5",
     "@types/jest": "^29.5.5",
     "@types/mime": "^3.0.1",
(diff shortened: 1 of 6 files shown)


Pauan:
There are a few things that must be done before this is merged:

* The Wasm functions should also send the header
* The header should be customizable for `ProgramManager`
* Add in unit tests
* Somehow make the version not hardcoded

Pauan:
This is now ready for review:

* The Wasm functions call SnarkVM functions, so that can't be fixed in this PR.
* Making the header customizable is a breaking change that must be done in a different PR.
* I added in unit tests.
* The version is no longer hardcoded.

iamalwaysuncomfortable (approved):

PR #880, opened 2024-04-26
Vanilla Template Update

oguzutku1745:
## Motivation

The old Vanilla template lacked a deploy function. This new template has deploy function.

## Test Plan

A deploy function has been added to both the frontend and workers.

image

## Related PRs

No functional change, this is a part of template updates. For the next.js template, check following [PR](https://github.com/AleoHQ/sdk/pull/879)

### DIFF
--- a/create-aleo-app/template-vanilla/worker.js
+++ b/create-aleo-app/template-vanilla/worker.js
@@ -11,7 +11,7 @@ import {
   await initThreadPool();
   
   const hello_hello_program =
-      "program hello_hello_123.aleo;\n" +
+      "program hello_hello.aleo;\n" +
       "\n" +
       "function hello:\n" +
       "    input r0 as u32.public;\n" +
@@ -83,11 +83,12 @@ import {
           const result = getPrivateKey();
           postMessage(result);
       } else if (e.data === "deploy") {
-        deployProgram(hello_hello_program).then(result => {
+        try {
+            const result = await deployProgram(hello_hello_program);
             postMessage(result);
-        }).catch(error => {
+          } catch (error) {
             postMessage({ error: error.message });
-        });
+        }
     }
     
   };
--- a/create-aleo-app/template-vanilla/worker.js
+++ b/create-aleo-app/template-vanilla/worker.js
@@ -1,47 +1,94 @@
 import {
     Account,
-    initThreadPool,
-    PrivateKey,
     ProgramManager,
-} from "@aleohq/sdk";
-
-await initThreadPool();
-
-const hello_hello_program =`
-program hello_hello.aleo;
-
-function hello:
-    input r0 as u32.public;
-    input r1 as u32.private;
-    add r0 r1 into r2;
-    output r2 as u32.private;`
-
-async function localProgramExecution(program, aleoFunction, inputs) {
-    const programManager = new ProgramManager();
-
-    // Create a temporary account for the execution of the program
-    const account = new Account();
-    programManager.setAccount(account);
-
-    const executionResponse = await programManager.run(
-        program,
-        aleoFunction,
-        inputs,
-        false,
+    PrivateKey,
+    initThreadPool,
+    AleoKeyProvider,
+    AleoNetworkClient,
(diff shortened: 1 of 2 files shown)


create-aleo-app/template-vanilla/worker.js
```diff
@@ -1,47 +1,94 @@
...
+    PrivateKey,
+    initThreadPool,
+    AleoKeyProvider,
+    AleoNetworkClient,
+    NetworkRecordProvider,
+  } from "@aleohq/sdk";
+  
+  await initThreadPool();
+  
+  const hello_hello_program =
```
Pauan: What is the reason for changing the string?

create-aleo-app/template-vanilla/worker.js
```diff
@@ -1,47 +1,94 @@
...
+  
+  async function localProgramExecution(program, aleoFunction, inputs) {
+      const programManager = new ProgramManager();
+  
+      // Create a temporary account for the execution of the program
+      const account = new Account();
+      programManager.setAccount(account);
+  
+      const executionResponse = await programManager.run(
+          hello_hello_program,
```
Pauan: Why this change?

create-aleo-app/template-vanilla/worker.js
```diff
@@ -1,47 +1,94 @@
...
+  
+  onmessage = async function (e) {
+      if (e.data === "execute") {
+          const result = await localProgramExecution();
+          postMessage(result);
+      } else if (e.data === "key") {
+          const result = getPrivateKey();
+          postMessage(result);
+      } else if (e.data === "deploy") {
+        deployProgram(hello_hello_program).then(result => {
```
Pauan: It would be better to write this using `await` syntax:

```js
        try {
            const result = await deployProgram(hello_hello_program);
            postMessage(result);
        } catch (error) {
            postMessage({ error: error.message });
        }
```

Pauan (changes requested):

create-aleo-app/template-vanilla/worker.js
oguzutku1745: I changed the file name to be unique in order to test the deploy function. It looks like I forgot to change it back. Thanks for pointing this out; I'm updating it immediately!

create-aleo-app/template-vanilla/worker.js
oguzutku1745: The function `executeOffline` has been renamed to "run" with [this commit](https://github.com/AleoHQ/sdk/commit/45ac2ad7645499e8efbf3e76e421e4233308c9dd#diff-d5500ae2c32e9410d2a9a44bb20b6daecae531f19c3a01e1e9c8e3c90e5f21bb). That is the reason for the change.

create-aleo-app/template-vanilla/worker.js
oguzutku1745: Updating this now, thanks!

create-aleo-app/template-vanilla/worker.js
Pauan: Thanks, but why did you change the way that the string is created?

create-aleo-app/template-vanilla/worker.js
Pauan: What I mean is, the old code passed it in as arguments, but now you're hardcoding it.

create-aleo-app/template-vanilla/worker.js
oguzutku1745: My bad, I tested the templates with both versions of this string. I thought there was no difference between the two and pushed with this one, but the older version is definitely clearer. Updated

create-aleo-app/template-vanilla/worker.js
oguzutku1745: Yes, that was a mistake I didn't realize. It seems I was too focused on deploying functionality. I've corrected it now

Pauan (approved):
Looks good, thanks!

PR #881, opened 2024-04-27
Adding in thread_pool::spawn function

Pauan:
This adds in a `thread_pool::spawn` function which can be used to run some Rust code on multiple Workers for multithreading.

This is the first step toward implementing #790


iamalwaysuncomfortable (approved):

iamalwaysuncomfortable:
This looks good to merge, one thing we will want however is to ensure we're writing some integration tests around spawning functions onto the threadpool.

Pauan:
Absolutely, that will be in the next PR that actually makes use of the spawning.

PR #884, opened 2024-05-16
Upgrading versions of snarkvm

Pauan:
This upgrades the version of snarkvm to the `mainnet-staging` branch.

It still uses the testnet, it does not use the mainnet. I will add in mainnet capabilities in a future PR.

Most of the changes are trivial (renaming imports, etc.) but the `estimate_execution_fee` and `estimate_finalize_fee` functions required more significant changes, so they should be reviewed more carefully.


.github/workflows/website.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly-2023-05-24
+          toolchain: nightly-2024-05-14
```
iamalwaysuncomfortable, line 21: I think this might be able to go to `nightly-latest`? Opinions?

wasm/Cargo.toml
```diff
@@ -22,33 +22,49 @@ crate-type = [ "cdylib", "rlib" ]
 doctest = false
 
 [dependencies.snarkvm-circuit-network]
-version = "=0.16.12"
+version = "0.16.19"
+git = "https://github.com/AleoNet/snarkVM.git"
+rev = "3ebe60c499285a140ec2171fe8dd06662c325531"
```
iamalwaysuncomfortable: Current Rev is: 140ff26 (I know this changed from under us since this process started)

wasm/src/programs/program.rs
```diff
@@ -553,33 +573,46 @@ function add_and_double:
...
-        Reflect::set(&committee, &JsValue::from_str("value_type"), &JsValue::from_str("committee_state")).unwrap();
-        Reflect::set(&unbonding, &JsValue::from_str("name"), &JsValue::from_str("unbonding")).unwrap();
-        Reflect::set(&unbonding, &JsValue::from_str("key_type"), &JsValue::from_str("address")).unwrap();
-        Reflect::set(&unbonding, &JsValue::from_str("value_type"), &JsValue::from_str("unbond_state")).unwrap();
-
-        array.push(&committee);
-        array.push(&bonded);
-        array.push(&unbonding);
-        array.push(&account);
+        let expected = array![
```
iamalwaysuncomfortable, line 576: Much cleaner thank you!

wasm/src/programs/verifying_key/credits.rs
```diff
@@ -134,15 +134,15 @@ impl VerifyingKey {
...
+        VerifyingKey::from_bytes(&snarkvm_parameters::testnet::UnbondPublicVerifier::load_bytes().unwrap()).unwrap()
     }
 
     /// Returns the verifying key for the bond_public function
     ///
     /// @returns {VerifyingKey} Verifying key for the bond_public function
     #[wasm_bindgen(js_name = "isBondPublicVerifier")]
     pub fn is_bond_public_verifier(&self) -> bool {
-        self == &VerifyingKey::from_bytes(&snarkvm_parameters::testnet3::BondPublicVerifier::load_bytes().unwrap())
+        self == &VerifyingKey::from_bytes(&snarkvm_parameters::testnet::BondPublicVerifier::load_bytes().unwrap())
```
iamalwaysuncomfortable, line 145: This is a good start, though we'll need to support mainnet soon as well. I'm fine focusing on only testnet for now and adding mainnet support in a future PR

iamalwaysuncomfortable (commented):
Had a few comments, but overall these look good, I'll check on the VerifyingKey issue you flagged

wasm/src/programs/verifying_key/credits.rs
Pauan: Yes, I will support mainnet in another PR, that's a much bigger change.

wasm/Cargo.toml
Pauan: Absolutely, I will update it before this gets merged.

.github/workflows/website.yml
Pauan: I always lock rustc to a specific version, so that way it's reproducible and consistent for everybody. If it's `nightly-latest` then it can cause breakage.

Note that the CI matches the version in `rust-toolchain.yaml`, so that way the CI will match local dev builds.

Pauan: Also note that locking the version when using nightly is [standard best practice](https://www.justanotherdot.com/posts/how-to-pick-stable-beta-or-nightly-rust).

iamalwaysuncomfortable: Agreed

sdk/src/index.ts
```diff
@@ -1,91 +1,45 @@
...
-        verifier: "unbond_delegator_as_validator.verifier.9585609",
-        verifyingKey: VerifyingKey.unbondDelegatorAsValidatorVerifier
-    },
-    unbond_public: {
-        locator: "credits.aleo/unbond_public",
-        prover: KEY_STORE + "unbond_public.prover.9547c05",
-        verifier: "unbond_public.verifier.09873cd",
-        verifyingKey: VerifyingKey.unbondPublicVerifier
-    },
+    bond_public: convert(Metadata.bond_public()),
```
iamalwaysuncomfortable, line 29: I like this approach

iamalwaysuncomfortable (approved):
This is good to go in. We'll need to update this to the very latest revision, but we can do that in a follow-on PR.

PR #891, opened 2024-07-01
[Update] Update SDK methods to support latest `credits.aleo` method signatures

iamalwaysuncomfortable:
## Motivation

The latest SnarkVM revision published on the [AleoNet/SnarkVM repo](https://github.com/AleoNet/snarkVM/commit/d170a9f7c5ef980f9392301dc899dee355599ca6) (rev: d170a9f) has updated the `credits.aleo` program to support [ARC-41](https://github.com/AleoNet/ARCs/discussions/72). 

This PR updates the Aleo-wasm snarkVM rev to `d170a9f` and updates all method signatures in the SDK to match the latest method signatures in `credits.aleo`, it also adds the `bond_validator` and `transfer_public_as_signer` methods to the SDK's api. 

The work in this PR is meant to represent a release candidate for the SDK that is updated to work with the currently released Aleo Testnet.

## Test Plan

All unit tests are now updated to reflect the latest changes.

## Related PRs

The work in this PR is a continuation of the work done in #884.


sdk/src/offline-key-provider.ts
```diff
@@ -106,21 +126,21 @@ class OfflineSearchParams implements KeySearchParams {
     }
 
     /**
-     * Create a new OfflineSearchParams instance for the transfer_public_to_private function of the credits.aleo program.
+     * Create a new OfflineSearchParams instance for the transfer_public_as_signer function of the credits.aleo program.
      */
-    static transferPublicToPrivateKeyParams(): OfflineSearchParams {
-        return new OfflineSearchParams(CREDITS_PROGRAM_KEYS.transfer_public_to_private.locator, true);
+    static transferPublicAsSignerKeyParams(): OfflineSearchParams {
```
Pauan, line 131: Isn't this a breaking change? What is our policy for that? Release with a new major version?

sdk/src/program-manager.ts
```diff
@@ -898,41 +1027,53 @@ class ProgramManager {
...
+     *
+     * // Create the claim unbonded transaction object for later use.
+     * const tx = await programManager.buildClaimUnbondPublicTransaction("aleo1jx8s4dvjepculny4wfrzwyhs3tlyv65r58ns3g6q2gm2esh7ps8sqy9s5j");
+     * console.log(tx);
+     *
+     * // The transaction can be submitted later to the network using the network client.
+     * programManager.networkClient.submitTransaction(tx);
      */
-    async buildClaimUnbondPublicTransaction(options: Partial<ExecuteOptions> = {}): Promise<Transaction | Error> {
+    async buildClaimUnbondPublicTransaction(staker_address: string, options: Partial<ExecuteOptions> = {}): Promise<Transaction | Error> {
```
Pauan, line 1070: Why is this a separate argument instead of being a field inside of `ExecuteOptions`?

wasm/src/programs/manager/transfer.rs
```diff
@@ -113,10 +114,17 @@ impl ProgramManager {
                 ("transfer_private_to_public", inputs)
             }
             "public" | "transfer_public" | "transferPublic" => {
+                let inputs = Array::new_with_length(3);
```
Pauan: This can be better written like this:

```rust
let inputs = [
    JsValue::from(&caller.unwrap()),
    JsValue::from(recipient),
    JsValue::from(&amount_microcredits.to_string().add("u64")),
].into_iter().collect::<js_sys::Array>();
```

create-aleo-app/template-offline-public-transaction-ts/src/index.ts
```diff
@@ -93,7 +93,9 @@ async function buildBondingTxOffline(recipientAddress: Address, amount: number,
     }
 
     const bondTx = <Transaction>await programManager.buildBondPublicTransaction(
-        recipientAddress.to_string(),
+        stakerAddress.to_string(),
```
Pauan, line 96: Since we're making breaking changes anyways, I think it would be nice if it used an options object, instead of using multiple arguments.

wasm/src/programs/verifying_key/credits.rs
```diff
@@ -26,249 +26,264 @@ impl VerifyingKey {
         VerifyingKey::from_bytes(&snarkvm_parameters::testnet::BondPublicVerifier::load_bytes().unwrap()).unwrap()
     }
 
+    /// Returns the verifying key for the bond_validator function
+    ///
+    /// @returns {VerifyingKey} Verifying key for the bond_validator function
+    #[wasm_bindgen(js_name = "bondValidatorVerifier")]
+    pub fn bond_validator_verifier() -> VerifyingKey {
+        let vk = CurrentNetwork::get_credits_verifying_key("bond_validator".to_string()).unwrap().clone();
```
Pauan, line 34: I'm seeing a lot of code duplication in these methods, it's probably worthwhile to create a helper method to make the code shorter.

wasm/src/programs/verifying_key/credits.rs
```diff
@@ -26,249 +26,264 @@ impl VerifyingKey {
...
+    }
+
     /// Returns the verifying key for the claim_delegator function
     ///
     /// @returns {VerifyingKey} Verifying key for the claim_unbond_public function
     #[wasm_bindgen(js_name = "claimUnbondPublicVerifier")]
     pub fn claim_unbond_public_verifier() -> VerifyingKey {
-        VerifyingKey::from_bytes(&snarkvm_parameters::testnet::ClaimUnbondPublicVerifier::load_bytes().unwrap())
-            .unwrap()
+        let vk = CurrentNetwork::get_credits_verifying_key("bond_validator".to_string()).unwrap().clone();
```
Pauan, line 44: Why is it now using hardcoded strings, whereas before it used methods (which are statically typed and safer)?

wasm/src/programs/verifying_key/credits.rs
```diff
@@ -26,249 +26,264 @@ impl VerifyingKey {
...
     }
 
     /// Returns the verifying key for the bond_public function
     ///
     /// @returns {VerifyingKey} Verifying key for the bond_public function
     #[wasm_bindgen(js_name = "isBondPublicVerifier")]
     pub fn is_bond_public_verifier(&self) -> bool {
-        self == &VerifyingKey::from_bytes(&snarkvm_parameters::testnet::BondPublicVerifier::load_bytes().unwrap())
-            .unwrap()
+        self == &Self::bond_public_verifier()
```
Pauan, line 174: This is a nice improvement.

Pauan (dismissed):

sdk/src/offline-key-provider.ts
iamalwaysuncomfortable: In the sense of usual software yes, but the SDK tracks API changes of the Aleo network which will likely start major version updates post mainnet, so I'd like to continue to track the Network's conventions here. However, it does deserve at least a minor version upgrade.

create-aleo-app/template-offline-public-transaction-ts/src/index.ts
iamalwaysuncomfortable: I agree, but don't currently have the allocation to do so? Would you be able to make that change? Or maybe @onetrickwolf?

wasm/src/programs/verifying_key/credits.rs
iamalwaysuncomfortable: I agree, just don't have the allocation to clean it up right now, I'm open to having others make the changes if you have the allocation.

wasm/src/programs/verifying_key/credits.rs
iamalwaysuncomfortable: Because the VerifyingKeys in SnarkVM are for different object than the verifying key (something called the CircuitVerifyingKey, which VerifyingKey is based off of). Loading it directly from bytes fails (feel free to try it and confirm).

However this method requires the actual VerifyingKey object to be accepted for Executions and thus this uses the convenience method in the Network object to do this.

wasm/src/programs/verifying_key/credits.rs
iamalwaysuncomfortable: Thanks!

sdk/src/program-manager.ts
iamalwaysuncomfortable: The emphasize which address is being used. In general most of the methods that mirror credits.aleo have the parameters specified outside of the ExecutionOptions so that it fits the same form factor as the Aleo Instructions/Leo methods. For all credits.aleo transactions it's extremely important to get inputs right, so they're structured this way to force the caller to consider the options properly/fit the form factor of Aleo instructions.

sdk/src/offline-key-provider.ts
Pauan: The SDK is currently on version `0.6.9`. So a major version bump would mean going to `0.7.0`.

This is because the `0.x.x` versions are special, and so a minor version bump is treated the same as a major version bump:

https://www.npmjs.com/package/semver

> Many authors treat a `0.x` version as if the `x` were the major "breaking-change" indicator.
>
> Caret ranges are ideal when an author may make breaking changes between `0.2.4` and `0.3.0` releases, which is a common practice. However, it presumes that there will not be breaking changes between `0.2.4` and `0.2.5`.

create-aleo-app/template-offline-public-transaction-ts/src/index.ts
Pauan: Sure, I can do that in another PR.

wasm/src/programs/verifying_key/credits.rs
Pauan: Sure, that's fine, we can do it later.

sdk/src/program-manager.ts
Pauan: In general I think we should move toward an API where everything is specified with options objects. That makes it *much* easier to add new fields over time, and also makes the API clearer for users.

But that can be done in a future PR, so I'll accept this for now.

wasm/src/programs/manager/transfer.rs
iamalwaysuncomfortable: Great suggestion, changed.

create-aleo-app/template-offline-public-transaction-ts/src/index.ts
iamalwaysuncomfortable: In the case of the method that mirror credits.aleo, I'd like to keep this as is so that users ensure that users are considering the arguments going into Aleo instructions.

sdk/src/offline-key-provider.ts
iamalwaysuncomfortable: Ah yes, you're right, a major version bump is in order. It was bumped in the latest commit.

Pauan (approved):
LGTM!

PR #898, opened 2024-07-12
remove old header image until we have new branding

onetrickwolf:
## Motivation

Removing old Aleo branding until we have path forward for Provable branding.


iamalwaysuncomfortable (approved):

jaketarnow (approved):
lgtm!

Issue #892, opened 2024-07-03
[Bug] Block data parsing loses precision and typing of bigInts


WietzeSlagman:
## 🐛 Bug Report

Using `response.json()` as a parser for results fetched from Aleo nodes removes precision from bigInts and transforms them into strings (leading to users needing to handle casting and handling them correctly).

## Steps to Reproduce

get a block containing a solution generated by a prover i.e block 4844 on testnet-beta, and check the values returned in the `solutions` part of the block, the result is that the `counter` element in the `partial_solutions` has lost the last 4 digits in precision and is now unexpectedly a string.

#### Code snippet to reproduce

```
import { AleoNetworkClient } from "sdk"
const networkClient = new AleoNetworkClient(ALEO_NODE_ENDPOINT)

const block = await networkClient.getBlock(4844)
```
Check the results of the blocks and verify wrong types: `counter: "4175780142384957400"` but it should be `counter: 4175780142384957663` (as a bigInt)

#### Stack trace & error message

no error message to be shown

## Expected Behavior

Expected behaviour is for the blocks to be returned with the right precision and having the correct typing such as bigInt allowing clients using the SDK.

## Your Environment

- node 20 alpine

WietzeSlagman:
Proposed fix
Use [json-bigint](https://www.npmjs.com/package/json-bigint) to parse the response data instead of using `response.json()` in `fetchData`.
No PR for this atm as I do not have the full understanding of the impact of this throughout the whole SDK, however it should address the data issue.

iamalwaysuncomfortable:
@WietzeSlagman  Thanks for noting this. Would you be interested in making the PR to fix it? 

I'll ensure the SDK maintainers are aware of this.

WietzeSlagman:
> @WietzeSlagman Thanks for noting this. Would you be interested in making the PR to fix it?
> 
> I'll ensure the SDK maintainers are aware of this.

Yes i can make a PR with a fix for this, i'll have something ready within a day or 2

WietzeSlagman:
PR opened: #901

PR #901, opened 2024-07-18
Ensure number precision on BigInts

WietzeSlagman:
## Motivation

Solves #892 by utilizing `json-bigint` to parse incoming block data, allowing the integers that are too big in the web world to keep their precision and be workeable by developers.

## Test Plan

Ran `yarn test` in `sdk` directory, all passed

Ran webapp using the updated SDK to verify block data and the following is observed fetching block `327674` where the target in a proposed solution is a known bigInt.

image

As shown in the above image several elements are now transformed into bigInts and have the correct precision when translated to forinstance a string. which looks like this on the solution: `block.solutions.solutions.solutions[1].target.toString() = "2282730364275405"`
The following is returned from calling a node directly: `"target": 2282730364275405` (as part of the solution). Which shows that the value now has the correct precision and the developer can use the bigInt to make calculations with.

## Related PRs

N/A

### DIFF
--- a/sdk/package.json
+++ b/sdk/package.json
@@ -42,14 +42,16 @@
   },
   "homepage": "https://github.com/ProvableHQ/sdk#readme",
   "dependencies": {
-    "@provablehq/wasm": "^0.6.0",
+    "@provablehq/wasm": "^0.6.10",
     "comlink": "^4.4.1",
+    "json-bigint": "^1.0.0",
     "mime": "^3.0.0",
     "sync-request": "^6.1.0"
   },
   "devDependencies": {
     "@rollup/plugin-replace": "^5.0.5",
     "@types/jest": "^29.5.5",
+    "@types/json-bigint": "^1.0.4",
     "@types/mime": "^3.0.1",
     "@typescript-eslint/eslint-plugin": "^5.41.0",
     "@typescript-eslint/parser": "^5.41.0",
--- a/sdk/src/network-client.ts
+++ b/sdk/src/network-client.ts
@@ -10,6 +10,7 @@ import {
   TransactionModel,
   logAndThrow
 } from "./index";
+import JSONbig from "json-bigint";
 
 type ProgramImports = { [key: string]: string | Program };
 
@@ -88,7 +89,8 @@ class AleoNetworkClient {
       headers: this.headers
     });
 
-    return await response.json();
+    const responseText = await response.text()
+    return await JSONbig.parse(responseText);
     } catch (error) {
       throw new Error("Error fetching data.");
     }
@@ -650,7 +652,8 @@ class AleoNetworkClient {
       });
 
       try {
-        return await response.json();
+        const responseText = await response.text()
+        return await JSONbig.parse(responseText);
 
       } catch (error) {
         throw new Error(`Error posting transaction. Aleo network response: ${(error as Error).message}`);
--- a/yarn.lock
+++ b/yarn.lock
@@ -94,7 +94,7 @@
   resolved "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.23.2.tgz"
   integrity sha512-0S9TQMmDHlqAZ2ITT95irXKfxN9bncq8ZCoJhun3nHL/lLUxd2NKBJYoNGWH7S0hz6fRQwWlAWn/ILM0C70KZQ==
 
-"@babel/core@^7.0.0", "@babel/core@^7.0.0-0", "@babel/core@^7.0.0-0 || ^8.0.0-0 <8.0.0", "@babel/core@^7.11.6", "@babel/core@^7.12.0", "@babel/core@^7.12.3", "@babel/core@^7.13.0", "@babel/core@^7.17.2", "@babel/core@^7.22.20", "@babel/core@^7.22.9", "@babel/core@^7.4.0 || ^8.0.0-0 <8.0.0", "@babel/core@^7.7.5", "@babel/core@^7.8.0", "@babel/core@>=7.0.0-beta.0 <8":
+"@babel/core@^7.11.6", "@babel/core@^7.12.3", "@babel/core@^7.17.2", "@babel/core@^7.22.20", "@babel/core@^7.22.9", "@babel/core@^7.7.5":
   version "7.23.2"
   resolved "https://registry.npmjs.org/@babel/core/-/core-7.23.2.tgz"
   integrity sha512-n7s51eWdaWZ3vGT2tD4T7J6eJs3QoBXydv7vkUM06Bf1cbVD2Kc2UrkzhiQwobfV7NwOnQXYL7UBJ5VPU+RGoQ==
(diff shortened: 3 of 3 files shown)


WietzeSlagman:
There is one point in this proposed solution that could be deemed not ideal: Several numbers are typed as for instance `u64` in snarkVM which means they could become a too big number for the web to handle. However if the value is not yet too big to handle it is returned as a number, For example in the screenshot the `target` of `solution[0]` is parsed as a regular number (because it is not big enough to be a bigInt) however the `target` of `solution[1]` is. This leads to inconsistencies for devs and would lead to making extra checks to see which kind of operations they can do and how to parse values to display them correctly.
To solve this each known bigInt should be parsed to a bigInt irregardles of its current size, allowing for the most consistency in terms of experience. 

This issue i think should be solved in a seperate PR and get this in before, this ensures that developers are atleast working with the correct percision and the data is matching that which is present in the actual block.

Pauan:
Thanks for the PR!

There's a few issues I see with using json-bigint:

* Performance. JSON parsing is pretty slow, and I worry that json-bigint will make it even slower.

* json-bigint includes a lot of extra code that we don't need, bloating up the code size.

* As you mentioned, sometimes numbers are returned as bigint and sometimes they're not... that inconsistency needs to be fixed.

* This is a breaking change, since existing SDK users will be expecting `typeof "number"`, not `typeof "bigint"`.

Ideally we'd fix this on the SnarkOS side, so that the JSON uses strings for large numbers... but it's hard to change things in SnarkOS, so we might need a temporary stop-gap in the SDK...

It might be worthwhile to create a custom JSON parser that fixes the above issues.

Pauan:
Note that because this is a breaking change, it will need to be done in the next major version bump, which should be soon! So we'll make sure to prioritize this issue to get it in for the next major version bump.

Pauan:
I found this article about using native `JSON.parse` to parse BigInt:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON#using_json_numbers

```js
JSON.parse(str, (key, value, context) => {
    if (Number.isInteger(value)) {
        return BigInt(context.source);

    } else {
        return value;
    }
})
```

This will cause all integers to be parsed as a bigint. We can also easily adjust it so that it only parses certain keys as bigint.

This seems preferable to using `json-bigint`, though as a downside it does parse the number twice.

WietzeSlagman:
> Thanks for the PR!
> 
> There's a few issues I see with using json-bigint:
> 
> * Performance. JSON parsing is pretty slow, and I worry that json-bigint will make it even slower.
> * json-bigint includes a lot of extra code that we don't need, bloating up the code size.
> * As you mentioned, sometimes numbers are returned as bigint and sometimes they're not... that inconsistency needs to be fixed.
> * This is a breaking change, since existing SDK users will be expecting `typeof "number"`, not `typeof "bigint"`.
> 
> Ideally we'd fix this on the SnarkOS side, so that the JSON uses strings for large numbers... but it's hard to change things in SnarkOS, so we might need a temporary stop-gap in the SDK...
> 
> It might be worthwhile to create a custom JSON parser that fixes the above issues.

Thanks for the reply. i do agree it is some additional overhead but i wonder how big that impact actually is (in our applications i havent noticed a real impact yet).

This is indeed a breaking change that would require code updates for users so makes sense to put it in in a later date, however getting this fix to users quickly will help integrate them faster.

On your point of adding it to SnarkOS i am not sure if that is an ideal way to go, especially seeing that is an even bigger impact for users (as it not only affects JS/TS users but also Rust) as well as that several languages do handle this fine without needing the additional parse, this means you are making the experience and performance worse accross the board unnecesarily. The only way i would see this work in SnarkOS is if you would be able to provide a flag of how you want the data but that seems like a lat of engineering effort for little gain.

Pauan:
> On your point of adding it to SnarkOS i am not sure if that is an ideal way to go, especially seeing that is an even bigger impact for users (as it not only affects JS/TS users but also Rust) as well as that several languages do handle this fine without needing the additional parse

Do you know of any languages that parse JSON numbers as a bigint by default?

WietzeSlagman:
> > On your point of adding it to SnarkOS i am not sure if that is an ideal way to go, especially seeing that is an even bigger impact for users (as it not only affects JS/TS users but also Rust) as well as that several languages do handle this fine without needing the additional parse
> 
> Do you know of any languages that parse JSON numbers as a bigint by default?

bigInt as a form is (almost) exclusively used within web and JS/TS frameworks, other languages would I assume parse it to either INT64 (or 32) or UINT if provided. It of course highly depends on the language implementation and packages used. My point with this is that most lower level languages have built in types for larger numbers and do not need something like a bigInt to be defined.

Pauan:
> bigInt as a form is (almost) exclusively used within web and JS/TS frameworks

Bigint originated outside of JavaScript, all the way back in the 1960s, it was only very recently that JavaScript received bigint support.

> other languages would I assume parse it to either INT64 (or 32) or UINT if provided.

I would assume that most languages would parse it as 64-bit floating point, just like how JavaScript does.

I'm unaware of any languages which parse JSON numbers as integers, which is why I asked if you knew of any examples.

> My point with this is that most lower level languages have built in types for larger numbers and do not need something like a bigInt to be defined.

Regardless of whether the language supports larger numbers or not, if the *JSON parser* for the language uses 64-bit floating points for JSON numbers, then it will have the exact same precision loss issues as JavaScript.

That's why it's common practice for APIs to return large numbers as strings, so that way it won't accidentally be parsed as a 64-bit floating point number.

WietzeSlagman:
> > other languages would I assume parse it to either INT64 (or 32) or UINT if provided.
> 
> I would assume that most languages would parse it as 64-bit floating point, just like how JavaScript does.
> 
> I'm unaware of any languages which parse JSON numbers as integers, which is why I asked if you knew of any examples.
> 
> > My point with this is that most lower level languages have built in types for larger numbers and do not need something like a bigInt to be defined.
> 
> Regardless of whether the language supports larger numbers or not, if the _JSON parser_ for the language uses 64-bit floating points for JSON numbers, then it will have the exact same precision loss issues as JavaScript.
> 
> That's why it's common practice for APIs to return large numbers as strings, so that way it won't accidentally be parsed as a 64-bit floating point number.

You are correct that it al heavily depends on the JSON parser for any language implementation, I have no insights into the inter workings of these parsers and how they handles them.

Do note I am not opposed to the string change into SnarkOS, I just mentioned it as this could lead to bigger consequences that are effecting more than just the JS/TS SDK, and thus would lead to more change-management being needed as it is a bigger impactful change. From some parties that have used our API's I've noticed that they do have parsers implemented expecting actual number formated items and not string when it comes to those.

Pauan:
> I just mentioned it as this could lead to bigger consequences that are effecting more than just the JS/TS SDK, and thus would lead to more change-management being needed as it is a bigger impactful change.

Yes, of course, it is definitely a bigger change.

Pauan:
Thanks for your work on this, and for reporting this issue!

We decided to go with a slightly different implementation approach (https://github.com/ProvableHQ/sdk/pull/925), so I'm going to close this PR.

The next release will have proper bigint support!

PR #905, opened 2024-07-26
Removes extra caller input from transfer_public

Zack-Xb:
## Motivation
Executing a public transfer using the program manager `transfer` function fails as the `transfer_public` option in `wasm/src/programs/manager/transfer.rs` has 3 inputs not 2 as required by the program function as you can see here https://explorer.aleo.org/program/credits.aleo. Removing the first parameter should resolve this, and also the `caller` address parameter is not required for the transfer function.

## Test Plan
Re compile the rust to wasm and test out executing `ProgramManager.transfer({amount}, {recipient}, {transfer_type}, {fee})`.


jaketarnow (approved):
Thanks for this @Zack-Xb! 
lgtm

christianwwwwwwww (approved):
LGTM 🤓

iamalwaysuncomfortable (changes requested):
These do not include the changes to the javascript end of the SDK nor the Provable.tools website. Those signatures should also be updated - #907 addresses these.

iamalwaysuncomfortable:
Changes have been addressed in succeeding PRs, thanks for the callout

PR #907, opened 2024-08-01
[Fix] Ensure credits.aleo keys are fetched from SnarkVM instead of a remote

iamalwaysuncomfortable:
## Motivation

Some SDK users have been having issues with the Key Provider. This is because formerly during Testnet3, verifying keys were fetched from a remote. They are now provided via SnarkVM, this PR updates the KeyProvider to use the keys directly from SnarkVM.


Pauan (approved):
LGTM!

jaketarnow (approved):
Validated fixes allow for program deployment, execution, account creation, and transfers!

LGTM

PR #916, opened 2024-08-14
Create SDK homepage with links to SDK tabs 

michellewong793:
## Motivation

Create a new homepage for the SDK so user doesn't get immediately ported into the /account tab 

## Test Plan
Screenshot of mobile: 
screencapture-localhost-5173-2024-08-14-14_18_46

Screenshot of desktop: 
screencapture-localhost-5173-2024-08-14-14_18_31


christianwwwwwwww (commented):
This looks great! I have some thoughts:

1. I think we should use "zero-knowledge" rather than "zero knowledge" just for consistency with the rest of the SDK and the website 

2. Can we add a link to the Provable website on the link in the top right of the home page?

3. On the `/account` and other pages of the website, can we have the "Aleo SDK" link in the top left link back to the home page?

4. For the buttons, the link is attached to the text itself, rather than the button so when you click on a button but don't click on the text it does nothing. I'd rather the whole button be a link.

website/src/pages/Homepage.jsx
```diff
@@ -0,0 +1,78 @@
...
+                    </Link>
+
+                    <Link to="/transfer" className="actionItem">
+                        <li>
+                            Manage program state and data
+                        </li>
+                    </Link>
+                </ul>
+                <p className="secondSubheader">
+                    Build your first zero knowledge app with create-leo-app
```
christianwwwwwwww: zero-knowledge plz

christianwwwwwwww (approved):
LGTM 🥳

PR #917, opened 2024-08-19
SDK - Update Develop Tab to include Create Account

michellewong793:
## Motivation

Adds "Create Account: to the Develop tab so that user can create a new account before trying to deploy a program in the Develop tab 

References to Monday here: https://provable.monday.com/boards/6731777034/views/146375343/pulses/7220947284

## Test Plan

Screenshot: 
screencapture-localhost-5173-develop-2024-08-19-14_23_59


christianwwwwwwww (approved):
LGTM!

onetrickwolf (approved):
LGTM!

I know this is to the spec of the ticket, but the execute program has an inline "generate random key" for people who wanted to just test a local execution real quick. Could do the same with deploy for people who just want to estimate fee without loading in a real account. 

Can resolve in future tickets though.

PR #918, opened 2024-08-19
SDK - Update UX of Record Tab

michellewong793:
## Motivation

- Add more context about records to the record tab 
- Make items in the main pages more centered for better reading / rather than stretching across the entire screen 

## Test Plan

1. Home page adds a "Try it now" button to lead into the SDK, rather than just one button that leads to docs as the main action in the hero
Screenshot: 
image

2. Record tab updates: 
screencapture-localhost-5173-record-2024-08-19-14_47_16

### DIFF
--- a/website/src/main.jsx
+++ b/website/src/main.jsx
@@ -104,7 +104,7 @@ function Main() {
                         />
                     </Sider>
                     <Layout>
-                        <Content style={{ padding: "50px 50px" }}>
+                        <Content style={{ padding: "50px 50px", margin: "0 auto", minWidth: "850px" }}>
                             <Outlet />
                         </Content>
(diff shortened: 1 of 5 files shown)


onetrickwolf (approved):

onetrickwolf:
Not sure if it'd be useful to include a link to this as well: https://developer.aleo.org/concepts/records/

onetrickwolf:
Adding @iamalwaysuncomfortable as optional reviewer too in case he wants to double check the record explanation 😅

michellewong793:
> Not sure if it'd be useful to include a link to this as well: https://developer.aleo.org/concepts/records/

good idea! ill add

PR #919, opened 2024-08-20
Add playwright tests

robwafle:
## Motivation

Add playwright tests to the react template so we can eventually add these in the build and keep the template stable.

## Test Plan

Ran tests locally for now, they passed locally and they can be added to ci at any time

### DIFF
--- a/create-leo-app/template-react-leo/.gitignore
+++ b/create-leo-app/template-react-leo/.gitignore
@@ -1,0 +1,2 @@
+playwright-report/**
+screenshots/**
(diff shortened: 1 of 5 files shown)


onetrickwolf (changes requested):
Hey general idea looks good but you generally want to avoid manual waiting in e2e tests. Please look at these docs: https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting

https://playwright.dev/docs/actionability

The times will not be consistent across machines, network conditions, etc this can lead to flake and also slows the tests down if they load faster than the manual wait.

PR #925, opened 2024-09-07
Changing JSON parsing to use bigint for all integers

### LINKED ISSUE #892: [Bug] Block data parsing loses precision and typing of bigInts

Pauan:
Fixes #892.

This is an alternate solution to #901

This parses all integers as a BigInt. We might want to be more selective, such as only parsing certain keys as BigInt.

### DIFF
--- a/sdk/src/network-client.ts
+++ b/sdk/src/network-client.ts
@@ -390,9 +390,9 @@ class AleoNetworkClient {
    * @example
    * const latestHeight = networkClient.getLatestHeight();
    */
-  async getLatestHeight(): Promise<number> {
+  async getLatestHeight(): Promise<bigint> {
     try {
-      return await this.fetchData<number>("/latest/height");
+      return await this.fetchData<bigint>("/latest/height");
     } catch (error) {
       throw new Error("Error fetching latest height.");
     }
--- a/sdk/src/network-client.ts
+++ b/sdk/src/network-client.ts
@@ -390,9 +390,9 @@ class AleoNetworkClient {
    * @example
    * const latestHeight = networkClient.getLatestHeight();
    */
-  async getLatestHeight(): Promise<bigint> {
+  async getLatestHeight(): Promise<number> {
     try {
-      return await this.fetchData<bigint>("/latest/height");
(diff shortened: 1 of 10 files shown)


sdk/src/network-client.ts
```diff
@@ -390,9 +392,9 @@ class AleoNetworkClient {
    * @example
    * const latestHeight = networkClient.getLatestHeight();
    */
-  async getLatestHeight(): Promise<number | Error> {
+  async getLatestHeight(): Promise<bigint | Error> {
```
iamalwaysuncomfortable: Block height should likely stay `number` in return type because block height is a u32 (and thus within the range of `number`) and several methods take block height as a `number` type which would lead to a frustrating explicit cast for end users. 

Even though with `ParseJson` we'd get it back as a big int, it'll always be a safe conversion to `number`, so we should probably do that directly in this method. 

Otherwise treating all numbers in the block JSON as BigInt is a safe approach as there are a ton of numbers represented as `u64` (and even the stray `u128` here and there) in rust.

Pauan: Yeah that's reasonable, I made that change.

jaketarnow (approved):
LGTM

PR #931, opened 2024-09-25
Adding in change-version and deploy scripts

Pauan:
Currently, when releasing a new version, we have to manually update the version of multiple packages, and then update all of the templates, and manually do an `npm publish`.

This is a lot of work, and it's easy to make a mistake.

So this PR adds in a new `yarn change-version` and `yarn deploy` commands, which automates the entire process.

Now it's possible to release a new version by doing this:

```bash
yarn change-version 0.7.0
yarn deploy
```


onetrickwolf (approved):
very slick, thanks!

jaketarnow (approved):
lgtm!!

PR #932, opened 2024-09-25
Fixing bug with change-version script

Pauan:
There was a small bug with the `change-version` script which caused `build:create-leo-app` to fail. This PR fixes the bug.


jaketarnow (approved):
Thank you!!

PR #934, opened 2024-09-26
Update WASM testing

alexpitsikoulis:
- Removed test.js file, removal of 'type': 'module' for testing is now handled directly in the npm test script.
- Node version requirement for WASM modules to make network requests is now listed in readme and package.json.


jaketarnow (approved):
LGTM

PR #935, opened 2024-09-27
Fixing build errors and change-version script

Pauan:
This fixes a build error with Node 22, and also fixes the `change-version` script so that it correctly updates the `Cargo.toml` file.


jaketarnow (approved):
LGTM

PR #940, opened 2024-10-11
create-leo-app: explicit template definition, version bump

esren0x:
* Updated the template definitions to be explicit instead of pattern, due to yarn/npm working with these differently.
* Version bump
* New publish of create-leo-app here https://www.npmjs.com/package/create-leo-app/v/0.7.3?activeTab=code


jaketarnow (approved):
LGTM

PR #944, opened 2024-11-26
[Feature] Add arithmetic types, plaintext, and conversion of plaintext & records to JS objects

iamalwaysuncomfortable:
## Motivation

From several requests and issues surrounding the SDK, the following features have been flagged as important to add.

### Addition Field, Scalar, and Group Types
Addition of `Field`, `Group`, and `Scalar` types and core arithmetic methods on those types allowing useful work to be done using algebraic types.

### Accessible Plaintext and Record Data + Conversion of Aleo types to JS type.
This PR adds the ability to access:
* Methods to access the field of of a `Record` directly
* Methods to access `Struct` fields
* Methods for converting `Plaintext` types into JS objects. This includes Aleo `Literal` ( i.e. `address`, `boolean`, `i8/i16/i32/i64/i128/u8/u16/u32/u64/u128`, `field`, `scalar`, `signature`), `Struct`, and `Array` types) into JS types.

### Transition and Transaction Data Accessor Methods
The ability to search for `records`, and get the `inputs` and `outputs` of `Transitions` and `Transactions`.

### Addition of `GraphKey` and `ComputeKey`
This PR adds both the `GraphKey` and `ComputeKey` to the sdk, allowing GraphKeys to be used directly to determine ownership `records`.

## Test Plan

* This PR adds unit tests for all new features.


wasm/src/ledger/transaction.rs
```diff
@@ -0,0 +1,623 @@
...
+        self.0.to_string()
+    }
+
+    /// Get the transaction as a Uint8Array of left endian bytes.
+    ///
+    /// @returns {Uint8Array} Uint8Array representation of the transaction
+    #[wasm_bindgen(js_name = toBytesLe)]
+    pub fn to_bytes_le(&self) -> Result<Uint8Array, String> {
+        let bytes = self.0.to_bytes_le().map_err(|e| e.to_string())?;
+        let array = Uint8Array::new_with_length(bytes.len() as u32);
```
Pauan: This should instead use [`Uint8Array::from`](https://docs.rs/js-sys/latest/js_sys/struct.Uint8Array.html#impl-From%3C%26%5Bu8%5D%3E-for-Uint8Array), which is simpler and faster.

wasm/src/ledger/transaction.rs
```diff
@@ -0,0 +1,623 @@
...
+        self.0.is_fee()
+    }
+
+    /// Get the record plaintext present in a transaction owned by a specific view key.
+    ///
+    /// @param {ViewKey} view_key View key used to decrypt the ciphertext
+    ///
+    /// @returns {Array<RecordPlaintext>} Array of record plaintext objects
+    pub fn owned_records(&self, view_key: &ViewKey) -> Array {
+        let array = Array::new();
```
Pauan: This should instead use iterators and `collect`:

```rust
self.0.records()
    .into_iter()
    .filter_map(|(_, record_ciphertext)| {
        ...
    })
    .collect()
```

wasm/src/ledger/transaction.rs
```diff
@@ -0,0 +1,623 @@
...
+            }
+        });
+        array
+    }
+
+    /// Get the records present in a transaction and their commitments.
+    ///
+    /// @returns {Array<{commitment: Field, record: RecordCiphertext}>} Array of record ciphertext objects
+    pub fn records(&self) -> Array {
+        let array = Array::new();
```
Pauan: Same as above, this should use `map` + `collect`.

wasm/src/ledger/transaction.rs
```diff
@@ -0,0 +1,623 @@
...
+        self.0.records().for_each(|(commitment, record_ciphertext)| {
+            let object = Object::new();
```
Pauan: This should use the `object!` utility macro.

wasm/src/ledger/transaction.rs
```diff
@@ -0,0 +1,623 @@
...
+    /// If the transaction is a deployment transaction, this function will return the program id and
+    /// a list of the functions and their verifying keys, constraint, and variable counts.
+    ///
+    /// @param {boolean} convert_to_js If true the inputs and outputs will be converted to JS objects,
+    /// if false the inputs and outputs will be in wasm format.
+    ///
+    /// @returns {Object} Transaction summary
+    pub fn summary(&self, convert_to_js: bool) -> Result<Object, String> {
+        // Create a transaction object.
+        let transaction = Object::new();
```
Pauan: Same as above, it would be nice to use the `object!` utility macro.

wasm/src/ledger/transaction.rs
```diff
@@ -0,0 +1,623 @@
...
+                        .unwrap();
+                } else {
+                    Reflect::set(&transition_object, &JsValue::from_str("tpk"), &JsValue::from(tpk)).unwrap();
+                    Reflect::set(&transition_object, &JsValue::from_str("tcm"), &JsValue::from(tcm)).unwrap();
+                    Reflect::set(&transition_object, &JsValue::from_str("scm"), &JsValue::from(scm)).unwrap();
+                }
+
+                // Get the inputs.
+                let inputs = Array::new();
+                for input in transition.inputs() {
```
Pauan: This should use iterators + `collect`.

wasm/src/ledger/transaction.rs
```diff
@@ -0,0 +1,623 @@
...
+                    inputs.push(&input_to_js_value(input, convert_to_js));
+                }
+                Reflect::set(&transition_object, &JsValue::from_str("inputs"), &inputs).unwrap();
+
+                // Get the outputs.
+                let outputs = Array::new();
```
Pauan: This should use iterators + `collect`.

wasm/src/ledger/transaction.rs
```diff
@@ -0,0 +1,623 @@
...
+            &JsValue::from_str(&self.transaction_type().to_string()),
+        )
+        .unwrap();
+        Reflect::set(&transaction, &JsValue::from_str("baseFee"), &JsValue::from(base_fee)).unwrap();
+        Reflect::set(&transaction, &JsValue::from_str("fee"), &JsValue::from(fee)).unwrap();
+        Reflect::set(&transaction, &JsValue::from_str("priorityFee"), &JsValue::from(priority_fee)).unwrap();
+
+        // If the transaction is an execution, create an array of transitions.
+        if self.0.is_execute() {
+            let transitions = Array::new();
```
Pauan: This should use iterators + `collect`.

wasm/src/ledger/transaction.rs
```diff
@@ -0,0 +1,623 @@
...
+        // If the transaction is a deployment, summarize the deployment.
+        if let Some(deployment) = self.0.deployment() {
+            let deployment_object = Object::new();
+            Reflect::set(
+                &deployment_object,
+                &JsValue::from_str("programId"),
+                &JsValue::from_str(&deployment.program_id().to_string()),
+            )
+            .unwrap();
+            let functions = Array::new();
```
Pauan: This should use iterators + `collect`.

wasm/src/ledger/transaction.rs
```diff
@@ -0,0 +1,623 @@
...
+        match &self.0 {
+            TransactionNative::Deploy(..) => "deploy".to_string(),
+            TransactionNative::Execute(..) => "execute".to_string(),
+            TransactionNative::Fee(..) => "fee".to_string(),
+        }
+    }
+
+    /// Get the transitions in a transaction.
+    pub fn transitions(&self) -> Array {
+        let transitions = Array::new();
```
Pauan: This should use iterators + `collect`.

wasm/src/ledger/transition.rs
```diff
@@ -0,0 +1,393 @@
...
+        self.0.to_string()
+    }
+
+    /// Get the transition as a Uint8Array of left endian bytes.
+    ///
+    /// @returns {Uint8Array} Uint8Array representation of the transition
+    #[wasm_bindgen(js_name = toBytesLe)]
+    pub fn to_bytes_le(&self) -> Result<Uint8Array, String> {
+        let bytes = self.0.to_bytes_le().map_err(|e| e.to_string())?;
+        let array = Uint8Array::new_with_length(bytes.len() as u32);
```
Pauan: This should use `Uint8Array::from`.

wasm/src/ledger/transition.rs
```diff
@@ -0,0 +1,393 @@
...
+        self.0.find_record(commitment).map(|record_ciphertext| RecordCiphertext::from(record_ciphertext))
+    }
+
+    /// Get the record plaintext present in a transition owned by a specific view key.
+    ///
+    /// @param {ViewKey} view_key The view key of the record owner.
+    ///
+    /// @returns {Array<RecordPlaintext>} Array of record plaintext objects
+    pub fn owned_records(&self, view_key: &ViewKey) -> Array {
+        let array = Array::new();
```
Pauan: This should use iterators + `collect`.

wasm/src/ledger/transition.rs
```diff
@@ -0,0 +1,393 @@
...
+        array
+    }
+
+    /// Get the records present in a transition and their commitments.
+    ///
+    /// @returns {Array<{commitment: Field, record: RecordCiphertext}>} Array of record ciphertext objects
+    pub fn records(&self) -> Array {
+        let array = Array::new();
+        self.0.records().for_each(|(commitment, record_ciphertext)| {
+            let object = Object::new();
```
Pauan: Ideally this would use the `object!` utility macro.

wasm/src/ledger/transition.rs
```diff
@@ -0,0 +1,393 @@
...
+            }
+        });
+        array
+    }
+
+    /// Get the records present in a transition and their commitments.
+    ///
+    /// @returns {Array<{commitment: Field, record: RecordCiphertext}>} Array of record ciphertext objects
+    pub fn records(&self) -> Array {
+        let array = Array::new();
```
Pauan: This should use iterators + `collect`.

wasm/src/ledger/transition.rs
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+    /// Get the inputs of the transition.
+    ///
+    /// @param {bool} convert_to_js If true the inputs will be converted to JS objects, if false
+    /// the inputs will be in wasm format.
+    ///
+    /// @returns {Array} Array of inputs
+    pub fn inputs(&self, convert_to_js: bool) -> Array {
+        let js_array = Array::new();
```
Pauan: This should use iterators + `collect`.

wasm/src/ledger/transition.rs
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+    /// Get the outputs of the transition.
+    ///
+    /// @param {bool} convert_to_js If true the outputs will be converted to JS objects, if false
+    /// the outputs will be in wasm format.
+    ///
+    /// @returns {Array} Array of outputs
+    pub fn outputs(&self, convert_to_js: bool) -> Result<Array, String> {
+        let js_array = Array::new();
```
Pauan: This should use iterators + `collect`.

wasm/src/ledger/transaction.rs
```diff
@@ -0,0 +1,623 @@
...
+        .unwrap();
+        Reflect::set(&transaction, &JsValue::from_str("baseFee"), &JsValue::from(base_fee)).unwrap();
+        Reflect::set(&transaction, &JsValue::from_str("fee"), &JsValue::from(fee)).unwrap();
+        Reflect::set(&transaction, &JsValue::from_str("priorityFee"), &JsValue::from(priority_fee)).unwrap();
+
+        // If the transaction is an execution, create an array of transitions.
+        if self.0.is_execute() {
+            let transitions = Array::new();
+            for transition in self.0.transitions() {
+                let transition_object = Object::new();
```
Pauan: This should use the `object!` utility macro.

wasm/src/ledger/transaction.rs
```diff
@@ -0,0 +1,623 @@
...
+                    &transition_object,
+                    &JsValue::from_str("transitionId"),
+                    &JsValue::from_str(&transition.id().to_string()),
+                )
+                .unwrap();
+
+                let tpk = Group::from(transition.tpk());
+                let tcm = Field::from(transition.tcm());
+                let scm = Field::from(transition.scm());
+                if convert_to_js {
```
Pauan: This logic can instead be written like this:

```rust
object! {
    "tpk" => if convert_to_js { ... } else { ... },
    "tcm" => if convert_to_js { ... } else { ... },
    "scm" => if convert_to_js { ... } else { ... },
}
```

wasm/src/ledger/transaction.rs
```diff
@@ -0,0 +1,623 @@
...
+                }
+                Reflect::set(&transition_object, &JsValue::from_str("outputs"), &outputs).unwrap();
+                transitions.push(&transition_object);
+            }
+            Reflect::set(&transaction, &JsValue::from_str("transitions"), &transitions).unwrap();
+        }
+
+        // If the transaction is a deployment, summarize the deployment.
+        if let Some(deployment) = self.0.deployment() {
+            let deployment_object = Object::new();
```
Pauan: This should use the `object!` utility macro.

wasm/src/ledger/transaction.rs
```diff
@@ -0,0 +1,623 @@
...
+            Reflect::set(
+                &deployment_object,
+                &JsValue::from_str("programId"),
+                &JsValue::from_str(&deployment.program_id().to_string()),
+            )
+            .unwrap();
+            let functions = Array::new();
+            for (function_name, (verifying_key, _)) in deployment.verifying_keys() {
+                let function = Object::new();
```
Pauan: This should use the `object!` utility macro.

wasm/src/lib.rs
```diff
@@ -196,6 +199,28 @@ extern "C" {
     pub fn log(s: &str);
 }
 
+#[macro_export]
+macro_rules! array {
+        ($($value:expr),*$(,)?) => {{
+            let array = Array::new();
```
Pauan: I know this isn't your fault, this code was just copy-pasted from earlier code, but for extra robustness it should use `::js_sys::Array::new()` and `::wasm_bindgen::JsValue::from($value)`

wasm/src/lib.rs
```diff
@@ -196,6 +199,28 @@ extern "C" {
...
+            $(array.push(&JsValue::from($value));)*
+
+            array
+        }};
+    }
+
+#[macro_export]
+macro_rules! object {
+        ($($key:literal: $value:expr,)*) => {{
+            let object = Object::new();
```
Pauan: Similarly, this should use `::js_sys::Object::new()` and `::js_sys::Reflect::set` and `::wasm_bindgen::JsValue::from_str` and `::wasm_bindgen::JsValue::from`

wasm/src/programs/data/helpers/future.rs
```diff
@@ -0,0 +1,55 @@
...
+    types::native::{ArgumentNative, FutureNative},
+    Plaintext,
+};
+
+use js_sys::{Array, JsString, Object, Reflect};
+use wasm_bindgen::JsValue;
+
+/// Convert a future to a javascript value.
+pub fn future_to_js_value(argument: &FutureNative, convert_to_js: bool) -> JsValue {
+    let future_object = Object::new();
```
Pauan: This should use the `object!` utility macro.

wasm/src/programs/data/helpers/future.rs
```diff
@@ -0,0 +1,55 @@
...
+    Reflect::set(&future_object, &JsString::from("type"), &JsString::from("future")).unwrap();
+    Reflect::set(&future_object, &JsString::from("programId"), &JsString::from(argument.program_id().to_string()))
+        .unwrap();
+    Reflect::set(
+        &future_object,
+        &JsString::from("functionName"),
+        &JsString::from(argument.function_name().to_string()),
+    )
+    .unwrap();
+    let arguments = Array::new();
```
Pauan: This should use iterators + `collect`.

wasm/src/programs/data/helpers/input.rs
```diff
@@ -0,0 +1,80 @@
...
+                    JsValue::from_str(&ciphertext.to_string())
+                } else {
+                    JsValue::from(Ciphertext::from(ciphertext))
+                }
+            } else {
+                JsValue::NULL
+            }
+        }
+        InputNative::Record(serial_number, tag) => {
+            let record = Object::new();
```
Pauan: This should use the `object!` utility macro.

wasm/src/programs/data/helpers/input.rs
```diff
@@ -0,0 +1,80 @@
...
+                Reflect::set(&record, &JsString::from("tag"), &JsValue::from(Field::from(tag).to_string())).unwrap();
+            } else {
+                Reflect::set(&record, &JsString::from("serialNumber"), &JsValue::from(Field::from(serial_number)))
+                    .unwrap();
+                Reflect::set(&record, &JsString::from("tag"), &JsValue::from(Field::from(tag))).unwrap();
+            }
+            JsValue::from(record)
+        }
+        InputNative::ExternalRecord(input_commitment) => {
+            let record = Object::new();
```
Pauan: This should use the `object!` utility macro.

wasm/src/programs/data/helpers/literal.rs
```diff
@@ -0,0 +1,93 @@
...
+        LiteralNative::Scalar(literal) => {
+            let js_string = literal.to_string();
+            (&js_string).into()
+        }
+        LiteralNative::Signature(literal) => {
+            let js_string = literal.to_string();
+            (&js_string).into()
+        }
+        LiteralNative::String(literal) => {
+            let js_string = literal.to_string();
```
Pauan: There is a `JsValue::from` [impl for `&str`](https://docs.rs/wasm-bindgen/0.2.97/wasm_bindgen/struct.JsValue.html#impl-From%3C%26str%3E-for-JsValue), so this probably doesn't need the `to_string()`, which causes an extra unneeded allocation.

wasm/src/programs/data/helpers/output.rs
```diff
@@ -0,0 +1,114 @@
...
+                    JsValue::from_str(&ciphertext.to_string())
+                } else {
+                    JsValue::from(Ciphertext::from(ciphertext))
+                }
+            } else {
+                JsValue::NULL
+            }
+        }
+        OutputNative::Record(commitment, checksum, record_ciphertext) => {
+            let record = Object::new();
```
Pauan: This should use the `object!` utility macro.

wasm/src/programs/data/helpers/output.rs
```diff
@@ -0,0 +1,114 @@
...
+                        &JsString::from("recordCiphertext"),
+                        &JsValue::from(RecordCiphertext::from(record_ciphertext)),
+                    )
+                    .unwrap();
+                }
+            }
+            JsValue::from(&record)
+        }
+        OutputNative::ExternalRecord(output_commitment) => {
+            let external_record_object = Object::new();
```
Pauan, line 91: This should use the `object!` utility macro.

wasm/src/programs/data/helpers/plaintext.rs
```diff
@@ -0,0 +1,69 @@
...
+        PlaintextNative::Literal(literal, _) => {
+            let js_value = literal_to_js_value(literal);
+            Reflect::set(js_object, &key.to_string().into(), &js_value.into()).unwrap();
+        }
+        PlaintextNative::Struct(struct_members, ..) => {
+            let struct_object = struct_to_js_object(struct_members);
+            Reflect::set(&js_object, &key.to_string().into(), &struct_object.into()).unwrap();
+        }
+        PlaintextNative::Array(plaintext, ..) => {
+            let js_array = js_sys::Array::new();
```
Pauan: This should use iterators + `collect`.

wasm/src/programs/data/helpers/plaintext.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+}
+
+/// Convert a plaintext to a javascript value.
+pub fn plaintext_to_js_value(plaintext: &PlaintextNative) -> JsValue {
+    match plaintext {
+        PlaintextNative::Literal(literal, _) => literal_to_js_value(literal),
+        PlaintextNative::Struct(struct_members, ..) => JsValue::from(struct_to_js_object(struct_members)),
+        PlaintextNative::Array(plaintext, ..) => {
+            let js_array = js_sys::Array::new();
```
Pauan: This should use iterators + `collect`.

wasm/src/programs/data/helpers/record.rs
```diff
@@ -0,0 +1,53 @@
...
+use crate::{
+    insert_plaintext,
+    types::native::{EntryNative, RecordPlaintextNative},
+};
+use js_sys::{Object, Reflect};
+
+/// Convert a record to a javascript object.
+pub fn record_to_js_object(record: &RecordPlaintextNative) -> Result<Object, String> {
+    // Create a new javascript object to house the record data.
+    let js_record = Object::new();
```
Pauan: This should use the `object!` utility macro.

wasm/src/programs/data/plaintext.rs
```diff
@@ -0,0 +1,255 @@
...
+        Ok(Self(native))
+    }
+
+    /// Generate a random plaintext element from a series of bytes.
+    ///
+    /// @param {Uint8Array} bytes A left endian byte array representing the plaintext.
+    #[wasm_bindgen(js_name = "toBytesLe")]
+    pub fn to_bytes_le(&self) -> Result<Uint8Array, String> {
+        let rust_bytes = self.0.to_bytes_le().map_err(|e| e.to_string())?;
+        let array = Uint8Array::new_with_length(rust_bytes.len() as u32);
```
Pauan: This should use `Uint8Array::from`.

wasm/src/programs/program.rs
```diff
@@ -501,29 +501,10 @@ impl FromStr for Program {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::{array, object};
```
Pauan, line 504: I approve.

wasm/src/types/field.rs
```diff
@@ -14,27 +14,95 @@
...
+        Field(self.0 / other.0)
+    }
+
+    /// Power of a field element.
+    pub fn pow(&self, other: &Field) -> Field {
+        Field(self.0.pow(other.0))
+    }
+
+    /// Invert the field element.
+    pub fn inverse(&self) -> Field {
```
Pauan, line 86: This should be called `negate`.

wasm/src/types/scalar.rs
```diff
@@ -0,0 +1,128 @@
...
+        Scalar(self.0.double())
+    }
+
+    /// Power of a field element.
+    pub fn pow(&self, other: &Scalar) -> Scalar {
+        Scalar(self.0.pow(other.0))
+    }
+
+    /// Invert the field element.
+    pub fn inverse(&self) -> Scalar {
```
Pauan, line 88: This should be called `negate`.

website/src/workers/worker.js
```diff
@@ -418,5 +418,7 @@ self.addEventListener("message", (ev) => {
                 programManager.setHost(defaultHost);
             }
         })();
+    } else if (ev.data.type === "PROPOSE_GAME") {
```
Pauan: What's the purpose of this?

Pauan (changes requested):
There's various small nits that need to be fixed, otherwise it looks good to me.

wasm/src/types/field.rs
iamalwaysuncomfortable: Negate isn't the right terminology in Field Theory given there's no general concept of `negative` as there is in the real or complex fields. In field theory `inverse` has very specific meaning and it's the corresponding member of the field that when added together produces the additive identity element (i.e. 0). If you look in SnarkVM Field traits, `inverse` is the terminology used for this as it matches the formal definitions in theory, so this should stay `inverse`.

wasm/src/types/group.rs
```diff
@@ -0,0 +1,125 @@
...
+
+    /// Double the group element.
+    pub fn double(&self) -> Group {
+        Group(self.0.double())
+    }
+
+    /// Get the negation of the group element.
+    pub fn negate(&self) -> Group {
+        Group(-self.0)
+    }
```
iamalwaysuncomfortable, line 92: @Pauan the reason this is called `negate` is because describes a different operation than the `inverse` operation of a field. A group element is an eliptic curve element and a `negation` is the definition given to the operation of flipping a point about it axis of symmetry, which is what `-self.0` encapsulates. So that's why the difference here.

wasm/src/types/scalar.rs
iamalwaysuncomfortable: This should stay `inverse` same justification as `Field` (`Scalar` is also a finite field). `Inverse` is the mathematical definition of what's happening here and there's no standard `negation` in finite fields.

website/src/workers/worker.js
iamalwaysuncomfortable: This is a mistake, I shall remove.

wasm/src/ledger/transaction.rs
iamalwaysuncomfortable: This will only work if its first wrapped in JsValue, which seems acceptable.

iamalwaysuncomfortable (commented):
Feedback addressed, @Pauan - when there's time to double check, let me know.

wasm/src/ledger/transaction.rs
Pauan: I'm not sure what you mean, it shouldn't need any wrapping.

wasm/src/types/field.rs
Pauan: Hmmmm, in that case I don't think it should be using the `-` operator... since `-` should have the meaning of "negate", not "invert".

iamalwaysuncomfortable: We're generally used to a `negation` being defined for `signed` numeric types, but theoretically it is 
creating the inverse element to the positive number, so it is indeed referring to an `inversion` operation. 

A finite field is different however in that there are no negative numbers, and thus no signed bit to denote the sign. The confusion here stems from the fact that in unsigned integers, there's no inverse element (and thus it's not a field, ring, or even a group by definition) but since a finite field is indeed a field, it follows field axioms and thus has an additive inverse even though it has no negative numbers. However in theoretical notation (and in SnarkVM) the negative sign is used to denote the inverse element, so it's consistent with common notation in Field theory. The number it ends up though is usually a large number.

For instance, in the `ed25519` field (integers mod `2^255 − 19`)

You can write `(-2) + 2 = 0` to say "element 2 + its additive inverse equals zero", but it is shorthand notation for `(2^255 - 21) + 2 = 0`. Thus `(-2) + 2 = (2^255 - 21) + 2 = 0`. This is why using the `Neg` trait in SnarkVM is valid, and using it here is valid as well.

Because we are defining a finite field, there are no negatives and `negate` isn't a standard term in Field arithmetic, but using the negative operator (as defined in SnarkVM) is consistent. Calling this `negate` here is incorrect because it may cause conflation of concepts and lead people to believe this is a signed object and potentially make mistakes in cryptographic code. `Inverse` is the operation defined here and that's the name that needs to stay on it.

Pauan: Thanks for the explanation, that helps a lot.

wasm/src/lib.rs
```diff
@@ -196,6 +199,28 @@ extern "C" {
...
+            array
+        }};
+    }
+
+#[macro_export]
+macro_rules! object {
+        ($($key:literal: $value:expr,)*) => {{
+            let object = ::js_sys::Object::new();
+
+            $(Reflect::set(&object, &::wasm_bindgen::JsValue::from_str($key), &::wasm_bindgen::JsValue::from($value)).unwrap();)*
```
Pauan, line 218: Minor nit: this should be `::js_sys::Reflect::set`

wasm/src/programs/data/helpers/future.rs
```diff
@@ -0,0 +1,50 @@
...
+use wasm_bindgen::JsValue;
+
+/// Convert a future to a javascript value.
+pub fn future_to_js_value(argument: &FutureNative, convert_to_js: bool) -> JsValue {
+    let future_object = object! {
+        "type" : "future",
+        "programId" : argument.program_id().to_string(),
+        "functionName" : argument.function_name().to_string(),
+    };
+    let arguments = argument
```
Pauan, line 34: This can be moved into the `object!` so it can avoid the `Reflect::set`.

wasm/src/programs/data/helpers/input.rs
```diff
@@ -0,0 +1,86 @@
...
+                }
+            } else {
+                JsValue::NULL
+            }
+        }
+        InputNative::Record(serial_number, tag) => {
+            let record = object! {
+                "type": "record",
+            };
+            if convert_to_js {
```
Pauan, line 53: I think it's cleaner to do something like this:

```rust
if convert_to_js {
    object! {
        "type": "record",
        ...
    }
} else {
    object! {
        "type": "record",
        ...
    }
}
```

wasm/src/programs/data/helpers/input.rs
```diff
@@ -0,0 +1,86 @@
...
+                    .unwrap();
+                Reflect::set(&record, &JsString::from("tag"), &JsValue::from(Field::from(tag))).unwrap();
+            }
+            JsValue::from(record)
+        }
+        InputNative::ExternalRecord(input_commitment) => {
+            let external_record = object! {
+                "type": "extneralRecord",
+            };
+            if convert_to_js {
```
Pauan, line 68: Same as above, better to use `object!` instead of `Reflect::set`

wasm/src/programs/data/helpers/output.rs
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+        OutputNative::Record(commitment, checksum, record_ciphertext) => {
+            // Create a record object.
+            let record = object! {
+                "type": "record",
+            };
+
+            // Create a record object.
+            if convert_to_js {
```
Pauan, line 65: Same as above, better to use `object!` instead of `Reflect::set`

Pauan (changes requested):
Couple more minor issues to fix, otherwise it looks good to me.

wasm/src/ledger/transaction.rs
iamalwaysuncomfortable: TPK/TCM/SCM aren't all the same concrete type.

Pauan (approved):

jaketarnow (approved):
LGTM

PR #948, opened 2024-12-10
[Feature] Add Field, Group, Scalar, Plaintext, Ciphertext, Transitions, and Transactions to JS SDK.

iamalwaysuncomfortable:
## Motivation

Several feature requests have requested that the ability to the following be added to the SDK.
* Be able to do Field, Group, and Scalar arithmetic and be generally able to work with those types 
* Easily peer into Transaction and Transition data and associated inputs and outputs
* Work with Ciphertexts and get items from `Struct` and `Array` types

This PR adds these types to the SDK with the ability to turn them into JavaScript/TypeScript objects for convenient extraction of data into a front or backend app as well as api methods that allow these new types to be directly instantiated when calling api endpoints.

It also provides several new typescript types that allow typescript users to work with the JS representation of these objects.

Once this PR is approved, it will constitute the release candidate for the next major version bump.

## Test Plan

This PR adds the following unit tests for the new objects:
* Field, Scalar, and Group arithmetic.
* Conversion of Transitions and Transactions to JS objects.

## Related PRs

#944 - this PR adds the exports of SnarkVM types to `wasm` and provides code that


sdk/src/models/confirmed_transaction.ts
```diff
@@ -1,7 +1,7 @@
-import { TransactionModel } from "./transactionModel";
+import { TransactionJSON } from "./transaction/transactionJSON";
 
 export type ConfirmedTransaction = {
```
Pauan: For objects, it's better to use an `interface` instead of `type`

sdk/src/models/deploy.ts
```diff
@@ -0,0 +1,6 @@
+import { FunctionObject } from "./functionObject";
+
+export type DeploymentMetadata = {
```
Pauan: Same as above, this should be an `interface`

sdk/src/models/executionJSON.ts
```diff
@@ -0,0 +1,6 @@
+import { TransitionJSON } from "./transition/transitionJSON";
+
+export type ExecutionJSON = {
```
Pauan: Same as above, this should be an `interface`

sdk/src/models/functionObject.ts
```diff
@@ -0,0 +1,8 @@
+import { VerifyingKey } from "@provablehq/wasm";
+
+export type FunctionObject = {
```
Pauan: Same as above, this should be an `interface`

sdk/src/models/input/inputJSON.ts
```diff
@@ -1,4 +1,7 @@
-export type Input = {
+/**
+ * Object representation of an Input as raw JSON returned from a SnarkOS node.
+ */
+export type InputJSON = {
```
Pauan: Same as above, this should be an `interface`

sdk/src/models/input/inputObject.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Aleo function Input represented as a typed typescript object.
+ */
+import { Ciphertext, Field } from "@provablehq/wasm";
+import { Plaintext } from "@provablehq/wasm/mainnet.js";
+import { PlaintextObject } from "../plaintext/plaintext";
+
+/**
+ * Object representation of an Input as raw JSON returned from a SnarkOS node.
+ */
+export type InputObject = {
```
Pauan: Same as above, this should be an `interface`

sdk/src/models/input/inputObject.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Aleo function Input represented as a typed typescript object.
+ */
+import { Ciphertext, Field } from "@provablehq/wasm";
+import { Plaintext } from "@provablehq/wasm/mainnet.js";
```
Pauan: Why is this importing `mainnet`?

sdk/src/models/input/inputObject.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Aleo function Input represented as a typed typescript object.
+ */
+import { Ciphertext, Field } from "@provablehq/wasm";
```
Pauan: This should be importing `../../wasm` instead of `@provablehq/wasm`

sdk/src/models/output/outputJSON.ts
```diff
@@ -1,4 +1,4 @@
-export type Output = {
+export type OutputJSON = {
```
Pauan: Same as above, this should be an `interface`

sdk/src/models/output/outputObject.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Aleo function Input represented as a typed typescript object.
+ */
+import { Field, Ciphertext, Plaintext } from "@provablehq/wasm";
+import { PlaintextObject } from "../plaintext/plaintext";
+
+/**
+ * Object representation of an Input as raw JSON returned from a SnarkOS node.
+ */
+export type OutputObject = {
```
Pauan: Same as above, this should be an `interface`

sdk/src/models/output/outputObject.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Aleo function Input represented as a typed typescript object.
+ */
+import { Field, Ciphertext, Plaintext } from "@provablehq/wasm";
```
Pauan: This should import `../../wasm` instead.

sdk/src/models/transaction/transactionJSON.ts
```diff
@@ -0,0 +1,7 @@
+import { ExecutionJSON } from "../executionJSON";
+
+export type TransactionJSON = {
```
Pauan: Same as above, this should be an `interface`

sdk/src/models/transaction/transactionSummary.ts
```diff
@@ -0,0 +1,12 @@
+import { TransitionObject } from "../transition/transitionObject";
+import { DeploymentMetadata } from "../deploy";
+
+export type TransactionSummary = {
```
Pauan: Same as above, this should be an `interface`

sdk/src/models/transition/transitionJSON.ts
```diff
@@ -0,0 +1,14 @@
+import { InputJSON } from "../input/inputJSON";
+import { OutputJSON } from "../output/outputJSON";
+
+export type TransitionJSON = {
```
Pauan: Same as above, this should be an `interface`

sdk/src/models/transition/transitionObject.ts
```diff
@@ -0,0 +1,16 @@
+import { InputObject } from "../input/inputObject";
+import { OutputObject } from "../output/outputObject";
+import { Field, Group } from "@provablehq/wasm";
+
+export type TransitionObject = {
```
Pauan: Same as above, this should be an `interface`

sdk/src/models/transition/transitionObject.ts
```diff
@@ -0,0 +1,16 @@
+import { InputObject } from "../input/inputObject";
+import { OutputObject } from "../output/outputObject";
+import { Field, Group } from "@provablehq/wasm";
```
Pauan: This should import `../../wasm`

sdk/src/network-client.ts
```diff
@@ -574,30 +634,85 @@ class AleoNetworkClient {
...
+   * @param {number} height
+   * @example
+   * const transactions = networkClient.getTransactionObjects(654);
+   *
+   * let transaction_summaries = transactions.map(transaction => transaction.summary());
+   */
+  async getTransactionObjects(height: number): Promise<Array<Transaction>> {
+    try {
+      return (await this.fetchData<Array<string>>("/block/" + height.toString() + "/transactions"))
+          .reduce<Array<Transaction>>((acc, transaction) => {
```
Pauan: Why is this using `reduce` instead of `map`?

sdk/src/network-client.ts
```diff
@@ -609,9 +724,27 @@ class AleoNetworkClient {
...
+  /**
+   * Returns the transactions in the memory pool as wasm objects.
+   *
+   * @example
+   * const transactions = networkClient.getTransactionsInMempool();
+   */
+  async getTransactionObjectsInMempool(): Promise<Array<Transaction>> {
+    try {
+      return (await this.fetchData<Array<string>>("/memoryPool/transactions"))
+          .reduce<Array<Transaction>>((acc, transaction) => {
```
Pauan: Similarly, why is this using `reduce`?

sdk/tests/key-provider.test.ts
```diff
@@ -11,134 +11,134 @@ describe('KeyProvider', () => {
...
-            expect(transferPrivateToPublicProverLocal.isTransferPrivateToPublicProver()).equal(true);
-            expect(transferPrivateToPublicVerifierLocal.isTransferPrivateToPublicVerifier()).equal(true);
-            expect(transferPublicProverLocal.isTransferPublicProver()).equal(true);
-            expect(transferPublicVerifierLocal.isTransferPublicVerifier()).equal(true);
-            expect(transferPublicToPrivateProverLocal.isTransferPublicToPrivateProver()).equal(true);
-            expect(transferPublicToPrivateVerifierLocal.isTransferPublicToPrivateVerifier()).equal(true);
-            expect(unbondPublicProverLocal.isUnbondPublicProver()).equal(true);
-            expect(unbondPublicVerifierLocal.isUnbondPublicVerifier()).equal(true);
-        });
+        // it('should not fetch invalid transfer keys', async () => {
```
Pauan: I think it's cleaner to use `describe.skip` or `it.skip` to skip the tests, instead of commenting it out.

sdk/tests/network-client.test.ts
```diff
@@ -1,9 +1,10 @@
 import sinon from "sinon";
 import { expect } from "chai";
-import {Account, Block, AleoNetworkClient, TransactionModel} from "../src/node";
-import {beaconAddressString, beaconPrivateKeyString} from "./data/account-data";
-import {log} from "console";
-
+import { Account, BlockJSON, AleoNetworkClient, TransactionSummary, InputObject, OutputObject } from "../src/node";
+import {beaconPrivateKeyString} from "./data/account-data";
+import { Plaintext, Transition } from "@provablehq/wasm";
```
Pauan: This should be importing from `../src/node` instead of `@provablehq/wasm`.

sdk/tests/network-client.test.ts
```diff
@@ -1,9 +1,10 @@
...
+import { TransitionObject } from "../src/models/transition/transitionObject";
```
Pauan: This should probably also be importing from `../src/node`

sdk/tests/network-client.test.ts
```diff
@@ -1,9 +1,10 @@
...
+import { PlaintextObject } from "../src/models/plaintext/plaintext";
```
Pauan: This should probably also be importing from `../src/node`

Pauan (changes requested):
Some minor nits, but generally looks good.

Pauan:
Since the docs are being auto-generated by the `build:sdk-docs` script, shouldn't the `.html` files be added into `.gitignore`?

Pauan (approved):
LGTM

iamalwaysuncomfortable:
> Since the docs are being auto-generated by the `build:sdk-docs` script, shouldn't the `.html` files be added into `.gitignore`?

In theory yes, but in practice when users, especially ones new to the protocol or who aren't great at JS arrive here they don't think to run that. I've found users reaching for the HTML docs immediately and getting use out of it, so this is why it's done currently. We should however replace this with a very prominent link to these documents online. That doesn't exist yet, but I'll talk to @jaketarnow to see if we can make it so.

jaketarnow (approved):
LGTM

sdk/src/network-client.ts
```diff
@@ -551,16 +588,60 @@ class AleoNetworkClient {
...
+   * @param {string} mappingName - The name of the mapping to get the value of (e.g. "account")
+   * @param {string | Plaintext} key - The key of the mapping to get the value of (e.g. "aleo1rhgdu77hgyqd3xjj8ucu3jj9r2krwz6mnzyd80gncr5fxcwlh5rsvzp9px")
+   *
+   * @return {Promise<string>} String representation of the value of the mapping
+   */
+  async getProgramMappingPlaintext(programId: string, mappingName: string, key: string | Plaintext): Promise<Plaintext> {
+    try {
+      const keyString = key instanceof Plaintext ? key.toString() : key;
+      const value = await this.fetchRaw("/program/" + programId + "/mapping/" + mappingName + "/" + keyString);
+      return Plaintext.fromString(JSON.parse(value));
```
Pauan, line 637: Why is this using `JSON.parse` instead of `parseJSON`?

sdk/src/utils.ts
```diff
@@ -6,9 +6,8 @@ export function logAndThrow(message: string): never {
 
 export function parseJSON(json: string): any {
     function revive(key: string, value: any, context: any) {
-        if (Number.isInteger(value)) {
+        if (Number.isInteger(value) && value.length > 15) {
```
Pauan: We should fix this for real.

But in the meantime, if we want to go with the "conditionally returns bigint if it's too big" logic, then it should be like this:

```js
if (Number.isInteger(value) && !Number.isSafeInteger(value)) {
```

Pauan (changes requested):
One minor thing and this'll be good to go.

sdk/src/network-client.ts
iamalwaysuncomfortable: Because it doesn't deserialize into the rust object using parseJSON.

sdk/src/utils.ts
iamalwaysuncomfortable: This was an oversight, thanks for catching it.

sdk/tests/network-client.test.ts
```diff
@@ -357,7 +357,7 @@ describe('NodeConnection', () => {
                 expect(transactions.length).equal(4);
                 expect(transactions[0].status).equal("accepted");
                 expect(transactions[0].type).equal("execute");
-                expect(transactions[0].index).equal(0);
+                expect(transactions[0].index).equal(BigInt(0));
```
Pauan, line 360: You don't need to make any changes, but just for future reference you can use `0n` for bigint literals, which makes them a lot more convenient to use.

Pauan (approved):

sdk/tests/network-client.test.ts
iamalwaysuncomfortable: That's a good tip, thanks.

PR #957, opened 2025-01-23
Updating versions

Pauan:
This upgrades all of the NPM and Cargo versions. It has a couple other benefits:

* It's no longer necessary to update the `build.js` file when adding a new export from Wasm.

* The `.wasm` file size has been reduced from 8.9 megabytes to 7.3 megabytes (18% reduction).

   With `opt-level = "z"` the `.wasm` file size can be reduced down to 3.5 MB, but that's left for a future PR.


jaketarnow (approved):
LGTM - thanks for working through all the updates @Pauan

wasm/build.js
```diff
@@ -19,21 +20,28 @@ async function buildRollup(input, output) {
 async function buildWasm(network) {
     await buildRollup({
         input: {
-            "aleo_wasm": "entry",
+            "aleo_wasm": "./Cargo.toml",
+            "aleo_wasm_custom": "./Cargo.toml?custom",
```
iamalwaysuncomfortable, line 24: What does this change do?

wasm/build.js
```diff
@@ -19,21 +20,28 @@ async function buildRollup(input, output) {
...
+
+                            // Strips out debug information
+                            "-Z", "location-detail=none",
+                            "-Z", "fmt-debug=none",
+                        ]),
+
+                        "--no-default-features",
+                        "--features", `browser,${network}`,
+                    ],
+                    wasmOpt: ["-O", "--enable-threads", "--enable-bulk-memory", "--enable-bulk-memory-opt"],
```
iamalwaysuncomfortable, line 43: If I remember correctly, you've rebuilt the interface in the rollup between Rayon and the web workers correct?

wasm/build.js
```diff
@@ -19,21 +20,28 @@ async function buildRollup(input, output) {
...
+                extraArgs: {
+                    cargo: [
+                        "--config", "build.rustflags=" + JSON.stringify([
+                            // This enables multi-threading
+                            "-C", "target-feature=+atomics,+bulk-memory,+mutable-globals",
+                            "-C", "link-arg=--max-memory=4294967296",
+
+                            // Strips out debug information
+                            "-Z", "location-detail=none",
+                            "-Z", "fmt-debug=none",
```
iamalwaysuncomfortable, line 37: Does this takeout any console.log statements?

wasm/build.js
```diff
@@ -260,6 +184,11 @@ async function build(network) {
         buildJS(network),
         buildWorker(network),
     ]);
+
+    await Promise.all([
+        $fs.rm($path.join("dist", network, "tmp"), { recursive: true }),
+        $fs.rm($path.join("dist", network, "aleo_wasm_custom.d.ts")),
+    ]);
```
iamalwaysuncomfortable, line 191: What does this do ultimately?

iamalwaysuncomfortable (commented):
Seems good, but asking a few questions so the rest of the maintainers are lucid about what's happening under the hood.

wasm/build.js
Pauan: For Workers we need to manually initialize the Wasm, so it [uses `?custom`](https://github.com/wasm-tool/rollup-plugin-rust?tab=readme-ov-file#customizing-the-wasm-loading) to allow for that.

wasm/build.js
Pauan: This doesn't touch the Worker/Rayon stuff, it's just upgrading to the latest version of `@wasm-tool/rollup-plugin-rust`.

There was a bug that was preventing wasm-opt from running, this PR fixes that, which is why it needs special wasm-opt arguments.

wasm/build.js
Pauan: No, it just removes all uses of `{:?}`.

wasm/build.js
Pauan: The build process generates some files which aren't needed, so it cleans them up in order to reduce the size of the npm package.

iamalwaysuncomfortable (approved):

PR #960, opened 2025-01-27
[Fix] Fix minor UI issues with scalar multiplication and multiplicative inverse of fields.

iamalwaysuncomfortable:
## Motivation

Two small UI errors on the provable.tools existed on the new algebra page.
* The multiplicative inverse calculation wasn't triggered without two field elements being in the react state. (The second field element wasn't used, the logic simply set the Result field to empty).
* The scalar multiplication calculation didn't update with the "Random" button.

## Changes
* Logic is updated to ensure when the multiplicative inverse operation is selected, only the first field element field needs to be populated to proceed the inverse calculation.
* The `random` button in the scalar input field triggers a result update.


jaketarnow (approved):
LGTM

PR #961, opened 2025-01-30
Fixing issue with polyfill

### LINKED ISSUE #956: Global fetch method impacts other SDKs
### The Problem

The SDK calls `fetch` to load the wasm file, but nodejs' native `fetch` method cannot read the file so the global fetch method has been modified to cater for this. This affects other sdks that run in the same environment as this sdk which creates issues for certain companies to integrate Aleo using this sdk.

### Potential Solution

To get around this issue the [global fetch polyfill](https://github.com/ProvableHQ/sdk/blob/mainnet/sdk/src/polyfill/fetch.ts) can possibly be renamed to something else, for example `new_fetch()`. All the current uses of `fetch()` within the sdk would be renamed to `new_fetch()`. 

Like this the global fetch polyfill would not impact other sdks used within the same environment.

Pauan:
Fixes #956


iamalwaysuncomfortable (commented):
I figured out that most of the people speaking about this are operating in node.

Also, paraphrasing @onetrickwolf to provide context.

"I looked a little more into [Issue 956](https://github.com/ProvableHQ/sdk/issues/956). I thought this was a scoped export but I think since we are literally setting the globalThis.fetch it would replace anyone else's polyfill/fetch. I think the issue is, someone is importing the SDK into a `nodejs` project where they are already using a fetch polyfill for some other reason, then this overwrites it maybe breaking other functionality.

I think we did this as a fix just to quick fix all instances of fetch in the SDK that would want to read from file in nodejs, but I think we would need to do what @zkxuerb is saying: give it a custom name then replace all instances of fetch in the SDK with a customFetch. I think it's probably better practice to do that anyway, but I also don't know if we had to do this due to something in the wasm or something that we wouldn't be able to rename?"

anthonyjoeseph:
The polyfill breaks usage of the `axios` package as well. Not sure if this is already fixed by this PR but I wanted to mention just in case

```js
import { ViewKey } from '@aleohq/sdk';
import axios from 'axios';

console.log(!!ViewKey);

axios.get('https://jsonplaceholder.typicode.com/todos/1').then((resp) => {
  console.log(resp.data);
});
```

output:
> Error: Async XMLHttpRequest is not implemented yet

[replit of this](https://replit.com/@AnthonyGabriele/aleo-axios-error)

Pauan:
@anthonyjoeseph Thanks for bringing that to our attention, we're looking into it.

Pauan:
@anthonyjoeseph I updated this PR to fix the issue with axios, thanks!

jaketarnow (approved):
LGTM

PR #974, opened 2025-02-20
[Documentation] WASM Objects

alexpitsikoulis:
This PR modifies the sdk-docs build script to include the WASM objects


scripts/jsdoc.sh
```diff
@@ -0,0 +1,13 @@
...
+
+if [ -f "wasm/dist/mainnet/aleo_wasm.d.ts" ]; then
+  echo "WASM already built, skipping..."
+else
+    yarn build:wasm
+fi
+
+cp wasm/dist/mainnet/aleo_wasm.d.ts sdk/src/wasm.ts
+
+jsdoc -c sdk/jsdoc.json
```
iamalwaysuncomfortable, line 11: Can we find some way to make the output for this be Markdown files instead of HTML? And can we have the target directory be `docs/api_reference` (within the top-level API docs folder)

iamalwaysuncomfortable (commented):
A few changes requested:

1. Can we find some way to make the output for this be Markdown files instead of HTML? 2. can we have the target directory be `docs/api_reference` (i.e. the docs folder in the root of the entire repo, not the one in the sdk subdirectory)
3. Do you know why the class definitions are being duplicated? If you know a way of de-duping this that would be excellent.

iamalwaysuncomfortable:
@jaketarnow Can you check the rendering of this in stoplight?

sdk/src/wasm/index.ts
```diff
@@ -0,0 +1,2067 @@
+/* tslint:disable */
```
iamalwaysuncomfortable: Where did this file come from? It should likely be deleted no?

iamalwaysuncomfortable (commented):
Overall looking great. @jaketarnow any way we can check to see how this renders in stoplight?

A couple nits:
* I think we can eliminate the `12_api.md` and `swagger.json` can be eliminated as those are already in stoplight.
* I believe `sdk/src/wasm/index.ts` was added in error.

sdk/src/wasm/index.ts
alexpitsikoulis: whoops yes, removed

scripts/jsdoc.sh
alexpitsikoulis: done

iamalwaysuncomfortable:
@jaketarnow can you give the final review on this? It's approved from my side but I want to ensure it looks good on spotlight before we publish.

iamalwaysuncomfortable (approved):

PR #978, opened 2025-02-24
[Documentation] Amend transfer examples to use buildTransferTransaction

alexpitsikoulis:
This PR amends the transfer examples to use buildTransferTransaction instead of buildExecutionTransaction

### DIFF
--- a/docs/examples/01_transfer_public.md
+++ b/docs/examples/01_transfer_public.md
@@ -21,17 +21,17 @@ programManager.setAccount(account);
 // Create recipient account.
 const recipient = new Account();
 
-// Execute `transfer_public` function in `credits.aleo`
+// Build a public transfer transaction.
 // Publicly send 5 microcredits to the recipient
-const transaction2 = await programManager.buildExecutionTransaction({
-  programName: "credits.aleo",
-  functionName: "transfer_public",
-  fee: 0.020,
-  privateFee: false,
-  inputs: [recipient.address(), "5u32"],
-  keySearchParams: { "cacheKey": "credits:transfer_public" }
-});
+const transaction = await programManager
+  .buildTransferPublicTransaction(
+    5,
+    recipient
+      .address()
+      .to_string(),
+    0.1
+  );
 
 // Broadcast the transaction to the Aleo network.
-const result2 = await programManager.networkClient.submitTransaction(transaction2);
+const result = await programManager.networkClient.submitTransaction(transaction);
 ```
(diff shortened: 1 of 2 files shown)


docs/examples/02_transfer_private.md
```diff
@@ -19,14 +19,16 @@ const programManager = new ProgramManager("https://api.explorer.provable.com/v1"
...
-});
+const transaction = await programManager
+    .buildTransferTransaction(
+        5,
+        account
+            .address()
+            .to_string(),
+        "publicToPrivate",
+        0.1,
+        false,
```
iamalwaysuncomfortable: Write a comment in front of each parameter, also ensure the transfer amount is more than 1/100 of an Aleo credit.

docs/examples/02_transfer_private.md
```diff
@@ -52,14 +54,16 @@ const recipient = new Account();
...
+const transaction2 = await programManager
+    .buildTransferTransaction(
+        5,
+        recipient
+            .address()
+            .to_string(),
+        "private",
+        0.1,
+        false
+    );
```
iamalwaysuncomfortable: Same comments here.

docs/examples/02_transfer_private.md
alexpitsikoulis: done

docs/examples/02_transfer_private.md
alexpitsikoulis: done

docs/examples/01_transfer_public.md
```diff
@@ -21,17 +21,17 @@ programManager.setAccount(account);
 // Create recipient account.
 const recipient = new Account();
 
-// Execute `transfer_public` function in `credits.aleo`
+// Build a public transfer transaction.
```
iamalwaysuncomfortable:
```suggestion
// Build a transfer_public transaction.
```

I'd say that either we need to describe the exact function being executed or describe semantically what's happening here so that users have full context on what's happening. 

I think a description of that would say something to the effect of: "Build a transaction that publicly transfers Aleo Credits by invoking the `transfer_public` function" would give the full context, but I think you should take your own interpretation on how to phrase this.

iamalwaysuncomfortable (commented):
One final nit and then this is approved.

docs/examples/01_transfer_public.md
alexpitsikoulis: done

PR #985, opened 2025-03-04
Fix Create Leo App's offline public transaction signing template

christianwwwwwwww:
Makes updates to the `create-leo-app` offline public transaction signing template. The SDK also had to be modified to remove the `stakerAddress` from `bond_public`


iamalwaysuncomfortable (commented):
Thanks so much for finding and fixing this!

iamalwaysuncomfortable (approved):

PR #989, opened 2025-03-07
[Feature] Add BHP, Pedersen, and Poseidon hash function implementations to the Provable SDK.

iamalwaysuncomfortable:
## Motivation

This PR adds BHP, Pedersen and Poseidon hash function impls to Provable SDK. This proposed functionality enables JS developers to use the Aleo implementations of these hash functions within their dapps.

These hash exports are useful (but not limited to) the following use-cases.
* **Deriving record and ciphertext view keys for private key owners:** Some apps such as wallet impls may want to be able to view plaintexts of ciphertexts on chain on behalf of a private key holder in order to execute critical functionality based on that hidden value.
* **Implementing client side cryptography:** Apps may want to be able to implement hash-function based cryptography client-side compatible with their private dapps such as creating private randomness, decoding elements that have been encoded within a circuit or `finalize` block, etc.
* **Research:** Exporting hash functions for researchers who desire to experiment with BHP, Pedersen or Poseidon hash functions.

## Changelog

* Implements individual hashing objects `BHP256`, `BHP512`, `BHP768`, `BHP1024`, `Pedersen64`, `Pedersen128`, `Poseidon2`, `Poseidon4`, and `Poseidon8`. The current rationale behind this is to enable different domain separators to ensure dapps implementing these hash functions within offchain protocols in their dapp do not share a global domain (which may lead to vulnerabilities such as witness discovery).
* Adds a `utilities` module that provides macros for converting `js arrays` of protocol objects to vectors of corresponding objects in `rust` & `wasm` 
* Implements convenience methods on the `Ciphertext` object to enable decryption of ciphertexts from transition inputs or outputs by the key `ViewKey` holder.

## Test Plan

* Adds tests within the wasm SDK to ensure the exported wasm implementations match the native hashing impls.
* Adds test vectors within the JS SDK to ensure hashing within javascript matches expected outputs


jaketarnow (approved):
LGTM

wasm/src/utilities/array.rs
```diff
@@ -0,0 +1,42 @@
...
+// You should have received a copy of the GNU General Public License
+// along with the Provable SDK library. If not, see <https://www.gnu.org/licenses/>.
+
+/// Converts a JS array of wasm-bindgen exported structs from JsValues to their native Rust type
+/// representation.
+#[macro_export]
+macro_rules! from_wasm_object_array {
+    ($input:expr, $wasm_type:ident) => {{
+        $input
+            .to_vec()
```
Pauan: I think this can use `.iter()` instead of `.to_vec().into_iter()`

wasm/src/utilities/array.rs
```diff
@@ -0,0 +1,42 @@
...
+            .collect::<Result<Vec<FieldNative>, String>>()
+    }};
+}
+
+/// Converts an array of JS types to a vector of Rust analog types.
+#[macro_export]
+macro_rules! from_js_typed_array {
+    ($input:expr, $method:ident, $_type:expr) => {{
+        $input
+            .to_vec()
```
Pauan: I think this can use `.iter()` instead of `.to_vec().iter()`

wasm/src/utilities/bits.rs
```diff
@@ -0,0 +1,26 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with the Provable SDK library. If not, see <https://www.gnu.org/licenses/>.
+
+#[macro_export]
+macro_rules! to_bits_array_le {
+    ($self:expr) => {{
+        let array = js_sys::Array::new();
```
Pauan: It's better to use `.collect::<js_sys::Array>` instead of manually pushing.

Pauan (dismissed):
A few minor comments

wasm/src/algorithms/poseidon/poseidon2.rs
```diff
@@ -0,0 +1,77 @@
...
+        let input = from_wasm_object_array!(input, Field)?;
+        self.0.hash_many(&input, num_outputs).into_iter().for_each(|field| {
+            array.push(&JsValue::from(Field::from(field)));
+        });
+        Ok(array)
+    }
+
+    /// Returns the Poseidon hash with an input rate of 2 on the scalar field.
+    #[wasm_bindgen(js_name = "hashToScalar")]
+    pub fn hash_to_scalar(&self, input: Array) -> Result<Scalar, String> {
```
iamalwaysuncomfortable, line 66: @Pauan all functions within the `Poseidon` hashers use the TryFromJsValue trait. They iterate through an array of exported structs (the members of the array currently are all `Field` type, but this could be generalized to any type)  and call TryFromJsValue which moves the data from those structs. Is there any way to simply copy the data instead leaving the original members of the array intact so that it behaves more similarly to true JS objects?

iamalwaysuncomfortable (commented):
@Pauan one question above about taking Arrays of exported structs.

PR #994, opened 2025-03-17
[Operations] Create SDK release github action

iamalwaysuncomfortable:
## Motivation

In order to provide end users with full information on what is included in a release, this github action is introduced to zip up all SDK builds as well as the source code, allowing users to download and inspect the sources.

`Windows` and `OSX-ARM` support should be added before this PR is merged.

## Test Plan

Ensure all releases are built properly before merging.


iamalwaysuncomfortable (commented):
Since both JS and wasm are portable, this theoretically should be able to be built on a single architecture. When this is revisited, we only need a single runner that:

* Builds the wasm and zips the dist
* Builds the SDK and zips the dist
* Build create-leo-app and zips the dist
* Zips the source code

PR #996, opened 2025-03-18
Feat/getter function for view key

### LINKED ISSUE #995: [Feature] Add get transition view key function into sdk
## 🚀 Feature

Add get transition view key function into sdk

## Motivation

There is a function to decrypt ciphertext with transition view key but there is no implementation to get transition view key to use this function.
https://github.com/ProvableHQ/sdk/blob/ad1eb95c6a13bb9173250df77b09b8b598f0dba6/wasm/src/programs/data/ciphertext.rs#L56

## Implementation

Add get transition view key function into sdk to enable usage of decrypt and encrypt non-record ciphertext.

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))
Yes

Roee-87:
## Motivation

This PR closes [Issue 995](https://github.com/ProvableHQ/sdk/issues/995).  A method `tvk()` is added to the impl block for the `Transition` struct that returns a transition view key.  This method takes a reference to a `ViewKey` as an input.

## Test Plan

I added an `assert_eq` to the `test_transition_helpers` test function which checks that the newly added `tvk()` method returns the correct transition view key. (NOTE:  Placeholder string used instead of the calculated value.  Once I know how to manually compute the transition view key, I'll mark this PR as ready for review.)


wasm/src/ledger/transition.rs
```diff
@@ -274,6 +285,11 @@ mod tests {
             Field::from_string("6845182532650964173356391969005331370591444046632036068754797772530920467754field")
                 .unwrap()
         );
+        assert_eq!(
+            transition.tvk(&view_key),
+            Field::from_string("_placeholder_") // Placeholder for the view key, will replace with manually computed value.
```
iamalwaysuncomfortable: You can make this value the actual tvk, we tend to not work with any non-valid values in the test unless we're specifically trying to fail the test.

iamalwaysuncomfortable (commented):
This looks good minus the test, once that's updated, this can be merged.

wasm/src/ledger/transition.rs
Roee-87: Done:  [7a9aab9](https://github.com/ProvableHQ/sdk/pull/996/commits/7a9aab93013d69cbe01173937efd028c6850d147)

wasm/src/ledger/transition.rs
```diff
@@ -176,6 +176,13 @@ impl Transition {
         Group::from(self.0.tpk())
     }
 
+    /// Get the transition view key of the transition.
+    pub fn tvk(&self, view_key: &ViewKey) -> Field {
+        let tpk = self.tpk();
+        let tvk = tpk.scalar_multiply(&view_key.to_scalar()).to_x_coordinate();
+        tvk
```
iamalwaysuncomfortable: We should make it a bit more concise
```suggestion
        tpk.scalar_multiply(&view_key.to_scalar()).to_x_coordinate()
```

iamalwaysuncomfortable (commented):
One minor suggestion and then this is good to go.

iamalwaysuncomfortable (approved):

PR #998, opened 2025-03-18
[Tests] create-leo-app e2e Playwright testing

alexpitsikoulis:
## Motivation

This PR introduces build tests for create-leo-app to ensure future changes to the SDK do not break the templates.

## Test Plan

Run dev server for each template and use playwright to visit the webpage.

## Related PRs

N/A


jaketarnow (changes requested):
I would add a check to ensure the user has `playwright` installed via `npx playwright install `. After that, can you add in the script of `npx playwright show-report` so it displays the results in the browser as well. Thanks

create-leo-app/template-react-leo/package.json
```diff
@@ -21,6 +22,8 @@
     "@babel/core": "^7.26.7",
     "@babel/preset-env": "^7.26.7",
     "@babel/preset-react": "^7.26.3",
+    "@playwright/test": "^1.51.1",
```
iamalwaysuncomfortable, line 25: Does this need to be in dependencies, or should be it in devDependencies. Doing it directly in dependencies will unnecessarily increase the size of the template for anyone trying to build off it in production. If we can move it to DevDependencies for all templates, this will be much better.

create-leo-app/template-nextjs-ts/.gitignore
```diff
@@ -0,0 +1,7 @@
+
+# Playwright
+node_modules/
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
```
iamalwaysuncomfortable, line 7: The .gitignore should be able to be moved up into the main .gitignore unless the goal is to ensure people have gitignores when they create the template?

create-leo-app/template-extension/package.json
```diff
@@ -4,7 +4,8 @@
   "version": "0.0.0",
   "type": "module",
   "scripts": {
-    "build": "rimraf static/js && rollup --config"
+    "build": "rimraf static/js && rollup --config",
+    "test:e2e": "npm run build"
```
iamalwaysuncomfortable, line 8: Will this run the tests on every build?

create-leo-app/template-vanilla/playwright.config.js
```diff
@@ -0,0 +1,83 @@
+// @ts-check
+import { defineConfig, devices } from '@playwright/test';
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// import dotenv from 'dotenv';
+// import path from 'path';
+// dotenv.config({ path: path.resolve(__dirname, '.env') });
```
iamalwaysuncomfortable, line 10: Any reason to have these still?

create-leo-app/template-react-leo/playwright.config.js
```diff
@@ -0,0 +1,81 @@
+// @ts-check
```
iamalwaysuncomfortable, line 1: Do the config files need to be included in every project @onetrickwolf - do you know if this is necessary? It would be cool to have it in one place, but if it's needed in every project than so be it.

iamalwaysuncomfortable (commented):
Some nits, but this is looking good, let me know the answers to the nits below and if we can address them.

create-leo-app/template-react-leo/playwright.config.js
alexpitsikoulis: the reason i did that is so that when the templates are instantiated they will have the config. i intended for this testing framework to be extensible by anyone who may be developing off of them

create-leo-app/template-vanilla/playwright.config.js
alexpitsikoulis: this is all autogenerated, the only thing ive changed manually in them is the timeout

create-leo-app/template-nextjs-ts/.gitignore
alexpitsikoulis: yes that is the idea. this is generated by playwright init and just includes the files associated with test runs

create-leo-app/template-extension/package.json
alexpitsikoulis: no, since this template isnt frontend we wont take the same approach to testing it. so here and in the node templates it just runs them to ensure they build without error. which is why `test:e2e` is just the same as the build script. i just needed all the templates to have `test:e2e` so that they can be iteratively run using the github action

create-leo-app/template-react-leo/package.json
alexpitsikoulis: it is in dev dependencies, its just not obvious the way that github has collapsed the unchanged lines

iamalwaysuncomfortable (commented):
This looks good from my end, however playwright e2e tests are new to me, so I'd like request review from someone experienced in the subject. @dms120 can you review this?

create-leo-app/template-vanilla/package.json
```diff
@@ -1,16 +1,19 @@
...
-    }
+  "name": "aleo-starter",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "vite build",
+    "preview": "vite preview",
+    "test:e2e": "sh scripts/test.sh"
```
onetrickwolf, line 10: What is the point of having this be a script? Would prefer to get rid of this across all templates unless there's a good reason. Would just replace this with: `"test:e2e": "playwright test --headed"`

create-leo-app/template-nextjs-ts/tests/example.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { test, expect } from '@playwright/test';
+
+test.beforeEach(async ({ page }) => {
+  await page.goto('http://localhost:3000/');
+});
+
+test('builds', async ({ page }) => {
```
onetrickwolf, line 7: this file seems incomplete?

create-leo-app/template-offline-public-transaction-ts/.gitignore
```diff
@@ -0,0 +1,7 @@
+
```
onetrickwolf, line 1: these .gitignores should not exist this repo is supposed to for generating templates this will likely break the existing mechanism.

create-leo-app/template-react-managed-worker/tests/example.spec.js
```diff
@@ -0,0 +1,38 @@
...
+
+test.beforeEach(async ({ page }) => {
+  await page.goto('http://localhost:5173/');
+  page.on('dialog', dialog => dialog.accept());
+});
+
+test('builds', async ({ page }) => {
+    const accountButton = await page.getByTestId("generate-account-button");
+    await expect(accountButton).toHaveText("Click to generate account");
+    await page.waitForTimeout(1000);
```
onetrickwolf, line 12: Would prefer not to have any timeouts. Should be waiting for something specific.

package.json
```diff
@@ -36,5 +36,6 @@
   "devDependencies": {
     "prettier": "3.4.2",
     "wasm-pack": "^0.13.1"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
onetrickwolf, line 40: What's the point of this?

create-leo-app/template-react-leo/playwright.config.js
onetrickwolf: Yeah if that's what we want to do that's fine, I think that is maybe a very different scope though we may want to consider. Do we want these tests to validate the templates or do we want to maintain testing templates people can grow off of?

onetrickwolf (changes requested):
Mostly nitpicks but would maybe like to have a larger conversation on how we want this to go long term conceptional we might want some changes.

create-leo-app/template-vanilla/package.json
alexpitsikoulis: so yeah the reason for this is that the dev server needs to be running so the script runs that detached and then stops it

create-leo-app/template-nextjs-ts/tests/example.spec.ts
alexpitsikoulis: i couldn't get this one to react to clicks at all so at least for now this is left as a test just to ensure it will build and run

package.json
alexpitsikoulis: ah this was included accidentally ill remove it

PR #1000, opened 2025-03-19
[Example] Zk Auction

iamalwaysuncomfortable:
## Motivation

This PR adds an example of the [auction.aleo](https://play.leo-lang.org/) program in order to provide a working example of private state management with the SDK. This example has a page each for `auctioneer` and `bidder`. The bidder can submit bids via Transactions submitted by the Leo Wallet, the auctioneer through usage of the Leo Wallet can then find the records created and display them.

This example is meant to showcase usage of private state.

## Changelog
* Addition of the zk auction example

## Test Plan
* This example should be integrated into Cypress/Playwright integration tests.

## Screenshots
Screenshot 2025-03-19 at 4 14 39 PM
Screenshot 2025-03-19 at 4 15 37 PM
Screenshot 2025-03-19 at 4 17 44 PM
Screenshot 2025-03-19 at 4 19 02 PM


create-leo-app/template-auction/src/components/AuctionState.jsx
```diff
@@ -0,0 +1,148 @@
...
+export const AuctionState = ({ children }) => {
+    const [auctionState, setAuctionState] = useState({
+        auctioneerRecords: [], // Raw records from the chain
+        auctioneerState: {}, // Organized by auction ID
+        bidderRecords: [], // Raw records from the chain
+        bidderState: {}, // Organized by bidder address
+        winningBids: {}, // Track winning bids by auction ID
+    });
+
+    const addNewBid = (amount, auctioneer, bidder, id, txId) => {
```
dms120, line 25: I would initialize the bid before the sets just for better readability: `const newBid = { amount, auctioneer, bidder, id, txId, isWinner: false };`

And then:

```
 bids: [
      ...(prevState.auctioneerState[id]?.bids || []),
      newBid,
  ],
```

create-leo-app/template-auction/src/components/AuctionState.jsx
```diff
@@ -0,0 +1,148 @@
...
+        return auctionState.auctioneerRecords.find(record => record.id === recordId);
+    }
+
+    // Find all auctioneer records for a given auction ID.
+    const findAuctioneerRecordsByAuctionId = (auctionId) => {
+        return  auctionState.auctioneerRecords.filter(record => record.data.id === `${auctionId}.private`);
+    }
+
+    // Find all unspent auctioneer records.
+    const findAllUnspentAuctionRecords = () => {
```
dms120, line 111: Would probably wrap some of this with useMemo() to avoid unnecessary recalculations

create-leo-app/template-auction/src/components/AuctionState.jsx
dms120, line 1: I'm not a super fan of ReactContext given his verbosity and performance issues. There are really cool libs like https://jotai.org/ that allows to use atomic state that simplifies this kind of state management and avoids unnecessary re-rendering out of the box. Just a note because React Context does it job ofc

dms120 (approved):

dms120:
Great job @iamalwaysuncomfortable ! Apart of usage of ReactContext and a slightly different component segregation, I wouldn't do it much better 😄 

I was not able to initialize the app because I was having installation errors, so I couldn't see it working. Probably would have something more to add in that case. Need to steal some minutes from someone from the team to get it working.

jaketarnow:
> Great job @iamalwaysuncomfortable ! Apart of usage of ReactContext and a slightly different component segregation, I wouldn't do it much better 😄
> 
> I was not able to initialize the app because I was having installation errors, so I couldn't see it working. Probably would have something more to add in that case. Need to steal some minutes from someone from the team to get it working.

If I try to run the app via `yarn && yarn build` then it fails to build. If I run it purely from `yarn dev` it builds. However, I have no idea how to interact with the app.

1. After connecting my wallet, I can't determine if an auction is still taking place or how to start a new auction.
2. Am I the `Auctioneer` by default or am I always a `Bidder`? If I am a bidder - how do I derive the `Auction ID` for placing a bid. 
3. There should be a "Disconnect" wallet button present instead of having it disappear after connecting.

PR #1001, opened 2025-03-20
[Feature] Plaintext toBitsLe method

alexpitsikoulis:
## Motivation

In order to hash a plaintext struct, one needs to be able to get the bit serialization of it. This PR exposes the bit serialization method from the WASM so that it may be provided to the hash function.

## Test Plan

N/A

## Related PRs

N/A

### DIFF
--- a/wasm/src/account/address.rs
+++ b/wasm/src/account/address.rs
@@ -16,8 +16,10 @@
 
 use crate::account::{PrivateKey, Signature, ViewKey};
 
-use crate::{account::compute_key::ComputeKey, types::native::AddressNative};
+use crate::{account::compute_key::ComputeKey, to_bits_array_le, types::native::AddressNative};
 use core::{convert::TryFrom, fmt, ops::Deref, str::FromStr};
+use js_sys::Array;
+use snarkvm_wasm::utilities::ToBits;
 use wasm_bindgen::prelude::*;
 
 /// Public address of an Aleo account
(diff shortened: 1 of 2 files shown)


jaketarnow (changes requested):
Please run `cargo fmt` to adhere to proper formatting

wasm/src/account/address.rs
```diff
@@ -16,8 +16,10 @@
 
 use crate::account::{PrivateKey, Signature, ViewKey};
 
-use crate::{account::compute_key::ComputeKey, types::native::AddressNative};
+use crate::{account::compute_key::ComputeKey, to_bits_array_le, types::native::AddressNative};
 use core::{convert::TryFrom, fmt, ops::Deref, str::FromStr};
+use js_sys::Array;
+use snarkvm_wasm::utilities::ToBits;
```
iamalwaysuncomfortable: Group the crate imports together and ensure it's:

crate imports
snarkvm/snarkos imports

wasm/src/programs/data/plaintext.rs
```diff
@@ -108,6 +110,12 @@ impl Plaintext {
         Ok(Uint8Array::from(rust_bytes.as_slice()))
     }
 
+    /// Get the left endian boolean array representation of the plaintext.
+    #[wasm_bindgen(js_name = "toBitsLe")]
+    pub fn to_bits_le(&self) -> Array {
+        to_bits_array_le!(self)
+    }
+
```
iamalwaysuncomfortable, line 117: We should create from_bits_le, from_fields, and to_fields on this methods and on the ciphertext object too.

PR #1010, opened 2025-03-27
[Feature] Expose bit, byte, and fields serialization methods for plaintext, ciphertext, records, and addresses

### LINKED ISSUE #1003: [Feature] Add type casts for hash functions inputs
## 🚀 Feature

The hash functions are expected input in the form of array of bits, but there is no clear method on how to type cast all supported types on Leo to array of bits in order to use these functions, especially struct and array types. https://github.com/ProvableHQ/sdk/blob/ba3cea3752313b2bd3e8dd54514e4971a9e8cd15/wasm/src/algorithms/bhp/bhp256.rs#L42-#L43

And for Poseidon hash it seems the inputs have to be array of Field instead of bits but there is no notes around this.
https://github.com/ProvableHQ/sdk/blob/ba3cea3752313b2bd3e8dd54514e4971a9e8cd15/wasm/src/algorithms/poseidon/poseidon2.rs#L48-#L49

## Motivation

Able to cast all supported types into compatible input type for the hash functions in the SDK.

## Implementation

Probably something like `type_to_bits_array()` and `type_to_field_array()` should do the job.

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))

Yes

alexpitsikoulis:
## Motivation

Currently there is not an ergonomic way to get serialized representations of plaintext and ciphertext objects. Adding these methods opens up client-side hashing on addresses and structs in their plaintext and ciphertext representations.

## Test Plan

Include unit tests that compare outputs of these methods with their expected outputs.

## Related PRs

### DIFF
--- a/wasm/src/account/address.rs
+++ b/wasm/src/account/address.rs
@@ -16,8 +16,10 @@
 
 use crate::account::{PrivateKey, Signature, ViewKey};
 
-use crate::{account::compute_key::ComputeKey, types::native::AddressNative};
+use crate::{account::compute_key::ComputeKey, to_bits_array_le, types::native::AddressNative};
 use core::{convert::TryFrom, fmt, ops::Deref, str::FromStr};
+use js_sys::Array;
+use snarkvm_wasm::utilities::ToBits;
 use wasm_bindgen::prelude::*;
 
 /// Public address of an Aleo account
@@ -65,6 +67,12 @@ impl Address {
     #[allow(clippy::inherent_to_string_shadow_display)]
     pub fn to_string(&self) -> String {
         self.0.to_string()
+    }
+
+    /// Get the left endian boolean array representation of the address.
+    #[wasm_bindgen(js_name = "toBitsLe")]
+    pub fn to_bits_le(&self) -> Array {
+        to_bits_array_le!(self)
     }
 
     /// Verify a signature for a message signed by the address
--- a/wasm/src/account/address.rs
+++ b/wasm/src/account/address.rs
@@ -17,10 +17,11 @@
 use crate::{
     Plaintext,
     account::{PrivateKey, Signature, ViewKey, compute_key::ComputeKey},
+    native::LiteralNative,
     types::native::AddressNative,
 };
 use core::{convert::TryFrom, fmt, ops::Deref, str::FromStr};
-use js_sys::Array;
+use js_sys::{Array, Uint8Array};
 use wasm_bindgen::prelude::*;
 
 /// Public address of an Aleo account
@@ -70,11 +71,18 @@ impl Address {
         self.0.to_string()
     }
 
-    /// Get the left endian boolean array representation of the address plaintext.
-    #[wasm_bindgen(js_name = "plaintextBitsLe")]
-    pub fn plaintext_bits_le(&self) -> Result<Array, String> {
-        let plaintext = Plaintext::from_string(&self.to_string())?;
-        Ok(plaintext.to_bits_le())
+    /// Get the left endian byte array representation of the address plaintext.
+    #[wasm_bindgen(js_name = "toPlaintextBytesLe")]
+    pub fn to_plaintext_bytes_le(&self) -> Result<Uint8Array, String> {
+        let plaintext = Plaintext::from(LiteralNative::Address(self.0));
+        plaintext.to_bytes_le()
+    }
+
+    /// Get the left endian boolean array representation of the address plaintext bits.
+    #[wasm_bindgen(js_name = "toPlaintextBitsLe")]
+    pub fn to_plaintext_bits_le(&self) -> Array {
+        let plaintext = Plaintext::from(LiteralNative::Address(self.0));
+        plaintext.to_bits_le()
     }
 
     /// Verify a signature for a message signed by the address
--- a/wasm/src/account/address.rs
+++ b/wasm/src/account/address.rs
@@ -14,12 +14,13 @@
 // You should have received a copy of the GNU General Public License
 // along with the Provable SDK library. If not, see <https://www.gnu.org/licenses/>.
 
-use crate::account::{PrivateKey, Signature, ViewKey};
-
-use crate::{account::compute_key::ComputeKey, to_bits_array_le, types::native::AddressNative};
+use crate::{
+    Plaintext,
+    account::{PrivateKey, Signature, ViewKey, compute_key::ComputeKey},
+    types::native::AddressNative,
+};
 use core::{convert::TryFrom, fmt, ops::Deref, str::FromStr};
 use js_sys::Array;
-use snarkvm_wasm::utilities::ToBits;
 use wasm_bindgen::prelude::*;
 
 /// Public address of an Aleo account
@@ -69,10 +70,11 @@ impl Address {
         self.0.to_string()
     }
 
-    /// Get the left endian boolean array representation of the address.
-    #[wasm_bindgen(js_name = "toBitsLe")]
-    pub fn to_bits_le(&self) -> Array {
-        to_bits_array_le!(self)
+    /// Get the left endian boolean array representation of the address plaintext.
+    #[wasm_bindgen(js_name = "plaintextBitsLe")]
+    pub fn plaintext_bits_le(&self) -> Result<Array, String> {
+        let plaintext = Plaintext::from_string(&self.to_string())?;
+        Ok(plaintext.to_bits_le())
     }
 
     /// Verify a signature for a message signed by the address
--- a/wasm/src/account/address.rs
+++ b/wasm/src/account/address.rs
@@ -16,6 +16,7 @@
 
 use crate::{
     Field,
+    Group,
     Plaintext,
     account::{PrivateKey, Signature, ViewKey, compute_key::ComputeKey},
     from_js_typed_array,
@@ -25,7 +26,6 @@ use crate::{
     to_bits_array_le,
     types::native::AddressNative,
 };
-
 use snarkvm_console::prelude::{FromBits, FromBytes, FromFields, ToBits, ToBytes, ToFields};
 
 use core::{convert::TryFrom, fmt, ops::Deref, str::FromStr};
@@ -119,6 +119,22 @@ impl Address {
         Ok(js_array_from_fields!(&native_fields))
     }
 
+    /// Get an address object from a group.
+    ///
+    /// @param {Group} group The group object.
+    ///
+    /// @returns {Address} The address object.
+    #[wasm_bindgen(js_name = "fromGroup")]
+    pub fn from_group(group: Group) -> Self {
+        Self::from(group)
+    }
+
+    /// Get the group representation of the address object.
+    #[wasm_bindgen(js_name = "toGroup")]
+    pub fn to_group(&self) -> Group {
+        Group::from(self)
+    }
+
     /// Create an aleo address object from a string representation of an address
     ///
     /// @param {string} address String representation of an addressm
@@ -186,6 +202,31 @@ impl From<&AddressNative> for Address {
 impl From<&Address> for AddressNative {
     fn from(value: &Address) -> Self {
         value.0
+    }
+}
+
+impl From<AddressNative> for Group {
+    fn from(value: AddressNative) -> Self {
+        let vm_group = value.to_group();
+        Self::from(vm_group)
+    }
+}
+
+impl From<Address> for Group {
+    fn from(value: Address) -> Self {
+        Self::from(value.0)
+    }
+}
+
+impl From<&AddressNative> for Group {
+    fn from(value: &AddressNative) -> Self {
+        Self::from(value.clone())
+    }
+}
+
+impl From<&Address> for Group {
+    fn from(value: &Address) -> Self {
+        Self::from(value.0)
     }
 }
 
--- a/wasm/src/account/address.rs
+++ b/wasm/src/account/address.rs
@@ -142,24 +142,6 @@ impl Address {
         Plaintext::from(LiteralNative::Address(self.0))
     }
 
-    /// Get the left endian byte array representation of the address plaintext.
-    #[wasm_bindgen(js_name = "toPlaintextBytesLe")]
-    pub fn to_plaintext_bytes_le(&self) -> Result<Uint8Array, String> {
-        self.to_plaintext().to_bytes_le()
-    }
-
-    /// Get the left endian boolean array representation of the address plaintext bits.
-    #[wasm_bindgen(js_name = "toPlaintextBitsLe")]
-    pub fn to_plaintext_bits_le(&self) -> Array {
-        self.to_plaintext().to_bits_le()
-    }
-
-    /// Get the field array representation of the address plaintext.
-    #[wasm_bindgen(js_name = "toPlaintextFields")]
-    pub fn to_plaintext_fields(&self) -> Result<Array, String> {
-        self.to_plaintext().to_fields()
-    }
-
     /// Verify a signature for a message signed by the address
     ///
     /// @param {Uint8Array} Byte array representing a message signed by the address
--- a/wasm/src/account/address.rs
+++ b/wasm/src/account/address.rs
@@ -15,14 +15,22 @@
 // along with the Provable SDK library. If not, see <https://www.gnu.org/licenses/>.
 
 use crate::{
+    Field,
     Plaintext,
     account::{PrivateKey, Signature, ViewKey, compute_key::ComputeKey},
-    native::LiteralNative,
+    from_js_typed_array,
+    from_wasm_object_array,
+    js_array_from_fields,
+    native::{FieldNative, LiteralNative},
+    to_bits_array_le,
     types::native::AddressNative,
 };
+
+use snarkvm_console::prelude::{FromBits, FromBytes, FromFields, ToBits, ToBytes, ToFields};
+
 use core::{convert::TryFrom, fmt, ops::Deref, str::FromStr};
 use js_sys::{Array, Uint8Array};
-use wasm_bindgen::prelude::*;
+use wasm_bindgen::{convert::TryFromJsValue, prelude::*};
 
 /// Public address of an Aleo account
 #[wasm_bindgen]
@@ -54,6 +62,63 @@ impl Address {
         compute_key.address()
     }
 
+    /// Get an address from a series of bytes.
+    ///
+    /// @param {Uint8Array} bytes A left endian byte array representing the address.
+    ///
+    /// @returns {Address} The address object.
+    #[wasm_bindgen(js_name = "fromBytesLe")]
+    pub fn from_bytes_le(bytes: Uint8Array) -> Result<Self, String> {
+        let rust_bytes = bytes.to_vec();
+        let native = AddressNative::from_bytes_le(rust_bytes.as_slice()).map_err(|e| e.to_string())?;
+        Ok(Self(native))
+    }
+
+    /// Get the left endian byte array representation of the address.
+    #[wasm_bindgen(js_name = "toBytesLe")]
+    pub fn to_bytes_le(&self) -> Result<Uint8Array, String> {
+        let rust_bytes = self.0.to_bytes_le().map_err(|e| e.to_string())?;
+        Ok(Uint8Array::from(rust_bytes.as_slice()))
+    }
+
+    /// Get an address from a series of bits represented as a boolean array.
+    ///
+    /// @param {Array} bits A left endian boolean array representing the bits of the address.
+    ///
+    /// @returns {Address} The address object.
+    #[wasm_bindgen(js_name = "fromBitsLe")]
+    pub fn from_bits_le(bits: Array) -> Result<Self, String> {
+        let rust_bits = from_js_typed_array!(bits, as_bool, "boolean")?;
+        let native = AddressNative::from_bits_le(&rust_bits).map_err(|e| e.to_string())?;
+        Ok(Self(native))
+    }
+
+    /// Get the left endian boolean array representation of the bits of the address.
+    #[wasm_bindgen(js_name = "toBitsLe")]
+    pub fn to_bits_le(&self) -> Array {
+        to_bits_array_le!(self)
+    }
+
+    /// Get an address object from an array of fields.
+    ///
+    /// @param {Array} fields An array of fields.
+    ///
+    /// @returns {Plaintext} The address object.
+    #[wasm_bindgen(js_name = "fromFields")]
+    pub fn from_fields(fields: Array) -> Result<Self, String> {
+        let native_fields = from_wasm_object_array!(fields, Field)?;
+        let native = AddressNative::from_fields(&native_fields).map_err(|e| e.to_string())?;
+        Ok(Self(native))
+    }
+
+    /// Get the field array representation of the address.
+    #[wasm_bindgen(js_name = "toFields")]
+    pub fn to_fields(&self) -> Result<Array, String> {
+        let native = self.0.clone();
+        let native_fields = native.to_fields().map_err(|e| e.to_string())?;
+        Ok(js_array_from_fields!(&native_fields))
+    }
+
     /// Create an aleo address object from a string representation of an address
     ///
     /// @param {string} address String representation of an addressm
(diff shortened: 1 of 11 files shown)


wasm/src/account/signature.rs
```diff
@@ -76,6 +77,20 @@ impl Signature {
...
+
+    /// Get the left endian byte array representation of the signature plaintext.
+    #[wasm_bindgen(js_name = "toPlaintextBytesLe")]
+    pub fn to_plaintext_bytes_le(&self) -> Result<Uint8Array, String> {
+        let plaintext = Plaintext::from(LiteralNative::Signature(Box::new(self.0)));
+        plaintext.to_bytes_le()
+    }
+
+    /// Get the left endian boolean array representation of the signature plaintext bits.
+    #[wasm_bindgen(js_name = "toPlaintextBitsKeLe")]
```
jaketarnow: Please fix misspell to `toPlaintextBitsLe`

jaketarnow (changes requested):

wasm/src/account/signature.rs
alexpitsikoulis: yikes good catch, fixed

wasm/src/programs/data/plaintext.rs
```diff
@@ -170,6 +190,19 @@ impl From<&Plaintext> for PlaintextNative {
     }
 }
 
+impl From<LiteralNative> for Plaintext {
+    fn from(value: LiteralNative) -> Self {
+        let native = PlaintextNative::Literal(value, OnceCell::new());
+        Self(native)
+    }
+}
```
iamalwaysuncomfortable, line 220: Nice touch

wasm/src/programs/data/plaintext.rs
```diff
@@ -99,15 +103,31 @@ impl Plaintext {
...
+        let native = PlaintextNative::from_bits_le(&rust_bits).map_err(|e| e.to_string())?;
+        Ok(Self(native))
+    }
+
+    /// Get the left endian boolean array representation of the bits plaintext.
+    #[wasm_bindgen(js_name = "toBitsLe")]
+    pub fn to_bits_le(&self) -> Array {
+        to_bits_array_le!(self)
+    }
+
```
iamalwaysuncomfortable: Can you also add the toFields and fromFields methods here? This is important as we'll need those for Poseidon hashing.

These utility macros can assist you in building these methods:
https://github.com/ProvableHQ/sdk/blob/mainnet/wasm/src/utilities/array.rs

wasm/src/programs/data/plaintext.rs
```diff
@@ -19,13 +19,17 @@ use crate::{
...
     Scalar,
+    from_js_typed_array,
+    native::LiteralNative,
     plaintext_to_js_value,
+    to_bits_array_le,
     types::native::{FromBytes, IdentifierNative, PlaintextNative, ToBytes},
 };
-use std::ops::Deref;
+use once_cell::sync::OnceCell;
+use snarkvm_console::prelude::{FromBits, ToBits};
```
iamalwaysuncomfortable: We'll define "EcoSystem" imports as imports from `snarkvm` or `crate` (i.e. SDK imports) and "Foreign" imports as any other crate.

For ecosystem imports, `crate` imports should go first and `snarkvm` imports should go second after crate without a space. These should all go alphabetically.

Foreign imports but should be grouped together alphabetically as well and should begin after EcoSystem imports with a single space.

```suggestion
use snarkvm_console::prelude::{FromBits, ToBits};
```

wasm/src/programs/data/plaintext.rs
```diff
@@ -19,13 +19,17 @@ use crate::{
...
 
-use js_sys::Uint8Array;
-use std::str::FromStr;
+use js_sys::{Array, Uint8Array};
+use std::{ops::Deref, str::FromStr};
```
iamalwaysuncomfortable:
```suggestion
use js_sys::{Array, Uint8Array};
use once_cell::sync::OnceCell;
use std::{ops::Deref, str::FromStr};
```

wasm/src/programs/data/ciphertext.rs
```diff
@@ -117,6 +123,31 @@ impl Ciphertext {
...
+        let rust_bits = from_js_typed_array!(bits, as_bool, "boolean")?;
+        let native = CiphertextNative::from_bits_le(&rust_bits).map_err(|e| e.to_string())?;
+        Ok(Ciphertext(native))
+    }
+
+    /// Get the left endian boolean array representation of the bits of the ciphertext.
+    #[wasm_bindgen(js_name = "toBitsLe")]
+    pub fn to_bits_le(&self) -> Array {
+        to_bits_array_le!(self)
+    }
```
iamalwaysuncomfortable, line 151: Can you add the `toFields` and `fromFields` method for `CipherText` as well?

wasm/src/record/record_ciphertext.rs
```diff
@@ -73,6 +76,20 @@ impl RecordCiphertext {
...
+        let bytes_vec = self.0.to_bytes_le().map_err(|e| e.to_string())?;
+        let bytes = bytes_vec.as_slice();
+        Uint8Array::try_from(bytes).map_err(|e| e.to_string())
+    }
+
+    /// Get the left endian boolean array representation of the record ciphertext bits.
+    #[wasm_bindgen(js_name = "toBitsLe")]
+    pub fn to_bits_le(&self) -> Array {
+        to_bits_array_le!(self)
+    }
```
iamalwaysuncomfortable, line 111: Can you add the `fromBytesLe`, `toFields` here as well?

wasm/src/record/record_plaintext.rs
```diff
@@ -155,6 +157,24 @@ impl RecordPlaintext {
...
+    }
+
+    /// Returns the left endian boolean array representation of the record plaintext bits.
+    ///
+    /// @returns {Array} Boolean array representation of the record plaintext bits.
+    #[wasm_bindgen(js_name = "toBitsLe")]
+    pub fn to_bits_le(&self) -> Array {
+        to_bits_array_le!(self)
+    }
+
```
iamalwaysuncomfortable, line 192: Same comment as above, `toFields` and `fromBytesLe` methods here as well.

wasm/src/account/signature.rs
```diff
@@ -76,6 +77,20 @@ impl Signature {
...
+        let plaintext = Plaintext::from(LiteralNative::Signature(Box::new(self.0)));
+        plaintext.to_bytes_le()
+    }
+
+    /// Get the left endian boolean array representation of the signature plaintext bits.
+    #[wasm_bindgen(js_name = "toPlaintextBitsLe")]
+    pub fn to_plaintext_bits_le(&self) -> Array {
+        let plaintext = Plaintext::from(LiteralNative::Signature(Box::new(self.0)));
+        plaintext.to_bits_le()
+    }
```
iamalwaysuncomfortable: Can you add `fromBytesLe`, `toBytesLe`, `fromBitsLe` `toBitsLe`, `toPlaintext`, `toFields`, `fromFields` on this  type as well?

wasm/src/account/address.rs
```diff
@@ -67,6 +71,20 @@ impl Address {
...
+        plaintext.to_bytes_le()
+    }
+
+    /// Get the left endian boolean array representation of the address plaintext bits.
+    #[wasm_bindgen(js_name = "toPlaintextBitsLe")]
+    pub fn to_plaintext_bits_le(&self) -> Array {
+        let plaintext = Plaintext::from(LiteralNative::Address(self.0));
+        plaintext.to_bits_le()
+    }
+
```
iamalwaysuncomfortable: Can you add `fromBytesLe`, `toBytesLe`, `fromBitsLe` `toBitsLe`, `toPlaintext`, `toFields`, `fromFields` on this?

wasm/src/account/address.rs
```diff
@@ -14,10 +14,14 @@
...
 
-use crate::account::{PrivateKey, Signature, ViewKey};
-
-use crate::{account::compute_key::ComputeKey, types::native::AddressNative};
+use crate::{
+    Plaintext,
+    account::{PrivateKey, Signature, ViewKey, compute_key::ComputeKey},
+    native::LiteralNative,
+    types::native::AddressNative,
+};
```
iamalwaysuncomfortable, line 28:
```suggestion
};

```

iamalwaysuncomfortable (changes requested):
Added a few comments.

In general for all the `Literal` types, we want to support the interface of:
`fromBytesLe`, `toBytesLe`, `fromBitsLe` `toBitsLe`, `toPlaintext`, `toFields`, `fromFields` for serialization so that they have a consistent representation back and forth.

Which means we'll have to put `toPlaintext` also on the `Field`, `Group` and `Scalar` types already defined.

We may want to consider instead of `PlaintextBitsLe` `toPlaintext().toBytesLe()` in order to be explicit about what's actually happening and ensure we update the documentation on the hash methods to ensure this story is told explicitly. But I'm willing to have a debate on which semantics are best.

wasm/src/programs/data/plaintext.rs
alexpitsikoulis: done

wasm/src/account/address.rs
```diff
@@ -67,6 +136,30 @@ impl Address {
...
+    #[wasm_bindgen(js_name = "toPlaintext")]
+    pub fn to_plaintext(&self) -> Plaintext {
+        Plaintext::from(LiteralNative::Address(self.0))
+    }
+
+    /// Get the left endian byte array representation of the address plaintext.
+    #[wasm_bindgen(js_name = "toPlaintextBytesLe")]
+    pub fn to_plaintext_bytes_le(&self) -> Result<Uint8Array, String> {
+        self.to_plaintext().to_bytes_le()
+    }
```
iamalwaysuncomfortable: Don't need bytes

wasm/src/programs/data/ciphertext.rs
```diff
@@ -19,13 +19,22 @@ use crate::{
...
+    from_js_typed_array,
+    from_wasm_object_array,
+    js_array_from_fields,
     native::{CiphertextNative, CurrentNetwork, FieldNative, FromBytes, IdentifierNative, ProgramIDNative, ToBytes},
+    to_bits_array_le,
 };
-use snarkvm_console::{network::Network, program::compute_function_id, types::U16};
 
-use js_sys::Uint8Array;
+use snarkvm_console::{
```
iamalwaysuncomfortable, line 28: Eliminate the space right above this.

wasm/src/account/address.rs
```diff
@@ -50,6 +62,63 @@ impl Address {
...
+        Ok(Self(native))
+    }
+
+    /// Get the field array representation of the address.
+    #[wasm_bindgen(js_name = "toFields")]
+    pub fn to_fields(&self) -> Result<Array, String> {
+        let native = self.0.clone();
+        let native_fields = native.to_fields().map_err(|e| e.to_string())?;
+        Ok(js_array_from_fields!(&native_fields))
+    }
```
iamalwaysuncomfortable, line 120: Can you put from/to on the group type as well?

wasm/src/account/address.rs
```diff
@@ -14,11 +14,23 @@
...
+    Plaintext,
+    account::{PrivateKey, Signature, ViewKey, compute_key::ComputeKey},
+    from_js_typed_array,
+    from_wasm_object_array,
+    js_array_from_fields,
+    native::{FieldNative, LiteralNative},
+    to_bits_array_le,
+    types::native::AddressNative,
+};
+
```
iamalwaysuncomfortable:
```suggestion
```

iamalwaysuncomfortable: No space between Crate and snarkvm imports

wasm/src/record/record_ciphertext.rs
```diff
@@ -14,8 +14,19 @@
...
+use crate::{
+    Field,
+    GraphKey,
+    RecordPlaintext,
+    ViewKey,
+    js_array_from_fields,
+    to_bits_array_le,
+    types::native::RecordCiphertextNative,
+};
+
```
iamalwaysuncomfortable:
```suggestion
```

wasm/src/record/record_plaintext.rs
```diff
@@ -33,10 +35,14 @@ use crate::{
         },
     },
 };
-use snarkvm_console::program::Owner;
+
```
iamalwaysuncomfortable:
```suggestion
```

wasm/src/algorithms/poseidon/mod.rs
```diff
@@ -56,31 +57,31 @@ mod tests {
...
-            let hash_8_group = poseidon8.hash_to_group(js_array_from_fields(&native_field_array)).unwrap();
+            let hash_2 = poseidon2.hash(js_array_from_fields!(&native_field_array)).unwrap();
+            let hash_4 = poseidon4.hash(js_array_from_fields!(&native_field_array)).unwrap();
+            let hash_8 = poseidon8.hash(js_array_from_fields!(&native_field_array)).unwrap();
+            let hash_2_scalar = poseidon2.hash_to_scalar(js_array_from_fields!(&native_field_array)).unwrap();
+            let hash_4_scalar = poseidon4.hash_to_scalar(js_array_from_fields!(&native_field_array)).unwrap();
+            let hash_8_scalar = poseidon8.hash_to_scalar(js_array_from_fields!(&native_field_array)).unwrap();
+            let hash_2_group = poseidon2.hash_to_group(js_array_from_fields!(&native_field_array)).unwrap();
+            let hash_4_group = poseidon4.hash_to_group(js_array_from_fields!(&native_field_array)).unwrap();
+            let hash_8_group = poseidon8.hash_to_group(js_array_from_fields!(&native_field_array)).unwrap();
```
iamalwaysuncomfortable, line 78: Could you add an extra `wasm_bindgen_test` 

That:
1. For all the literal types that go from literal.to_plaintext().to_fields(), hashes them and then checks against the native implementation?
2. For a struct type that includes all the plaintext literal types we currently support, does the same as above?

Could you to the same for `BHP` and `Pedersen`? (Of course going from `literal.to_plaintext().to_bits()` instead)

iamalwaysuncomfortable (commented):
Looking good, a couple nits + asks for tests in Rust. 

In addition, could you make some tests in JS side of the SDK that hash plaintexts for:

* All the literal types we current have available (`Group`, `Scalar`, `Field`, `Address`, `Signature`)
* Struct plaintexts containing those types
* An array plaintext

And ensure they match what Leo puts out? You can create the leo tests by just creating a small program that does each type of hash on the inputs above and outputs the hash value publicly, and uses that as test vectors for the JS tests data.

Alternatively, you could write something in CI that installs the Leo CLI, runs a JS program that gets hashes out of specific fields values, and matches them inside the program. This would ultimately be the more robust approach to maintain longterm integration testing that the hash functions of plaintext match leo representation.

wasm/src/types/group.rs
```diff
@@ -177,3 +179,28 @@ impl From<&Group> for GroupNative {
         group.0
     }
 }
+
+impl From<GroupNative> for Address {
+    fn from(value: GroupNative) -> Self {
+        let native = AddressNative::new(value);
+        Address::from(native)
+    }
+}
```
iamalwaysuncomfortable, line 198: Could you add to/from fields also for the Group type?

wasm/src/algorithms/poseidon/mod.rs
```diff
@@ -27,14 +27,17 @@ pub use poseidon8::Poseidon8;
...
-        Scalar,
-        types::native::{FieldNative, Poseidon2Native, Poseidon4Native, Poseidon8Native},
-        utilities::test::{create_native_field_vector, js_array_from_fields},
+        js_array_from_fields, native::{GroupNative, LiteralNative, PlaintextNative, ScalarNative}, test::TEST_STRUCT, types::native::{FieldNative, Poseidon2Native, Poseidon4Native, Poseidon8Native}, utilities::test::create_native_field_vector, Field, Group, Scalar
+    };
+    use snarkvm_console::{
+        algorithms::{Hash, HashMany, HashToGroup, HashToScalar},
+        prelude::ToFields,
+        program::{Literal, Plaintext},
+        types::{Address, Boolean},
```
iamalwaysuncomfortable: We don't need to test Boolean yet since we don't export it in the SDK, the address imported should be AddressNative as this type is the same type as that.

```suggestion
        types::{Address, Boolean},
```

wasm/src/algorithms/poseidon/mod.rs
```diff
@@ -27,14 +27,17 @@ pub use poseidon8::Poseidon8;
...
-        Group,
-        Scalar,
-        types::native::{FieldNative, Poseidon2Native, Poseidon4Native, Poseidon8Native},
-        utilities::test::{create_native_field_vector, js_array_from_fields},
+        js_array_from_fields, native::{GroupNative, LiteralNative, PlaintextNative, ScalarNative}, test::TEST_STRUCT, types::native::{FieldNative, Poseidon2Native, Poseidon4Native, Poseidon8Native}, utilities::test::create_native_field_vector, Field, Group, Scalar
+    };
+    use snarkvm_console::{
+        algorithms::{Hash, HashMany, HashToGroup, HashToScalar},
+        prelude::ToFields,
+        program::{Literal, Plaintext},
```
iamalwaysuncomfortable: These are the same types as LiteralNative and PlaintextNative respectively. So they shouldn't be directly imported. Any imports of a type we export like `Plaintext` should import the `wasm` struct we define for it.

```suggestion
```

wasm/src/algorithms/poseidon/mod.rs
```diff
@@ -27,14 +27,17 @@ pub use poseidon8::Poseidon8;
 mod tests {
     use super::*;
     use crate::{
-        Field,
-        Group,
-        Scalar,
-        types::native::{FieldNative, Poseidon2Native, Poseidon4Native, Poseidon8Native},
-        utilities::test::{create_native_field_vector, js_array_from_fields},
+        js_array_from_fields, native::{GroupNative, LiteralNative, PlaintextNative, ScalarNative}, test::TEST_STRUCT, types::native::{FieldNative, Poseidon2Native, Poseidon4Native, Poseidon8Native}, utilities::test::create_native_field_vector, Field, Group, Scalar
```
iamalwaysuncomfortable:
```suggestion
        js_array_from_fields, native::{GroupNative, LiteralNative, PlaintextNative, ScalarNative}, test::TEST_STRUCT, types::native::{FieldNative, Poseidon2Native, Poseidon4Native, Poseidon8Native}, utilities::test::create_native_field_vector, Address, Field, Group, Scalar
```

wasm/src/algorithms/poseidon/mod.rs
```diff
@@ -116,4 +119,82 @@ mod tests {
...
+
+        // Create native instances of hasher structs.
+        let native_poseidon2 = Poseidon2Native::setup("AleoPoseidon2").unwrap();
+        let native_poseidon4 = Poseidon4Native::setup("AleoPoseidon4").unwrap();
+        let native_poseidon8 = Poseidon8Native::setup("AleoPoseidon8").unwrap();
+
+        let address = Address::zero();
+
+        let literals = vec![
+            Literal::Address(address),
```
iamalwaysuncomfortable:
```suggestion
            address,
```

All of these should create the `wasm` representation of these objects, and use `to_plaintext` method on them.

wasm/src/algorithms/poseidon/mod.rs
```diff
@@ -116,4 +119,82 @@ mod tests {
...
+            Literal::Boolean(Boolean::new(false)),
```
iamalwaysuncomfortable:
```suggestion
```

We don't export Boolean, so no need to test it. However, we do export Signature, so we should test that.

wasm/src/algorithms/poseidon/mod.rs
```diff
@@ -116,4 +119,82 @@ mod tests {
...
+                GroupNative::from_str(
+                    "7243206743250892049702172909169115544952822465955921992746259936160368017976group",
+                )
+                .unwrap(),
+            ),
+            LiteralNative::Scalar(ScalarNative::from_field_lossy(&snarkvm_console::types::Field::from_u8(2))),
+        ];
+
+        for i in 0..literals.len() {
+            let literal_fields = Plaintext::Literal(literals[i].clone(), OnceCell::new()).to_fields().unwrap();
```
iamalwaysuncomfortable: Here you're using the `Plaintext` exported from SnarkVM, so it's the exact same type as `PlaintextNative` and thus this test doesn't test anything just yet about the was representation. This should just use the `to_fields` method on the exported plaintext type and you won't need to do the explicit conversion here with the macro.

wasm/src/algorithms/bhp/mod.rs
```diff
@@ -125,4 +123,168 @@ mod tests {
...
+                GroupNative::from_str(
+                    "7243206743250892049702172909169115544952822465955921992746259936160368017976group",
+                )
+                .unwrap(),
+            ),
+            LiteralNative::Scalar(ScalarNative::from_field_lossy(&snarkvm_console::types::Field::from_u8(2))),
+        ];
+
+        for i in 0..literals.len() {
+            let literal_bits = Plaintext::Literal(literals[i].clone(), OnceCell::new()).to_bits_le();
```
iamalwaysuncomfortable: Same comment here as in the Poseidon hash comments, this is just the Plaintext type directly from SnarkVM. We want to test the wasm representation, so the definition above of the `literal` variable should create our exported `Plaintext` structs instead with the `to_plaintext` method.

wasm/src/algorithms/bhp/mod.rs
```diff
@@ -125,4 +123,168 @@ mod tests {
...
+            assert_eq!(commit_512, Field::from(native_commit_512));
+            assert_eq!(commit_768, Field::from(native_commit_768));
+            assert_eq!(commit_1024, Field::from(native_commit_1024));
+            assert_eq!(commit_to_group_256, Group::from(native_commit_to_group_256));
+            assert_eq!(commit_to_group_512, Group::from(native_commit_to_group_512));
+            assert_eq!(commit_to_group_768, Group::from(native_commit_to_group_768));
+            assert_eq!(commit_to_group_1024, Group::from(native_commit_to_group_1024));
+        }
+
+        let struct_pt: Plaintext<CurrentNetwork> = Plaintext::from_str(TEST_STRUCT).unwrap();
```
iamalwaysuncomfortable: Same here, this is the same type as `PlaintextNative`, use the `wasm` struct definition we've defined in the SDK instead.

iamalwaysuncomfortable (changes requested):
The tests are about the right structure, but they aren't testing the `Plaintext` representation that's been defined in the SDK. Once that change is made, these tests will be accurate.

wasm/src/algorithms/bhp/mod.rs
```diff
@@ -125,4 +130,166 @@ mod tests {
...
+
+        let literals = vec![
+            address.to_plaintext(),
+            Field::from_string("2field").unwrap().to_plaintext(),
+            Group::from_string("7243206743250892049702172909169115544952822465955921992746259936160368017976group")
+                .unwrap()
+                .to_plaintext(),
+            Scalar::from_string("836504693989570607341914239820012911582004515616146791081874852343183183566scalar")
+                .unwrap()
+                .to_plaintext(),
```
iamalwaysuncomfortable, line 162: Add the signature type here as well.

wasm/src/algorithms/poseidon/mod.rs
```diff
@@ -116,4 +126,81 @@ mod tests {
...
+
+        let literals = vec![
+            address.to_plaintext(),
+            Field::from_string("2field").unwrap().to_plaintext(),
+            Group::from_string("7243206743250892049702172909169115544952822465955921992746259936160368017976group")
+                .unwrap()
+                .to_plaintext(),
+            Scalar::from_string("836504693989570607341914239820012911582004515616146791081874852343183183566scalar")
+                .unwrap()
+                .to_plaintext(),
```
iamalwaysuncomfortable, line 153: Add the signature type here

iamalwaysuncomfortable (commented):
A couple more nits

iamalwaysuncomfortable (approved):

jaketarnow (approved):
LGTM

PR #1011, opened 2025-03-27
Create Leo App - specify template via command line

christianwwwwwwww:
This PR fixes the ability to add desired template in when runnning create leo app. 

Eg `npm create leo-app@latest my-leo-app --template vanilla`

### DIFF
--- a/create-leo-app/src/index.ts
+++ b/create-leo-app/src/index.ts
@@ -92,6 +92,20 @@ async function init() {
   const getProjectName = () =>
     targetDir === "." ? path.basename(path.resolve()) : targetDir;
 
+  let selectedFramework: Framework | undefined;
+  let selectedVariant: string | undefined;
+
+  if (argTemplate && TEMPLATES.includes(argTemplate)) {
+    for (const framework of FRAMEWORKS) {
+      const match = framework.variants.find((v) => v.name === argTemplate);
+      if (match) {
+        selectedFramework = framework;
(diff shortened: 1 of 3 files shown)


create-leo-app/src/index.ts
```diff
@@ -92,6 +92,20 @@ async function init() {
...
+    for (const framework of FRAMEWORKS) {
+      const match = framework.variants.find((v) => v.name === argTemplate);
+      if (match) {
+        selectedFramework = framework;
+        selectedVariant = match.name;
+        break;
+      }
+    }
+  }
+
```
iamalwaysuncomfortable, line 108: This should likely error if there's no matching template with a helpful error message that specifies which templates are available.

iamalwaysuncomfortable (changes requested):
Two minor nits: 

The templates should error if they're not found as generally the point of using the cli command in this way is to avoid an interactive cli.

Could you also write some testing of this functionality in circleci to ensure it builds properly?  

Otherwise this is great. We need more small, digestible changes like this.

create-leo-app/src/index.ts
christianwwwwwwww: It currently does this! 

If you try for example `--template nonsense` it will give `"nonsense" isn't a valid template. Please choose from below:` and lists the available templates

iamalwaysuncomfortable (approved):

PR #1022, opened 2025-04-11
wait for tx confirmation - client side await

### LINKED ISSUE #779: [Feature] Client Side Await for Aleo Futures
## Motivation
[Futures have been introduced into Aleo Instructions](https://github.com/AleoHQ/snarkVM/pull/2009). However on the client side, currently the client would have to implement their own logic in order to check if a future leading to a finalize scope was successful. Optimally, if an Aleo instructions function leads returns a future type it should return a promise that resolves when the future has completed executing on chain.

This will allow users to allow their apps to react directly to on-chain state

## Implementation 
- [ ] If a function return type is of type `future`, create javascript promise that polls the for a finalize transaction matching the transaction id on chain. The future resolves when it detects a succcess, rejection, or a timeout.
- [ ] Add the inputs the finalize block was called on to the promise. If the inputs were also futures, add those to the data in the promise
- [ ] Add a typescript interface in order to allow SDK users to extend this promise type to enable custom behavior (i.e. like checking the resulting mapping, or triggering another action, etc.) so that users have an interface to reacting to updates in on-chain state

christianwwwwwwww:
This PR expands upon the `waitForTransactionConfirmation` function within `AleoNetworkClient` and adds unit tests.

### DIFF
--- a/sdk/src/network-client.ts
+++ b/sdk/src/network-client.ts
@@ -1214,12 +1214,18 @@ class AleoNetworkClient {
         }
   
         try {
-          const res = await fetch(`https://api.explorer.provable.com/v1/mainnet/transaction/confirmed/${transactionId}`);
-        
+          const res = await fetch(`${this.host}/transaction/confirmed/${transactionId}`, {
+            headers: this.headers
+          });
           if (!res.ok) {
-            const text = await res.text();
+            let text = "";
+            try {
+              text = await res.text();
+              console.warn("Response text from server:", text);
+            } catch (err) {
+              console.warn("Failed to read response text:", err);
+            }
           
-            // Break early if the response indicates an invalid format
             if (res.status >= 400 && res.status < 500 && text.includes("Invalid URL")) {
               clearInterval(interval);
               return reject(new Error(`Malformed transaction ID: ${text}`));
--- a/sdk/src/network-client.ts
+++ b/sdk/src/network-client.ts
@@ -1237,8 +1237,7 @@ class AleoNetworkClient {
             return;
           }
           
-          const text = await res.text();
-          const data = parseJSON(text);          
+          const data = parseJSON(await res.text());          
           if (data?.status === "accepted") {
             clearInterval(interval);
             return resolve(data);
@@ -1249,7 +1248,6 @@ class AleoNetworkClient {
             return reject(new Error(
               `Transaction ${transactionId} was rejected by the network. Ensure that the account paying the fee has enough credits and that the inputs to the on-chain function are valid.`));
             }
-          // No status? keep polling
         } catch (err) {
           console.error("Polling error:", err);
         }
--- a/sdk/src/network-client.ts
+++ b/sdk/src/network-client.ts
@@ -1218,9 +1218,16 @@ class AleoNetworkClient {
         
           if (!res.ok) {
             const text = await res.text();
-            console.error("Non-OK response:", res.status, text);
-            clearInterval(interval);
-            return reject(new Error(`HTTP ${res.status}: ${text}`));
+          
+            // Break early if the response indicates an invalid format
+            if (res.status >= 400 && res.status < 500 && text.includes("Invalid URL")) {
+              clearInterval(interval);
+              return reject(new Error(`Malformed transaction ID: ${text}`));
+            }
+          
+            // Log and continue polling for 404s or 5XX errors in case a tx doesn't exist yet
+            console.warn("Non-OK response (retrying):", res.status, text);
+            return;
           }
           
           const data = await res.json();
--- a/sdk/src/network-client.ts
+++ b/sdk/src/network-client.ts
@@ -1210,8 +1210,7 @@ class AleoNetworkClient {
   
         if (elapsed > timeout) {
           clearInterval(interval);
-          return reject(new Error("Transaction confirmation timed out"));
-        }
+          return reject(new Error(`Transaction ${transactionId} did not appear after the timeout period of ${interval}ms - consider resubmitting the transaction`));        }
   
         try {
           const res = await fetch(`${this.host}/transaction/confirmed/${transactionId}`, {
@@ -1226,6 +1225,8 @@ class AleoNetworkClient {
               console.warn("Failed to read response text:", err);
             }
           
+            // If the transaction ID is malformed (e.g. invalid checksum, wrong length), 
+            // the API returns a 4XX with "Invalid URL" — we treat this as a fatal error and stop polling.
             if (res.status >= 400 && res.status < 500 && text.includes("Invalid URL")) {
               clearInterval(interval);
               return reject(new Error(`Malformed transaction ID: ${text}`));
@@ -1245,10 +1246,10 @@ class AleoNetworkClient {
         
           if (data?.status === "rejected") {
             clearInterval(interval);
-            return reject(new Error(`Transaction ${transactionId} was rejected by the network`));
-          }
-        
-          // no status? keep polling
+            return reject(new Error(
+              `Transaction ${transactionId} was rejected by the network. Ensure that the account paying the fee has enough credits and that the inputs to the on-chain function are valid.`));
+            }
+          // No status? keep polling
         } catch (err) {
           console.error("Polling error:", err);
         }
--- a/sdk/src/network-client.ts
+++ b/sdk/src/network-client.ts
@@ -3,20 +3,20 @@ import { Account } from "./account";
 import { BlockJSON } from "./models/blockJSON";
 import { TransactionJSON } from "./models/transaction/transactionJSON";
 import {
-  Address,
-  Plaintext,
-  RecordCiphertext,
-  Program,
-  RecordPlaintext,
-  PrivateKey,
-  Transaction,
+    Address,
+    Plaintext,
+    RecordCiphertext,
+    Program,
+    RecordPlaintext,
+    PrivateKey,
+    Transaction,
 } from "./wasm";
 import { ConfirmedTransactionJSON } from "./models/confirmed_transaction";
 
 type ProgramImports = { [key: string]: string | Program };
 
 interface AleoNetworkClientOptions {
-  headers?: { [key: string]: string };
+    headers?: { [key: string]: string };
 }
 
 /**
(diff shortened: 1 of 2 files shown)


sdk/src/network-client.ts
```diff
@@ -1198,24 +1198,46 @@ class AleoNetworkClient {
...
-          resolve(transaction);
-          if (Date.now() - startTime > timeout) {
+          const res = await fetch(`https://api.explorer.provable.com/v1/mainnet/transaction/confirmed/${transactionId}`);
+        
+          if (!res.ok) {
+            const text = await res.text();
+            console.error("Non-OK response:", res.status, text);
+            clearInterval(interval);
+            return reject(new Error(`HTTP ${res.status}: ${text}`));
+          }
```
iamalwaysuncomfortable: In the following cases
1. If the transaction doesn't exist yet
2. If the API itself has an internal error (An an error 5XX)

Won't it given a non-okay response and cancel this awaitable immediately? If a tx doesn't exist yet, this should just keep polling until the time runs out except in the case that it detects an invalid URL.

If the transaction is simply missing from the ledger, it'll return a 4XX error:
`Something went wrong: Missing transaction for ID at15wupghsjdgv0rhxlk7rfx8umm3893xrsskhfnfndt7mtcaxw659s7rqzlt`

So in that case and in the case of 5XX errors, it should keep polling.

However, if the transaction ID is incorrectly formatted it'll return a 4XX error: 
`Invalid URL: Invalid byte size for a bech32m hash: 6 bytes`
`Invalid URL: invalid checksum`
`Invalid URL: invalid character (code=b)`

The way to detect that is to look for a 4XX error with `Invalid URL` in the response. When that happens this method can cancel the timeout.

sdk/tests/network-client.test.ts
```diff
@@ -198,6 +198,27 @@ describe('NodeConnection', () => {
...
+        it('should return rejected for a rejected tx ID', async () => {
+          const connection = new AleoNetworkClient("https://api.explorer.provable.com/");
+          const status = await connection.waitForTransactionConfirmation("at1x5ed0pyjpt3770e0m60psvdvqeww5z5f32t7velrmrjfhvzgysxqll4g6a");
+          expect(status).to.equal("rejected");
+        });
+      
+        it('should throw for an invalid tx ID', async () => {
+          const connection = new AleoNetworkClient("https://api.explorer.provable.com/");
+          await expectThrows(() =>
+            connection.waitForTransactionConfirmation("at1dl9lze8wscct0dee8x9tjnfmpj12345678jcnp5f0ywjn5552yrsperzl9")
```
iamalwaysuncomfortable: This is a good test, but we should ensure, that the main method detects a malformed transaction ID.

In the case of a malformed URL, you'll get responses like:
`Invalid URL: Invalid byte size for a bech32m hash: 6 bytes`
`Invalid URL: invalid checksum`
`Invalid URL: invalid character (code=b)`

So the main method should detect an invalid URL by the presence of `Invalid URL` in the response. And communicate that back to the user in the error.

sdk/src/network-client.ts
christianwwwwwwww: Good callouts thanks!

sdk/src/network-client.ts
```diff
@@ -1198,24 +1198,53 @@ class AleoNetworkClient {
...
+          clearInterval(interval);
+          return reject(new Error("Transaction confirmation timed out"));
+        }
+  
         try {
-          // Replace with actual Aleo transaction lookup API
-          const transaction = <Transaction>await this.getTransactionObject(transactionId);
-          resolve(transaction);
-          if (Date.now() - startTime > timeout) {
+          const res = await fetch(`https://api.explorer.provable.com/v1/mainnet/transaction/confirmed/${transactionId}`);
```
iamalwaysuncomfortable: We should be using the configured url.

```suggestion
const res = await get(this.host + `/transaction/confirmed/${transactionId}`, {
        headers: this.headers
      });
```

sdk/src/network-client.ts
```diff
@@ -1198,24 +1198,53 @@ class AleoNetworkClient {
...
    */
   async waitForTransactionConfirmation(
-      transactionId: string,
-      checkInterval: number = 2000,  // Poll every 2 seconds
-      timeout: number = 45000        // Timeout after 45 seconds
-  ): Promise<Transaction> {
+    transactionId: string,
+    checkInterval: number = 2000,
+    timeout: number = 45000
+  ): Promise<"accepted" | "rejected"> {
```
iamalwaysuncomfortable:
```suggestion
  ): Promise<ConfirmedTransactionJSON> {
```

sdk/src/network-client.ts
```diff
@@ -1198,24 +1198,53 @@ class AleoNetworkClient {
...
+              clearInterval(interval);
+              return reject(new Error(`Malformed transaction ID: ${text}`));
+            }
+          
+            // Log and continue polling for 404s or 5XX errors in case a tx doesn't exist yet
+            console.warn("Non-OK response (retrying):", res.status, text);
+            return;
+          }
+          
+          const data = await res.json();
```
iamalwaysuncomfortable:
```suggestion
          const data = parseJSON(res.text())
```

sdk/src/network-client.ts
```diff
@@ -1198,24 +1198,53 @@ class AleoNetworkClient {
...
+        
+          if (data?.status === "accepted") {
+            clearInterval(interval);
+            return resolve("accepted");
```
iamalwaysuncomfortable:
```suggestion
            return resolve(data);
```

sdk/src/network-client.ts
```diff
@@ -1198,24 +1198,53 @@ class AleoNetworkClient {
...
+          }
+        
+          if (data?.status === "rejected") {
             clearInterval(interval);
-            reject(new Error("Transaction confirmation timed out"));
+            return resolve("rejected");
```
iamalwaysuncomfortable: Here it should have an informative error such as this I think.

```suggestion
            reject(return reject(new Error(`Transaction ${transactoinID} was rejected by the network`));
```

sdk/src/network-client.ts
christianwwwwwwww: noted - I've updated to use the configured URL. However, I am still using `fetch` rather than `get` here, as our `get` method throws immediately on non-OK responses, which prevents us from inspecting the response body. Let me know if you think this is an issue.

sdk/src/network-client.ts
```diff
@@ -1198,24 +1198,59 @@ class AleoNetworkClient {
...
-
-    return new Promise<Transaction>((resolve, reject) => {
+  
+    return new Promise((resolve, reject) => {
       const interval = setInterval(async () => {
+        const elapsed = Date.now() - startTime;
+  
+        if (elapsed > timeout) {
+          clearInterval(interval);
+          return reject(new Error("Transaction confirmation timed out"));
```
iamalwaysuncomfortable: In general in this method we want the errors to provide helpful information about what action to take in response to the error.

```suggestion
          return reject(new Error(`Transaction ${transactionId} did not appear after the timeout period of ${interval}ms - consider resubmitting the transaction`));
```

sdk/src/network-client.ts
```diff
@@ -1198,24 +1198,59 @@ class AleoNetworkClient {
...
+          if (!res.ok) {
+            let text = "";
+            try {
+              text = await res.text();
+              console.warn("Response text from server:", text);
+            } catch (err) {
+              console.warn("Failed to read response text:", err);
+            }
+          
+            if (res.status >= 400 && res.status < 500 && text.includes("Invalid URL")) {
```
iamalwaysuncomfortable: We should add a comment above this line as to why this check is being made.

sdk/src/network-client.ts
```diff
@@ -1198,24 +1198,59 @@ class AleoNetworkClient {
...
+              return reject(new Error(`Malformed transaction ID: ${text}`));
+            }
+          
+            // Log and continue polling for 404s or 5XX errors in case a tx doesn't exist yet
+            console.warn("Non-OK response (retrying):", res.status, text);
+            return;
+          }
+          
+          const text = await res.text();
+          const data = parseJSON(text);
```
iamalwaysuncomfortable:
```suggestion
          const data = parseJSON(res.text());          
```

sdk/src/network-client.ts
```diff
@@ -1198,24 +1198,59 @@ class AleoNetworkClient {
...
+          const data = parseJSON(text);          
+          if (data?.status === "accepted") {
+            clearInterval(interval);
+            return resolve(data);
+          }
+        
+          if (data?.status === "rejected") {
             clearInterval(interval);
-            reject(new Error("Transaction confirmation timed out"));
+            return reject(new Error(`Transaction ${transactionId} was rejected by the network`));
```
iamalwaysuncomfortable: Also here we probably want to add more context, the below is just a suggestion, but in general we should explain the things the developer should check for to resolve the failure.

```suggestion
            return reject(new Error(`Transaction ${transactionId} was rejected by the network - ensure the account paying the fee has enough credits and that inputs to the on-chain part of the function are correct`));
```

sdk/src/network-client.ts
```diff
@@ -1198,24 +1198,59 @@ class AleoNetworkClient {
...
           }
-        } catch (error) {
-          console.error("Error checking transaction:", error);
+        
+          // no status? keep polling
```
iamalwaysuncomfortable:
```suggestion
         // No status? keep polling
```

sdk/tests/network-client.test.ts
```diff
@@ -198,6 +198,53 @@ describe('NodeConnection', () => {
...
+        const invalidTx = "at1dl9lze8wscct0dee8x9tjnfmpj12345678jcnp5f0ywjn5552yrsperzl9";
+      
+        const host = "https://api.explorer.provable.com/v1";
+      
+        function getTxId(connection: AleoNetworkClient, type: "accepted" | "rejected") {
+          const isTestnet = connection.host.includes("/testnet");
+          return type === "accepted"
+            ? (isTestnet ? testnetAcceptedTx : mainnetAcceptedTx)
+            : (isTestnet ? testnetRejectedTx : mainnetRejectedTx);
+        }
```
iamalwaysuncomfortable: Love this.

iamalwaysuncomfortable (commented):
Looks much better, a few more nits, but this should be good to go soon.

sdk/src/network-client.ts
christianwwwwwwww: Because `res.text()` is async we need to `await` it before we can parse it. Otherwise TypeScript cries

sdk/tests/network-client.test.ts
christianwwwwwwww: yay

sdk/src/network-client.ts
```diff
@@ -1198,24 +1198,60 @@ class AleoNetworkClient {
...
           }
-        } catch (error) {
-          console.error("Error checking transaction:", error);
+        
+          if (data?.status === "rejected") {
+            clearInterval(interval);
+            return reject(new Error(
+              `Transaction ${transactionId} was rejected by the network. Ensure that the account paying the fee has enough credits and that the inputs to the on-chain function are valid.`));
+            }
+          // No status? keep polling
```
iamalwaysuncomfortable: It's not standard to have comments at the end of code blocks, so I think we can just eliminate this one.

```suggestion
```

sdk/src/network-client.ts
iamalwaysuncomfortable: You're right but we should await directly in the block like so:

```javascript
const data = parseJSON(await res.text());  
```

iamalwaysuncomfortable (commented):
Just a few more nits.

sdk/src/network-client.ts
christianwwwwwwww: agree

iamalwaysuncomfortable (approved):
This is approved. However, before it's merged, can you run the `prettier` linter on it?

PR #1025, opened 2025-04-16
Implement network retry system

### LINKED ISSUE #781: [Bug] Implement network retry system
## 🐛 Bug Report

Various functions in the SDK suffer from spurious network errors (e.g. from network connectivity issues or from changes not reflecting in the API/chain fast enough).

## Steps to Reproduce

https://github.com/AleoHQ/sdk/pull/774#issuecomment-1764945987

## Expected Behavior

We should implement a retry system with proper backoff or expose helper methods for letting users implement their own.

### LINKED ISSUE #829: [Bug] `findUnspentRecords` should not use request failure to check unspent record
## 🐛 Bug Report
`findUnspentRecords` currently calls the `getTransactionId` API and, upon failure, assumes that the record is unspent. Using an unspecified catch block to handle core logic is an anti-pattern. The API errors can happen for unknown reasons, which would lead to false positives.

christianwwwwwwww:
I've written a `retryWithBackoff` helper function in `.utils` that can be used throughout the network client and program manager, for example in `fetchData`, `submitTransaction` etc. 

Resolves #781


sdk/src/utils.ts
```diff
@@ -39,3 +39,36 @@ export async function post(url: URL | string, options: RequestInit) {
...
+        const isLast = attempt === maxAttempts;
+  
+        // retry only for certain errors
+        const retryable =
+        //   err?. 404s but that will probably not be 404 soon ? 
+          error?.message?.includes('404') ||
+        //   retry for timeouts as well
+          error?.message?.includes('5');
+  
+        if (!retryable || isLast) throw error;
```
iamalwaysuncomfortable: These should probably be clarified be separated

if (!retryable) throw error;
if (isLast) throw new Error(`Failed after ${maxAttempts} attempts: ${error.message}`);

sdk/src/utils.ts
```diff
@@ -39,3 +39,36 @@ export async function post(url: URL | string, options: RequestInit) {
...
+      } catch (err: any) {
+        const error = err as Error & { code?: string };
+        const isLast = attempt === maxAttempts;
+  
+        // retry only for certain errors
+        const retryable =
+        //   err?. 404s but that will probably not be 404 soon ? 
+          error?.message?.includes('404') ||
+        //   retry for timeouts as well
+          error?.message?.includes('5');
```
iamalwaysuncomfortable: This is probably too general given the variety of API responses that can come back. There are some cases like querying mappings where you'd be explicitly testing for a 404 to see if a key exists or not. We probably want to make this configurable.

sdk/src/utils.ts
```diff
@@ -39,3 +39,36 @@ export async function post(url: URL | string, options: RequestInit) {
 
     return response;
 }
+
+export async function retryWithBackoff<T>(
+    fn: () => Promise<T>,
+    maxAttempts = 5,
+    baseDelay = 100
+  ): Promise<T> {
```
iamalwaysuncomfortable: To make it a little more robust, we probably want to add selective statuses to retry on for various functions + some logic to determine conditionally when to retry. Like for instance, if your transaction id is malformed, we don't want to retry, but if it doesn't exist yet, you may want to.

```
type RetryOptions = {
  maxAttempts?: number;
  baseDelay?: number;
  retryOnStatus?: number[];
  shouldRetry?: (err: any) => boolean;
};

export async function retryWithBackoff<T>(
  fn: () => Promise<T>,
  {
    maxAttempts = 5,
    baseDelay = 100,
    retryOnStatus = [],
    shouldRetry,
  }: RetryOptions = {}
): Promise<T> {
```

This might be called like so depending on the method.
```
await retryWithBackoff(() => fetchWithTimeout('/api/data'), {
  retryOnStatus: [404],
  shouldRetry: (e) => e.message?.includes("timeout"),
});
```

sdk/src/network-client.ts
```diff
@@ -101,7 +101,7 @@ class AleoNetworkClient {
      */
     async fetchData<Type>(url = "/"): Promise<Type> {
         try {
-            return parseJSON(await this.fetchRaw(url));
+            return retryWithBackoff(() => this.fetchRaw(url).then(parseJSON));
```
iamalwaysuncomfortable: For the GET, we probably don't want to wrap this directly as whether to retry depends on a lot of different factors for different methods. We should instead wrap the individual method gets themselves with error codes and conditions that are contextual to that.

iamalwaysuncomfortable (commented):
Added a few comments to make the error handling more robust.

sdk/src/network-client.ts
```diff
@@ -153,15 +160,35 @@ class AleoNetworkClient {
...
+          return await retryWithBackoff(async () => {
+              const response = await get(this.host + url, {
+                  headers: this.headers,
+              });
+              return await response.text();
+          }, {
+              retryOnStatus: [500, 502, 503, 504],
+              shouldRetry: (err) => {
+                  const msg = err?.message?.toLowerCase?.() || "";
+                  return msg.includes("network") || msg.includes("timeout") || msg.includes("503");
```
iamalwaysuncomfortable: One questoin is if we're missing any edge cases. Shouldn't we basically retry on most 500 cases?

sdk/src/network-client.ts
```diff
@@ -137,12 +137,19 @@ class AleoNetworkClient {
...
-        }
-    }
+      try {
+          return await retryWithBackoff(() => this.fetchRaw(url).then(parseJSON), {
+              retryOnStatus: [500, 502, 503, 504],
+              shouldRetry: (err) => {
+                  const msg = err?.message?.toLowerCase?.() || "";
+                  return msg.includes("network") || msg.includes("timeout") || msg.includes("503");
+              }
+          });
```
iamalwaysuncomfortable: Since this already handled by fetchRaw, it's probably redundant to have it here.

sdk/src/network-client.ts
```diff
@@ -1413,29 +1443,38 @@ class AleoNetworkClient {
...
+                  }),
+              }),
+              {
+                // Only retry on network-level / transient errors to avoid duplicate submission
+                  retryOnStatus: [500, 502, 503, 504],
+                  shouldRetry: (err) => {
+                      const msg = err?.message?.toLowerCase?.() || "";
+                      return msg.includes("timeout") || msg.includes("network") || msg.includes("503");
+                  },
+              }
```
iamalwaysuncomfortable: We can just retry in all cases with solution submission.

iamalwaysuncomfortable (commented):
Added a couple of nits + some questions about retrying on 500.

iamalwaysuncomfortable (commented):
Also we should ultimately do optional 400 based retries on specific `Transition`, `Transaction`, and `getMapping` functions but this should be done in a follow-on PR.

sdk/src/network-client.ts
christianwwwwwwww: Good callout yeah, I updated the helper method to make this more robust. It now checks `error.status` directly and lets `shouldRetry` handle everything else, so we don’t rely on string matching like "5" in the message anymore.

We still default to retrying on [500, 502, 503, 504] in most callers, but we can expand that to all 5XXs if we want to be more aggressive.

sdk/src/network-client.ts
christianwwwwwwww: It does seem redundant, but in this case I kept the retry wrapper here intentionally to catch cases where `parseJSON fails`, since that’s not inside `fetchRaw`.

sdk/src/network-client.ts
christianwwwwwwww: Makes sense, I'll update this as well

sdk/src/network-client.ts
iamalwaysuncomfortable: `parseJSON` isn't a network function capable of producing 500s however, so this would end up just retrying n^2 amount of times when a 500 gets piped upwards which will end up in people waiting for unreasonably long when there truly is a network error (i.e. about 6s)

sdk/src/network-client.ts
```diff
@@ -364,25 +391,27 @@ class AleoNetworkClient {
...
+                                                              // Otherwise record the nonce that has been found
+                                                              const serialNumber = recordPlaintext.serialNumberString(
+                                                                  resolvedPrivateKey,
+                                                                  "credits.aleo",
+                                                                  "credits",
+                                                              );
+                                                              // Attempt to see if the serial number is spent
+                                                              try {
+                                                                  await retryWithBackoff(() => this.getTransitionId(serialNumber), {
+                                                                      retryOnStatus: [500, 502, 503],
```
iamalwaysuncomfortable: This should cover 504 as well.

sdk/src/network-client.ts
```diff
@@ -137,12 +137,19 @@ class AleoNetworkClient {
...
-        } catch (error) {
-            throw new Error(`Error fetching data: ${error}`);
-        }
-    }
+      try {
+          return await retryWithBackoff(() => this.fetchRaw(url).then(parseJSON), {
+              retryOnStatus: [500, 502, 503, 504],
+              shouldRetry: (err) => {
+                  const msg = err?.message?.toLowerCase?.() || "";
+                  return msg.includes("network") || msg.includes("timeout") || msg.includes("503");
```
iamalwaysuncomfortable:
```suggestion
                  return msg.includes("network") || msg.includes("timeout") || msg.includes("503");
```

Why these 3 things in particular?

iamalwaysuncomfortable (commented):
Added a few more comments

sdk/src/network-client.ts
christianwwwwwwww: good catch thanks

sdk/src/network-client.ts
christianwwwwwwww: My intention here was to catch common transient network errors, but I'm definitely happy to refine if you have other error responses in mind!

christianwwwwwwww: I updated the helper to always retry on 5XXs and only apply `shouldRetry` for 4XXs, so `fetchData` now just uses the default behavior

sdk/src/network-client.ts
```diff
@@ -137,11 +137,12 @@ class AleoNetworkClient {
      * @param url The URL to fetch data from.
      */
     async fetchData<Type>(url = "/"): Promise<Type> {
-        try {
-            return parseJSON(await this.fetchRaw(url));
-        } catch (error) {
-            throw new Error(`Error fetching data: ${error}`);
-        }
+      try {
+        const raw = await retryWithBackoff(() => this.fetchRaw(url));
```
iamalwaysuncomfortable: The fetch here shouldn't have any retry since the retry is already taken care of in fetchRaw

```suggestion
        const raw = this.fetchRaw(url);
```

sdk/src/network-client.ts
```diff
@@ -364,25 +384,27 @@ class AleoNetworkClient {
...
+                                                                  "credits.aleo",
+                                                                  "credits",
+                                                              );
+                                                              // Attempt to see if the serial number is spent
+                                                              try {
+                                                                  await retryWithBackoff(() => this.getTransitionId(serialNumber), {
+                                                                      retryOnStatus: [500, 502, 503, 504],
+                                                                      shouldRetry: (err) => {
+                                                                          const msg = err?.message?.toLowerCase?.() || "";
+                                                                          return msg.includes("timeout") || msg.includes("503") || msg.includes("network");
```
iamalwaysuncomfortable: Same thing here, we don't need the 500 codes or conditionals here since it now always retries on 500 and we actually WANT this to fail on 400

sdk/src/network-client.ts
```diff
@@ -153,15 +154,34 @@ class AleoNetworkClient {
...
             });
             return await response.text();
-        } catch (error) {
-            throw new Error(`Error fetching data: ${error}`);
-        }
-    }
+        }, {
+            shouldRetry: (err) => {
+                const msg = err?.message?.toLowerCase?.() || "";
+                return msg.includes("network") || msg.includes("timeout");
```
iamalwaysuncomfortable: I don't actually think we need these conditions directly on `GET`s

iamalwaysuncomfortable (commented):
Couple more nits

sdk/src/network-client.ts
```diff
@@ -153,15 +154,34 @@ class AleoNetworkClient {
...
+
+
+    /**
+     * Wrapper around the POST helper to allow mocking in tests. Not meant for use in production.
+     *
+     * @param url The URL to POST to.
+     * @param options The RequestInit options for the POST request.
+     * @returns The Response object from the POST request.
+     */
+    private async sendPost(url: string, options: RequestInit) {
```
Pauan: If it's private it should probably be called `_sendPost` instead.

sdk/src/utils.ts
```diff
@@ -39,3 +39,52 @@ export async function post(url: URL | string, options: RequestInit) {
...
+          } else if (error.status >= 400 && shouldRetry) {
+            retryable = shouldRetry(error);
+          }
+        } else if (shouldRetry) {
+          retryable = shouldRetry(error);
+        }
+  
+        if (!retryable || isLast) throw error;
+  
+        const jitter = Math.floor(Math.random() * baseDelay);
```
Pauan: Maybe this should have an option for the amount of jitter, instead of tying it to the baseDelay?

Pauan (changes requested):
Thanks! I like this a lot. Just a couple very minor nits.

sdk/src/network-client.ts
christianwwwwwwww: good catch

sdk/src/network-client.ts
christianwwwwwwww: good callout thanks

sdk/src/utils.ts
christianwwwwwwww: Yeah good point -  I added a jitter option so developers can configure the max jitter independently from `baseDelay`

iamalwaysuncomfortable (approved):

Pauan (approved):

PR #1026, opened 2025-04-17
[Feature] Add setHeader and removeHeader methods for the AleoNetworkClient

### LINKED ISSUE #1023: Allow to set headers on programManager
https://github.com/ProvableHQ/sdk/blob/2759585439ef96d5a7e833811e8b8ec9349dffa8/sdk/src/program-manager.ts#L95

With `amareleo-chain`, one might run into CORS errors due to blocked `x-aleo-sdk-version` headers. It would be great if you could offer a possibility to overwrite the headers.

Image

alexpitsikoulis:
## Motivation

Presently the network client headers may only be set on initialization. If not specified, then a default header `X-Aleo-SDK-Version` is added. When a `ProgramManager` is constructed, it constructs the `AleoNetworkClient` with the default header. This can lead to [CORS issues](https://github.com/ProvableHQ/sdk/issues/1023).

## Changes

`setHeader` and `removeHeader` methods were created for the `AleoNetworkClient`. These methods were also added for `ProgramManager` and they act on the network client in the program manager. An optional parameter was also added to the `ProgramManager` constructor to allow initialization with desired headers.

## Test Plan

Tests for these new methods are added in `sdk/tests/network-client.test.ts` and `sdk/tests/program-manager.test.ts`

### DIFF
--- a/sdk/src/network-client.ts
+++ b/sdk/src/network-client.ts
@@ -37,7 +37,7 @@ class AleoNetworkClient {
     headers: { [key: string]: string };
     account: Account | undefined;
 
-    constructor(host: string, options?: AleoNetworkClientOptions) {
+    constructor(host: string, options?: AleoNetworkClientOptions | undefined) {
         this.host = host + "/%%NETWORK%%";
 
         if (options && options.headers) {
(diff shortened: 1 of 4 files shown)


iamalwaysuncomfortable (approved):
This is really well done thank you!

jaketarnow (approved):
Great work! LGTM

PR #1033, opened 2025-04-29
Updating to latest versions

Pauan:
Updates to the latest npm versions.


Pauan:
The error is caused by this, so we'll need to wait before we can get this merged:

https://github.com/uiwjs/react-codemirror/issues/731

PR #1038, opened 2025-05-10
Switching from CircleCI to GitHub Actions

Pauan:
We've had some issues with CircleCI, especially with permissions.

We currently use both CircleCI and GitHub Actions. This consolidates everything to only use GitHub Actions.

This simplifies the CI, and should fix the permission issues. This also allows us to re-run spurious tests.


Pauan:
Replaced by #1129

PR #1049, opened 2025-06-25
Add view key decryption for transitions and records

Roee-87:
## Motivation

The purpose of this PR is to create APIs that enable decryption of select transition outputs and records by third parties.  Specific transitions can be decrypted using the transition view key (tvk) while record decryption requires the record view key (rvk).  

Transition decryption is implemented as an additional associated method of the Transition struct while the logic for record decryption is implemented in a `decrypt.rs` file in the `Program` directory.

## Test Plan

Tests for transition decryption and record decryption, using the transition view key and record view key respectively, have been added to the `transition.rs`, `record_ciphertext.rs`, and `encrypt.rs` modules.  

For reviewers:  Run `yarn test` from this branch  to run  all of the the tests in the SDK.
Alternatively, use `wasm-pack test --node --features ${NETWORK} -- --nocapture ${TEST_NAME}` to run individual tests.

List of individual tests for both `testnet` and `mainnet`:
`test_record_view_key_generation`
`test_decrypt_with_record_vk`
`test_decrypt_record_symmetric`
`test_decrypt_record_with_wrong_rvk`

Mainnet test:
`wasm-pack test --node --features mainnet -- --nocapture test_transition_decryption_mainnet`

Testnet test: `wasm-pack test --node --features testate -- --nocapture test_transition_decryption`

PR #1050, opened 2025-07-03
Add methods and tests for tvk and rvk decryption in SDK

Roee-87:
## Motivation

This PR builds on [1049](https://github.com/ProvableHQ/sdk/pull/1049) and adds methods in the sdk for record decryption using the record view key.

## Test Plan

Tests are added in the test directory for the SDK

## Related PRs

(https://github.com/ProvableHQ/sdk/pull/1049)


sdk/src/account.ts
```diff
@@ -250,6 +263,103 @@ export class Account {
...
+  decryptTransitionWithViewKey(transition: string | Transition, transitionViewKey: string | Field): Transition {
+    if (typeof transition === 'string') {
+      transition = Transition.fromString(transition);
+    }
+    if (typeof transitionViewKey === 'string') {
+      transitionViewKey = Field.fromString(transitionViewKey);
+    }
+    return EncryptionToolkit.decryptTransitionWithVk(transitionViewKey);
+  }
+
```
iamalwaysuncomfortable:
```suggestion
```

Since these methods don't use anything in the account Object, we should delete these from it.

iamalwaysuncomfortable (commented):
There's a couple nits before this is approved:

* Delete the methods in the account object that don't use the key material within.
* Add tests for all the methods we introduced in wasm in JS
* Recreate the markdown docs with the documentation scripts (and ensure they're correct)

In future PRs we should consider encapsulating all of our encryption/decryption tools with the goal of allowing wallets to expose that fucntionality via adapaters, but we should strongly consider our approach to this before building this abstraction so just complete the items above and this can be mergeable.

sdk/src/account.ts
Roee-87: remove:  [bc8ac24](https://github.com/ProvableHQ/sdk/pull/1050/commits/bc8ac2492503ea0c76b9520a8abbe1b2fe23585f)

sdk/src/account.ts
```diff
@@ -127,6 +131,15 @@ export class Account {
...
 
+  /**
+   * Returns the Transition View Key associated with the transition public key.
+   * @returns {Field} The transition view key
+   * 
+   * @example
+   * import { Account } from "@provablehq/sdk/testnet.js";
+   * const account = 
+   */
+
```
iamalwaysuncomfortable:
```suggestion
```

iamalwaysuncomfortable (commented):
One final nit

iamalwaysuncomfortable (approved):

PR #1055, opened 2025-07-15
Add Boolean type

christianwwwwwwww:
## Motivation

Adds Boolean type to the wasm SDK with methods for serialization (to_string, to_bytes_le, to_bits_le, to_plaintext), logical operations (not, and, or, xor, nand, nor, equals), and helpers (random, zero, one, clone), along with from_* constructors.

### DIFF
--- a/sdk/tests/arithmetic.test.ts
+++ b/sdk/tests/arithmetic.test.ts
@@ -44,6 +44,58 @@ describe('Field and Group Arithmetic Tests', () => {
             expect(two_power_four.equals(two_times_eight)).equal(true);
         });
 
+        it('Check boolean creation and serialization', () => {
+            const t = Boolean.one();
+            const f = Boolean.zero();
+
+            expect(t.toString()).equals("true");
+            expect(f.toString()).equals("false");
+
+            const tBytes = t.toBytesLe();
+            const fBytes = f.toBytesLe();
+
+            const tFromBytes = Boolean.fromBytesLe(tBytes);
+            const fFromBytes = Boolean.fromBytesLe(fBytes);
+
+            expect(t.equals(tFromBytes)).equals(true);
+            expect(f.equals(fFromBytes)).equals(true);
+
+            const tBits = t.toBitsLe();
+            const fBits = f.toBitsLe();
+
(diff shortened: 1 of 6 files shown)


wasm/src/types/boolean.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    /// `false`
+    pub fn zero() -> Boolean {
+        Boolean(BooleanNative::zero())
+    }
+
+    /// `true`
+    pub fn one() -> Boolean {
+        Boolean(BooleanNative::one())
+    }
```
iamalwaysuncomfortable: These are deprecrated now, we can remove these (and the associated tests)

sdk/tests/arithmetic.test.ts
```diff
@@ -44,6 +44,58 @@ describe('Field and Group Arithmetic Tests', () => {
             expect(two_power_four.equals(two_times_eight)).equal(true);
         });
 
+        it('Check boolean creation and serialization', () => {
+            const t = Boolean.one();
+            const f = Boolean.zero();
```
iamalwaysuncomfortable: These should be removed as these methods are deprecated.

wasm/src/types/boolean.rs
```diff
@@ -0,0 +1,187 @@
...
+use wasm_bindgen::prelude::*;
+
+/// Boolean element.
+#[wasm_bindgen]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Boolean(BooleanNative);
+
+
+#[wasm_bindgen]
+impl Boolean {
```
iamalwaysuncomfortable, line 37: This function should have a constructor that takes a boolean and uses the `new` method on the native type.

iamalwaysuncomfortable (commented):
This PR as is from wasm does not compile.

Always make sure that from the top of the SDK repo, you to `yarn test:wasm && yarn test:sdk` to ensure your code builds and all tests run successfully.

iamalwaysuncomfortable (approved):
This is approved, excellent work @christianwwwwwwww. The failing tests have been addressed in a separate PR, so don't worry about fixing those here.

PR #1057, opened 2025-07-16
[Feature] Proving Requests

iamalwaysuncomfortable:
## Motivation

This PR creates the ability for the SDK to create the following objects.

### Proving Requests
Requests have the following structure. They  include an Authorization for both the top level function call and the fee authorization allowing any proving services to prove the authorization.

```rust
#[derive(Serialize, Deserialize)]
pub struct ProvingRequest<N: Network> {
    pub authorization: Authorization<N>,
    pub fee_authorization: Option<Authorization<A>>,
    pub broadcast: bool,
}
```

### Authorizations
The SDK is now capable of creating individual authorizations and allowing insepction of authorizations as wasm objects


sdk/src/program-manager.ts
```diff
@@ -696,6 +730,153 @@ class ProgramManager {
...
+     * if (!submitted) {
+     *   await programManager.networkClient.submitTransaction(tx);
+     * }
+     *
+     * // Wait to see if the transaction has appeared on chain.
+     * setTimeout(async () => {
+     *  const transaction = await programManager.networkClient.getTransaction(tx.id());
+     *  assert(transaction.id() === tx.id());
+     * }, 10000);
+     */
```
jaketarnow: This should be updated as there is no need for networkClient for crafting a request with generated payload

jaketarnow (changes requested):

jaketarnow (approved):
LGTM

jaketarnow (approved):
LGTM

PR #1081, opened 2025-08-27
Add auth inputs to delegated proving request

Roee-87:
## Motivation

This PR ammends the `submitProvingRequest` method to include API key and JWT inputs for authorized calls to the delegated proving service.

Scripts for obtaining the API key and JWT will be sent to select partners.  Eventually, authentication will be added to the SDK but that is beyond the scope of this PR.

The requirement for supplying a JWT will be deprecated in the near future in favor of providing an API key.

## Test Plan

Successful transactions are generated by the delegated proving service and returned to the user.
Proving requests would be expected to fail after the JWT expires (current expiration time is 30 minute).


sdk/src/network-client.ts
```diff
@@ -1599,36 +1599,48 @@ class AleoNetworkClient {
...
-        const prover_uri = url ? url : this.host;
-        const provingRequestString =
-            provingRequest instanceof ProvingRequest
-                ? provingRequest.toString()
-                : provingRequest;
+    async submitProvingRequest(
+        provingRequest: ProvingRequest | string, 
+        url?: string, 
+        apiKey?: string, 
+        delegatedProvingJWT?: string
```
iamalwaysuncomfortable: Per the lessons of the past with having multiple optional parameters. We should migrate this to a params object.

sdk/src/network-client.ts
```diff
@@ -1599,36 +1599,48 @@ class AleoNetworkClient {
...
+          "X-ALEO-METHOD": "submitProvingRequest",
+          "Content-Type": "application/json"
+        };
+
+        // Add auth header based on what's available
+        if (delegatedProvingJWT) {
+          headers["Authorization"] = `Bearer ${delegatedProvingJWT}`;
+        } else if (apiKey) {
+          headers["X-API-Key"] = apiKey;
+        }
```
iamalwaysuncomfortable: What possible tests can we develop around ensuring proper auth and API key @esren0x @dms120 do you have any suggestions?

@Roee-87 I'd like some tests around this if possible.

sdk/src/network-client.ts
```diff
@@ -1599,36 +1599,48 @@ class AleoNetworkClient {
      *
      * @param {ProvingRequest | string} provingRequest - The `ProvingRequest` (generated by the ProgramManager) to submit.
      * @param {string} url - (Optional) The url of the proving service.
+     * @param {string} apiKey - (Optional) API key for authentication.
+     * @param {string} delegatedProvingJWT - (Optional) JWT token for authentication. This will be deprecated in favor of apiKey in the future.
```
iamalwaysuncomfortable: All of these options except for the `ProvingRequest` to a ProvingRequestOptions object type in the `/models` directory so that people can specify options a-la-carte.

iamalwaysuncomfortable (changes requested):
A couple nits, but I appreciate the following things:

* The small size of the PR
* The dedication to authentication and shipping

Let's see what tests we can develop for this.

sdk/src/network-client.ts
Roee-87: done [4ae3d62](https://github.com/ProvableHQ/sdk/pull/1081/commits/4ae3d6271b9f980486b639ca14b33ddb216e358d)

sdk/src/network-client.ts
iamalwaysuncomfortable: @Roee-87 these params can now be specified via the object you defined above.

Roee-87: done [3aa109b](https://github.com/ProvableHQ/sdk/pull/1081/commits/3aa109bc28ddedfc4959acd6d4c74dc73d7f2baf)

iamalwaysuncomfortable (approved):

PR #1073, opened 2025-08-14
Implement snapshot query trait

christianwwwwwwww:
This PR adds a new `SnapshotQuery` struct to the SDK wasm and hooks it into the Program Manager's execute.rs as a replacement for `QueryNative`. 

Previously the query execution was consistently failing due to `stateRoot` mismatches happening as the query process was often slower than the ~2 second block time.

PR #1084, opened 2025-08-28
[Feature] Snapshot query trait extension for SnarkOS v4.2.1

iamalwaysuncomfortable:
## Motivation

SnarkOS node changes upcoming in SnarkOS v4.2.1 provide a `get_statepaths_for_commitments` endpoint which locks the block tree and ensure all statepaths fetched have a consistent stateroots so that transaction formation is guaranteed to succeed for valid record commitments. It also introduces a debug mode on the transaction broadcast endpoint which returns context on transaction broadcast failures if information within the Transaction was the cause of failure. 

To test these changes using client software written in JS/TS an advanced SDK release is necessary.

## Test Plan

Unit tests to properly test logic in this PR should be developed, but full integration tests will be developed closer to the time that SnarkOS v4.2.0 is released for testnet. 

## Related PRs

This PR builds on the excellent work done by @christianwwwwwwww in PR #1073

### DIFF
--- a/.github/workflows/staging-website.yml
+++ b/.github/workflows/staging-website.yml
@@ -18,7 +18,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly-2025-01-16
+          toolchain: nightly-2025-08-28
           override: true
           components: rustfmt, rust-src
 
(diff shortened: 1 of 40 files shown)


christianwwwwwwww (approved):
beautiful

PR #1092, opened 2025-09-15
Add cargo semver-checks to CI

vicsn:
## Motivation

Proof of concept to showcase: https://github.com/ProvableHQ/snarkVM/pull/2912

You may want to cache the `target` folder too like we do in that PR, where we get a speedup from from 40 minutes down to 15 minutes.

By default checks are made against the most recent published release. The approach I take in snarkVM is to manually append a `--baseline-rev eff84712b` if I want to just compare against a recent commit in between releases.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -152,6 +152,18 @@ jobs:
             cargo clippy --features testnet
             cargo clippy --features mainnet
 
+  check-cargo-semver:
+    executor: rust-node
+    steps:
+      - checkout
+      - cargo_fetch:
+          working_directory: wasm
+      - run:
+          working_directory: wasm
+          command: |
+            cargo install cargo-semver-checks@0.43.0 --locked
+            cargo semver-checks --workspace --features testnet
+
   check-fmt:
     executor: rust-node
     steps:
@@ -170,6 +182,7 @@ workflows:
   main-workflow:
     jobs:
       - check-fmt
+      - check-cargo-semver
       - check-clippy
       - sdk
       - sdk-test:


iamalwaysuncomfortable (commented):
This doesn't seem to work with our manual monomorphization for creating the two `wasm` binaries, a cursory 15 minutes of trying couldn't seem to get it to work, but we can investigate further later.

vicsn:
> This doesn't seem to work with our manual monomorphization for creating the two wasm binaries, a cursory 15 minutes of trying couldn't seem to get it to work, but we can investigate further later.

Yeah there's no rush on this. I see that it's `cargo doc` failing, which this relies on. It also failed for a snarkVM subcrate initially which I had to fix. So one added benefit of adding this in CI is that you also get a `cargo doc` check.

I think this might fix the issue at hand:

```
diff --git a/wasm/src/types/networks/mod.rs b/wasm/src/types/networks/mod.rs
index 67b6a928..7cb27bf4 100644
--- a/wasm/src/types/networks/mod.rs
+++ b/wasm/src/types/networks/mod.rs
@@ -18,10 +18,10 @@
 mod testnet;
 
 #[cfg(feature = "testnet")]
-pub use testnet::*;
+pub use testnet as snarkvm_network;
 
 #[cfg(feature = "mainnet")]
 mod mainnet;
 
 #[cfg(feature = "mainnet")]
-pub use mainnet::*;
+pub use mainnet as snarkvm_network;
```

vicsn:
I relent, on second look, the above is indeed a *lot* of work, not worth it right now is SDK interface is decently stable

iamalwaysuncomfortable:
If this can be given attention by a CI engineer, we're happy to have it, otherwise it might be too much of a lift on engineering resources at the moment to get this working against both the JS and rust code.

PR #1074, opened 2025-08-15
Update to RecordProvider Interface

alexpitsikoulis:
…scanner implementation of said interface

## Motivation

(Write your motivation here)

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)

PR #1100, opened 2025-09-22
[Feature] RecordScanner client

alexpitsikoulis:
## Motivation

The `RecordScanner` class which implements the `RecordProvider` interface serves as a means of access for the Record Scanner Service.

## Test Plan

Unit tests on initialization of the RecordScanner as well as all of its methods, ensuring that it is making the correct HTTP requests with the correct body and headers and returning the data received from these requests as expected.

## Related PRs

[Upgrade to `RecordProvider` interface](https://github.com/ProvableHQ/sdk/pull/1074)


iamalwaysuncomfortable:
@alexpitsikoulis I'll review this in a bit, basically though, as many tests as you can build proving it against a real RSS service with it, the better.

sdk/src/record-scanner.ts
```diff
@@ -0,0 +1,362 @@
...
+    readonly url: string;
+    private apiKey?: { header: string, value: string };
+    private uuid?: string;
+
+    constructor(options: RecordScannerOptions) {
+        this.url = options.url;
+        this.account = options.account;
+        this.apiKey = typeof options.apiKey === "string" ? { header: "X-Provable-API-Key", value: options.apiKey } : options.apiKey;
+    }
+
```
iamalwaysuncomfortable: We should add this function for computing the uuid.

```suggestion
    computeUUID(vk: ViewKey): Field {
        // Construct the material needed for the Poseidon oracle.
        const inputs = [Field.newDomainSeparator(RECORD_SCANNER_DOMAIN), vk().toField(), Field.one()]
        // Calculate the uuid.
        const hasher = new Poseidon4();
        return hasher.hash(inputs);
    }
```

You'll need to fold in this [PR 1105](This PR enables domain separators to be built: https://github.com/ProvableHQ/sdk/pull/1105.
) to make this work.

sdk/src/record-scanner.ts
```diff
@@ -0,0 +1,362 @@
...
+        this.apiKey = typeof options.apiKey === "string" ? { header: "X-Provable-API-Key", value: options.apiKey } : options.apiKey;
+    }
+    
+    /**
+     * Set the account to use for the record scanner.
+     * 
+     * @param {Account} account The account to use for the record scanner.
+     */
+    async setAccount(account: Account): Promise<void> {
+        this.uuid = undefined;
```
iamalwaysuncomfortable: We now have the tooling to calculate this, we should calculate the uuid based on the account.

```typescript
this.uuid = this.computeUUID(account.viewKey());
```

sdk/src/record-scanner.ts
```diff
@@ -0,0 +1,362 @@
...
+class RecordScanner implements RecordProvider {
+    account?: Account;
+    readonly url: string;
+    private apiKey?: { header: string, value: string };
+    private uuid?: string;
+
+    constructor(options: RecordScannerOptions) {
+        this.url = options.url;
+        this.account = options.account;
+        this.apiKey = typeof options.apiKey === "string" ? { header: "X-Provable-API-Key", value: options.apiKey } : options.apiKey;
```
iamalwaysuncomfortable, line 65: We should create the uuid here if an account is passed in.
```suggestion
        this.url = options.url;
        this.account = options.account;
        if this.account {
            this.uuid = computeUUID(this.account.viewKey()).toString();
        }
        this.apiKey = typeof options.apiKey === "string" ? { header: "X-Provable-API-Key", value: options.apiKey } : options.apiKey;
```

sdk/src/record-scanner.ts
```diff
@@ -0,0 +1,362 @@
...
+
+    /**
+     * Find records in the record scanner service.
+     * 
+     * @param {OwnedFilter} filter The filter to use to find the records.
+     * @returns {Promise<OwnedRecord[]>} The records.
+     */
+    async findRecords(filter: OwnedFilter): Promise<OwnedRecord[]> {
+        if (!this.uuid) {
+            throw new Error("Not registered");
```
iamalwaysuncomfortable: We should strive to have a Rust compiler level of informative errors so that users get good guidance for the next thing they should do if they experience an error.

```suggestion
            throw new Error("You are using the RecordScanner implementation of the RecordProvider. No account has been registered with the RecordScanner which is required to use the findRecords method. Please set an with the setAccount method before calling the findRecords method again.");
```

sdk/src/record-scanner.ts
```diff
@@ -0,0 +1,362 @@
...
+     * 
+     * @param {number} microcredits The amount of microcredits to find.
+     * @param {OwnedFilter} searchParameters The filter to use to find the record.
+     * @returns {Promise<OwnedRecord>} The record.
+     */
+    async findCreditsRecord(microcredits: number, searchParameters: OwnedFilter): Promise<OwnedRecord> {
+        try {
+            const records = await this.findRecords({
+                decrypt: true,
+                unspent: searchParameters.unspent ?? false,
```
iamalwaysuncomfortable: false shouldn't technically be the default here. If we don't find the unspent parameter here we should just not include it in the search, otherwise we'll always be serving unspent records.

sdk/src/record-scanner.ts
```diff
@@ -0,0 +1,362 @@
...
+
+            const record = records.find(record => {
+                const plaintext = RecordPlaintext.fromString(record.record_plaintext ?? '');
+                const amountStr = plaintext.getMember("microcredits").toString();
+                const amount = parseInt(amountStr.replace("u64", ""));
+                return amount >= microcredits;
+            });
+
+            if (!record) {
+                throw new Error("Record not found");
```
iamalwaysuncomfortable: We'll probably want to be a bit more informative on the error here.

```suggestion
                throw new Error(`No records found matching the supplied search filter:\n${JSON.stringify(searchParameters, null, 2)}`);
```

sdk/src/record-scanner.ts
```diff
@@ -0,0 +1,362 @@
...
+     * 
+     * @param {number[]} microcreditAmounts The amounts of microcredits to find.
+     * @param {OwnedFilter} searchParameters The filter to use to find the records.
+     * @returns {Promise<OwnedRecord[]>} The records
+     */
+    async findCreditsRecords(microcreditAmounts: number[], searchParameters: OwnedFilter): Promise<OwnedRecord[]> {
+        try {
+            const records = await this.findRecords({
+                decrypt: true,
+                unspent: searchParameters.unspent ?? false,
```
iamalwaysuncomfortable: Ditto here if we don't find the unspent flag, we shouldn't set it to false as default because that'll return only `spent` records. We just want to return spent OR unspent records in this case, and we just don't specify the flag if that's the case.

sdk/src/record-scanner.ts
iamalwaysuncomfortable: And the reason for this is we don't want to enforce a call to `register` each time a process spins up. If they have already registered, we want to be able to compute the UUID locally so they don't have to call that endpoint.

sdk/tests/record-scanner.test.ts
```diff
@@ -0,0 +1,423 @@
+import { expect } from "chai";
+import sinon from "sinon";
+import { RecordScanner } from "../src/record-scanner";
+import { Account } from "../src/account";
+import { CHECK_SNS_RESPONSE, CHECK_TAGS_RESPONSE, ENCRYPTED_RECORDS, OWNED_RECORDS } from "./data/records";
+import { RecordsResponseFilter } from "../src/models/record-scanner/recordsResponseFilter";
+import { RecordsFilter } from "../src/models/record-scanner/recordsFilter";
+import { OwnedFilter } from "../src/models/record-scanner/ownedFilter";
+import { OwnedRecordsResponseFilter } from "../src/models/record-scanner/ownedRecordsResponseFilter";
```
iamalwaysuncomfortable: Just make sure these are organized alphabetically

iamalwaysuncomfortable (commented):
Overall really great job here Alex, the tests are good. Left some comments, but they're mostly small changes.

What we want to do prior to folding this in is test it against an actual service. 

Can you write up another test file that tests it against the service? We can store all the test data related to that in envars for ci.

sdk/src/record-scanner.ts
alexpitsikoulis: done

sdk/src/record-scanner.ts
alexpitsikoulis: done

sdk/src/record-scanner.ts
alexpitsikoulis: done

sdk/src/record-scanner.ts
alexpitsikoulis: done

sdk/src/record-scanner.ts
alexpitsikoulis: done

sdk/src/record-scanner.ts
alexpitsikoulis: done

sdk/src/record-scanner.ts
alexpitsikoulis: done

sdk/tests/record-scanner.test.ts
alexpitsikoulis: done

iamalwaysuncomfortable (approved):

iamalwaysuncomfortable:
Excellent work @alexpitsikoulis this is set to merge.

PR #1105, opened 2025-09-30
[Feature] Domain separators

iamalwaysuncomfortable:
## Motivation

In order to create new key types and deterministic nonces, users of the SDK need to be able to create domains separators as field elements. This PR adds a method to the `Field` type to create standard Aleo domain separators as field elements.


alexpitsikoulis (approved):
LGTM

PR #1113, opened 2025-10-27
Adding in dynamic loadNetwork function

Pauan:
Right now networks have to be imported statically, like this:

```ts
import { ProgramManager } from "@provablehq/sdk/testnet.js";

function run() {
    const programManager = new ProgramManager();

    // ...
}

run();
```

However that locks the code into the `testnet` network. It's common for code to want to work with multiple networks.

This PR adds in a new `loadNetwork` function which allows for dynamically loading a network:

```ts
import { loadNetwork } from "@provablehq/sdk/dynamic.js";

const network = await loadNetwork("testnet");

function run(network) {
    const programManager = new network.ProgramManager();

    // ...
}

run(network);
```

It will dynamically load the network at runtime, and code can now be written so that it will work with multiple networks.


iamalwaysuncomfortable (approved):
Thanks for getting this in, this simple ergonomic change will help people out a lot a great deal.

PR #1129, opened 2025-11-22
Switching from CircleCI to GitHub Actions

Pauan:
We've had some issues with CircleCI, especially with permissions.

We currently use both CircleCI and GitHub Actions. This consolidates everything to only use GitHub Actions.

This simplifies the CI, and should fix the permission issues. This also allows us to re-run spurious tests.


vicsn:
Hi! Could this also trigger a CI run after a PR is merged, on the post-merge commit? Or would that be out of scope? :)

Pauan:
@vicsn What would be the benefit of doing that?

vicsn:
> @vicsn What would be the benefit of doing that?

To avoid CI running on older branch logic. The only way to get CI to run on the latest logic is to:
- either enforce that the PR branch syncs with `origin/mainnet` and pass CI **before** merging. Not sure if this is a github setting or already regular practice in this repo.
- or run CI **after** the PR is merged on the post-merge commit. This is what is done in e.g. snarkVM/snarkOS.

Pauan:
@vicsn We already enforce that the pull request must be up-to-date with the `mainnet` branch:

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-status-checks-before-merging

If the pull request is out of date, it cannot be merged until it is updated. This is the default behavior for GitHub.

So the CI tests will always be up to date before merging.

iamalwaysuncomfortable (approved):
This is excellent @Pauan this is approved!

PR #1134, opened 2025-11-28
[Feature]  Add method that returns a program's address from it's program ID string

Roee-87:
## Motivation

This PR adds an associated method to the Address object that allows a user to derive a program's address from its Program ID.

## Test Plan

A test was added in `wasm.test.ts` to ensure that `credits.aleo` input returns the correct address for both testnet and mainnet.


wasm/src/account/address.rs
```diff
@@ -165,6 +166,31 @@ impl Address {
...
+        let network_field = program_id_native.network().to_field().map_err(|e| e.to_string())?;
+
+        // Create a JS array of the fields.
+        let fields_array = js_array_from_fields!(&[name_field, network_field]);
+
+        // Initialize the Poseidon4 hasher.
+        let hasher = Poseidon4::new();
+
+        // Compute the group element corresponding to the program address.
+        let group = hasher.hash_to_group(fields_array).map_err(|e| e.to_string())?;
```
iamalwaysuncomfortable: Since at this point we're already in the context of `wasm` execution, we don't need use the SDK's external API in order to build this as it requires more roundtrips into JS land than necessary.

Once you have the name and network field this can be simplified to:
```suggestion
        // Compute the group element corresponding to the program address.
        let group = <CurrentNetwork as Network>::hash_to_group_psd4(&[name_field, network_field])?;
        Ok(Address::from(AddressNative::from(group)))
```

iamalwaysuncomfortable (changes requested):
Some small changes requested. 

After you've made them, make this PR green so CI can start on it.

iamalwaysuncomfortable (approved):

PR #1136, opened 2025-12-02
[Fix] Fee checks

iamalwaysuncomfortable:
## Motivation

Method within the program manager that send transactions on the user's behalf were doing fee checks on public balances even when private fees were specified, this PR fixes this behavior and forgoes fee checks when the fee is specified as private.


Roee-87 (approved):
LGTM

PR #1142, opened 2025-12-03
upddate yarn.lock

14MR:
## Motivation

`yarn install --frozen-lockfile` is failing now, this PR fixes it by updating lock file


iamalwaysuncomfortable (approved):

iamalwaysuncomfortable (changes requested):
Can you ensure this is run with the latest version of yarn first?

PR #1144, opened 2025-12-03
[Feature] Replace API key with JWT in submitProvingRequest method

Roee-87:
## Motivation

This PR modifies the `submitProvingRequest` method to accept a jwt instead of an API key.

## Test Plan

The following script was used to test a DPS request:
```typescript
import {Account, initThreadPool, ProgramManager, AleoKeyProvider, OfflineKeyProvider} from "@provablehq/sdk";

await initThreadPool();

const apiKey= "{YOUR_API_KEY_HERE}";
const consumerId = "{YOUR_CONSUMER_ID_HERE}";
const endpoint = " https://api.provable.com/prove/testnet/prove";


async function localProgramExecution() {
    const programManager = new ProgramManager();

    // Create a temporary account for the execution of the program
    const account = new Account({privateKey: "{YOUR_PRIVATE_KEY_HERE}"});
    programManager.setAccount(account);

    // Create a key provider in order to re-use the same key for each execution
    const keyProvider = new AleoKeyProvider();
    const offlineKeyProvider = new OfflineKeyProvider();
    keyProvider.useCache(true);
    programManager.setKeyProvider(keyProvider);

    // Build a proving request for the "spin" function of "puzzle_spinner_v002.aleo".
    const provingRequest = await programManager.provingRequest({
        programName: "credits.aleo",
        functionName: "transfer_public",
        baseFee: 0.01,
        priorityFee: 0,
        privateFee: false,
        inputs: [
            "{YOUR_ADDRESS_HERE}",
            "1000000u64",
        ],
        broadcast: false,
    });

    // Submit the proving request to the network client.
    const provingResponse = await programManager.networkClient.submitProvingRequest({
        provingRequest: provingRequest,
        url: endpoint,
        apiKey: apiKey,
        consumerId: consumerId,
});
    console.log("Proving response:", provingResponse);
}

const start = Date.now();
console.log("Starting execute!");
await localProgramExecution();
console.log("Execute finished!", Date.now() - start);
```

### DIFF
--- a/sdk/src/network-client.ts
+++ b/sdk/src/network-client.ts
@@ -1,5 +1,6 @@
 import { get, post, parseJSON, logAndThrow, retryWithBackoff, environment } from "./utils.js";
 import { Account } from "./account.js";
+import { FIVE_MINUTES } from "./constants.js";
 import { BlockJSON } from "./models/blockJSON.js";
 import { TransactionJSON } from "./models/transaction/transactionJSON.js";
 import {
@@ -60,6 +61,8 @@ interface DelegatedProvingParams {
  *
  * // Connection to a public beacon node
  * const account = Account.fromCiphertext(process.env.ciphertext, process.env.password);
+ * const apiKey = process.env.apiKey;
+ * const consumerId = process.env.consumerId;
  * const publicNetworkClient = new AleoNetworkClient("http://api.explorer.provable.com/v1", undefined, account);
  */
 class AleoNetworkClient {
@@ -1692,7 +1695,7 @@ class AleoNetworkClient {
         let jwtData = options.jwtData ?? this.jwtData
 
         // Check if JWT is expired or missing
-        const bufferTime = 5 * 60 * 1000; // 5 minutes buffer
+        const bufferTime = FIVE_MINUTES; // 5 minutes buffer
         const isExpired = jwtData && Date.now() >= jwtData.expiration - bufferTime;
         if (!jwtData || isExpired) {
             if (options.apiKey && options.consumerId) {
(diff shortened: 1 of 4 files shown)


alexpitsikoulis:
So overall I think what would be best is to keep the API key as part of the object and introduce a fetch/refresh JWT method. That way you don't have to manually update your JWT. JWT fetch/refresh is achieved by making a POST request to `/jwts/{{consumer_id}}` with the API key as the `X-Provable-API-Key` header. This will return the token expiration timestamp in the body and the token in the `Authorization` header of the request. At the beginning of the `submitProvingRequest` function it could check the exp (store as object state from the JWT request) and if its expired or close it could make the JWT request again. Could have that happen in a retry as well if you get a 401 or something for expired or invalid token.

alexpitsikoulis:
I think it makes sense to store the api key and jwt as fields of the class. Refresh would update `this.jwt` or something to that effect and the JWT is refreshed using `this.api_key`. @iamalwaysuncomfortable what do you think?

alexpitsikoulis (approved):
LGTM

sdk/src/network-client.ts
```diff
@@ -1638,24 +1687,42 @@ class AleoNetworkClient {
             ? options.provingRequest.toString()
             : options.provingRequest;
 
-        // Build headers with proper auth fallback
+        const apiKey = options.apiKey ?? this.apiKey;
+        const consumerId = options.consumerId ?? this.consumerId;    
+        let jwtData = options.jwtData ?? this.jwtData
+
+        // Check if JWT is expired or missing
+        const bufferTime = 5 * 60 * 1000; // 5 minutes buffer
```
iamalwaysuncomfortable: This is a magic number, please move this to the constants file.

iamalwaysuncomfortable (approved):
Move the parameter to the constants file, once that's done, I consider this approved.

sdk/src/network-client.ts
Roee-87: done [48b5f40](https://github.com/ProvableHQ/sdk/pull/1144/commits/48b5f40968b31ba13187ee84991df819b564d287)

PR #1147, opened 2025-12-11
Fix - Prevent excessive API calls in getProgramImports due to circular dependencies

### LINKED ISSUE #849: [Bug] Program imports don't go in the right order
## 🐛 Bug Report

Imports must be added to process in the right order, however they are not.
The code with the bug is somewhere around here: https://github.com/AleoHQ/sdk/blob/a556cd4a3b81eb40002b7aa7bbb308dfe4700b92/sdk/src/program-manager.ts#L208

(Write your description here)

## Steps to Reproduce

Call `football_game_v010.aleo/main(true)` from https://aleo.tools/develop.

None

image

## Expected Behavior

It should produce a transaction.

christianwwwwwwww:
## Motivation

Problem
The getProgramImports() function was making hundreds of redundant API calls when processing programs with overlapping imports (e.g., Pondo AMM). This caused rate limiting errors (429) and extremely slow transaction building in wallets.
Example:
Expected: 24 programs = 24 API calls
Actual: 24 programs = 200+ API calls 

Each recursive call created a new imports object, so the duplicate check only worked at the current recursion level. Programs imported by multiple dependencies were fetched repeatedly.

## Test Plan

Tested with `amm_orcl_intrfc_v.aleo` (23 imports with lots of overlap). Before it was doing exponential calls and getting rate limited, now it just does 23 calls.

Added a test in `/tests/network-client.test.ts` to ensure no duplicate program calls with a particularly problematic program.

### DIFF
--- a/sdk/src/network-client.ts
+++ b/sdk/src/network-client.ts
@@ -1059,19 +1059,40 @@ class AleoNetworkClient {
      * assert.deepStrictEqual(programImports, expectedImports);
      */
     async getProgramImports(inputProgram: Program | string): Promise<ProgramImports> {
-        try {
-            this.ctx = { "X-ALEO-METHOD": "getProgramImports" };
-            const imports: ProgramImports = {};
+            try {
+                this.ctx = { "X-ALEO-METHOD": "getProgramImports" };
+                const imports: ProgramImports = {};
+                const visited = new Set<string>();
+                
+                await this._getProgramImportsRecursive(inputProgram, imports, visited);
+                
+                return imports;
+            } catch (error: any) {
+                logAndThrow("Error fetching program imports: " + error.message);
+            } finally {
+                this.ctx = {};
+            }
+        }
 
+        private async _getProgramImportsRecursive(
+            inputProgram: Program | string,
+            imports: ProgramImports,
+            visited: Set<string>
+        ): Promise<void> {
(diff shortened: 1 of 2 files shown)


sdk/src/network-client.ts
```diff
@@ -1102,6 +1106,11 @@ class AleoNetworkClient {
                 }
             }
 
+            // Skip if already processed (prevents infinite recursion on circular dependencies)
+            if (imports.hasOwnProperty(programId)) {
```
miazn: i think this is unnecessary , line 1111 was already doing this check. I think all you have to do here is the optional imports param.

sdk/src/network-client.ts
```diff
@@ -1119,10 +1109,15 @@ class AleoNetworkClient {
                 const import_id = importList[i];
                 if (!imports.hasOwnProperty(import_id)) {
                     const programSource = <string>await this.getProgram(import_id);
+                    const nestedImports = <ProgramImports>await this.getProgramImports(programSource, imports);
```
miazn, line 1112: there was actually a second bug here which was causing an EXTRA call on top of the recursion, this function expects the actual program and not the program name. so it calls twice

iamalwaysuncomfortable (approved):

PR #1151, opened 2025-12-18
Simplify XMLHttpRequest polyfill

eranrund:
## Motivation

Hyperlane encountered a bundling issue with the Provable SDK due to its dependency on `sync-request`. One of `sync-request`’s transitive dependencies (`sync-rpc`) relies on legacy CommonJS require() syntax, which breaks Hyperlane’s ESM-based build:

```
node cli-bundle/index.js --version                               took  6s at  14:18:36
file:///Users/troykessler/work/hyperlane/hyperlane-monorepo/typescript/cli/cli-bundle/index.js:389077
    throw new Error(
          ^

Error: file:///Users/troykessler/work/hyperlane/hyperlane-monorepo/typescript/cli/cli-bundle/find-port.js:3
const getPort = require('get-port');
                ^

ReferenceError: require is not defined in ES module scope, you can use import instead
This file is being treated as an ES module because it has a '.js' file extension and '/Users/troykessler/work/hyperlane/hyperlane-monorepo/typescript/cli/cli-bundle/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
    at file:///Users/troykessler/work/hyperlane/hyperlane-monorepo/typescript/cli/cli-bundle/find-port.js:3:17
    at ModuleJob.run (node:internal/modules/esm/module_job:371:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:702:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)
```

```
@hyperlane-xyz/monorepo
└─┬ @hyperlane-xyz/cli@19.11.0 -> ./typescript/cli
  └─┬ @hyperlane-xyz/aleo-sdk@19.11.0 -> ./typescript/aleo-sdk
    └─┬ @provablehq/sdk@0.9.13
      └─┬ sync-request@6.1.0 <- package only used in xml polyfill
        └─┬ sync-rpc@1.3.6 <- this package uses old commonjs require syntax which breaks on our side
          └── get-port@3.2.0
```

The SDK currently uses both:
* `xmlhttprequest-ssl`
* `sync-request`
to polyfill XMLHttpRequest in Node.js.

This dual approach exists because the version of `xmlhttprequest-ssl` previously in use incorrectly decoded synchronous responses as UTF-8, corrupting binary data. To work around this, synchronous requests were routed through `sync-request`, which preserves raw bytes correctly.

However, newer versions of `xmlhttprequest-ssl` now handle response decoding properly, eliminating the need for `sync-request` entirely.

Note that sync requests are not desirable and long-term we should move away from that entirely, but that requires changes to snarkVM before the JS SDK can support that.

## Test Plan
To verify this work I ran Hyperlane's test utility (https://github.com/troykessler/aleo-sdk-deploy) and was able to get it to work ([with some minor changes](https://github.com/troykessler/aleo-sdk-deploy/pull/1)).

I've also ran the [aleo-multisig unit tests](https://github.com/AleoNet/aleo-multisig/tree/main/tests) with this modified SDK and they passed.

### DIFF
--- a/sdk/package.json
+++ b/sdk/package.json
@@ -56,7 +56,6 @@
     "comlink": "^4.4.2",
     "core-js": "^3.40.0",
     "mime": "^4.0.6",
-    "sync-request": "^6.1.0",
     "xmlhttprequest-ssl": "^4.0.0"
   },
   "devDependencies": {
--- a/sdk/package.json
+++ b/sdk/package.json
@@ -57,7 +57,7 @@
     "core-js": "^3.40.0",
     "mime": "^4.0.6",
     "sync-request": "^6.1.0",
-    "xmlhttprequest-ssl": "^3.1.0"
+    "xmlhttprequest-ssl": "^4.0.0"
   },
   "devDependencies": {
     "@rollup/plugin-replace": "^6.0.2",
--- a/sdk/src/polyfill/xmlhttprequest.ts
+++ b/sdk/src/polyfill/xmlhttprequest.ts
@@ -2,8 +2,9 @@
 import $xmlhttprequest from "xmlhttprequest-ssl";
 
 if (globalThis.XMLHttpRequest == null) {
-    function XMLHttpRequest(opts?: any) {
-        return new $xmlhttprequest.XMLHttpRequest({syncPolicy: "enabled", ...opts});
-    }
-    (globalThis as any).XMLHttpRequest = XMLHttpRequest;
+    (globalThis as any).XMLHttpRequest = class extends $xmlhttprequest.XMLHttpRequest {
+        constructor(opts?: any) {
+            super({ syncPolicy: "enabled", ...opts });
+        }
+    };
 }
--- a/sdk/src/polyfill/xmlhttprequest.ts
+++ b/sdk/src/polyfill/xmlhttprequest.ts
@@ -1,67 +1,9 @@
 // @ts-ignore
 import $xmlhttprequest from "xmlhttprequest-ssl";
-import $request from "sync-request";
 
 if (globalThis.XMLHttpRequest == null) {
-    globalThis.XMLHttpRequest = class extends $xmlhttprequest.XMLHttpRequest {
-        // We have to override the methods inside of the `constructor`
-        // because `xmlhttprequest-ssl` doesn't use a regular class,
-        // instead it defines all of the methods inside of the constructor.
-        constructor(...args: Array<any>) {
-            super(...args);
-
-            const open = (this as any).open;
-            const send = (this as any).send;
-
-            let _async: boolean = true;
-            let _url: null | string = null;
-            let _mime: string = "text/xml";
-
-            function reset() {
-                _async = true;
-                _url = null;
-                _mime = "text/xml";
-            }
-
-            (this as any).open = function (method: string, url: string, async: boolean, user?: string, password?: string) {
-                // Special behavior for synchronous requests
-                if (method === "GET" && !async && !user && !password) {
-                    _async = false;
-                    _url = url;
-
-                // Default to the normal polyfill for async requests
-                } else {
-                    reset();
-                    return open.call(this, method, url, async, user, password);
-                }
-            };
-
-            (this as any).send = function (data: any) {
-                if (_async) {
-                    return send.call(this, data);
-
-                // Use `sync-request` for synchronous requests.
-                } else {
-                    const response = $request("GET", _url!, {
-                        headers: {
-                            "Content-Type": _mime,
-                        }
-                    });
-
-                    const buffer = (response.body as Buffer).buffer as any;
-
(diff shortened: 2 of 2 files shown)


sdk/package.json
```diff
@@ -57,7 +57,7 @@
     "core-js": "^3.40.0",
     "mime": "^4.0.6",
     "sync-request": "^6.1.0",
```
troykessler: I think you can remove this dependency since it is unused now

sdk/src/polyfill/xmlhttprequest.ts
```diff
@@ -1,67 +1,9 @@
...
-                }
-            };
-
-            (this as any).overrideMimeType = function (mime: string) {
-                _mime = mime;
-            };
-        }
-    } as any;
-}
+    function XMLHttpRequest(opts?: any) {
```
Pauan: Why was this refactored to no longer be a `class`?

Pauan (changes requested):

sdk/src/polyfill/xmlhttprequest.ts
eranrund: Does it need to be a class? I believe this is working okay, and I think when I tried leaving it as a class it didn't work.
This might have to do with a change between xmlhttprequest-ssl 3.x and 4.x. In 3.x they were doing this:
https://github.com/mjwwit/node-XMLHttpRequest/blob/3.1.0/lib/XMLHttpRequest.js#L280
vs in 4.x:
https://github.com/mjwwit/node-XMLHttpRequest/blob/master/lib/XMLHttpRequest.js#L422

What is the issue with how it is currently done?

Pauan: It's not "wrong" per se, but... using functions as pseudo-classes is *extremely* old and outdated, it's not best practice and it's no longer recommended.

`class` is the modern standardized way to create classes, and `class` has various convenient features. So it's better to use it if we can.

This code should work...

```js
globalThis.XMLHttpRequest = class extends $xmlhttprequest.XMLHttpRequest {};
```

...or possibly even this code...

```js
globalThis.XMLHttpRequest = $xmlhttprequest.XMLHttpRequest;
```

eranrund: Updated to using the `class` syntax.

Pauan (approved):
Looks good to me, thanks!

Pauan:
@eranrund By the way... the SDK works natively on Node, so I'm curious why you needed to patch some things to make it work. What version of Node are you using?

eranrund:
> @eranrund By the way... the SDK works natively on Node, so I'm curious why you needed to patch some things to make it work. What version of Node are you using?

Please  see the PR description - the stock SDK depends on `sync-request`, which uses an outdated javascript module system and it broke the build of one of Aleo's partners (Hyperlane).

It works fine on NodeJS but their build process failed due to the `sync-request` dependency. That package is also several years old and no longer maintained.

Pauan:
> Please see the PR description - the stock SDK depends on sync-request, which uses an outdated javascript module system and it broke the build of one of Aleo's partners (Hyperlane).

Yes, I know about that, but I'm referring to this:

https://github.com/troykessler/aleo-sdk-deploy/pull/1/files#diff-cd55066e51ef17ad8b88f13f571779227571ac127f85f9fea83ce78d1b033ddcR27-R35

> It works fine on NodeJS but their build process failed due to the sync-request dependency. That package is also several years old and no longer maintained.

We test with various build systems, all of them can handle `require`, so I'm not sure what they're doing differently.

Also, `sync-request` is not the only dependency that uses `require`, so I'm not sure why they're having issues *specifically* with `sync-request` but not other packages.

eranrund:
> > Please see the PR description - the stock SDK depends on sync-request, which uses an outdated javascript module system and it broke the build of one of Aleo's partners (Hyperlane).
> 
> Yes, I know about that, but I'm referring to this:
> 
> https://github.com/troykessler/aleo-sdk-deploy/pull/1/files#diff-cd55066e51ef17ad8b88f13f571779227571ac127f85f9fea83ce78d1b033ddcR27-R35
> 
> > It works fine on NodeJS but their build process failed due to the sync-request dependency. That package is also several years old and no longer maintained.
> 
> We test with various build systems, all of them can handle `require`, so I'm not sure what they're doing differently.
> 
> Also, `sync-request` is not the only dependency that uses `require`, so I'm not sure why they're having issues _specifically_ with `sync-request` but not other packages.

To be honest, I am not super familiar with their build so I am not sure why they ran into this and others haven't.
I believe the patches I had to make to that script you linked to are because I am using `file:../sdk/sdk` to use my local SDK build, and I don't think this will be needed once the SDK is published and we are back to using a standard versioned dependency in `package.json`.

Pauan:
Alright, understood. :+1:

PR #1157, opened 2026-01-07
Simplify XMLHttpRequest polyfill (#1151)

iamalwaysuncomfortable:
## Motivation

The XMLHttpRequest Polyfill is required because SnarkVM downloads parameters synchronously within SnarkVM requiring a synchronous network call within Javascript. Since XMLHttpRequest isn't included in the NodeJS default library, it is provided via a NodeJs polyfill. Previously this Polyfill used the `sync-request` polyfill but this dependency introduces a reliance on a package incompatible with the latest typescript module resolution strategies.

This PR uses work from @eranrund to use the xmlhttp-ssl library which doesn't introduce problematic dependencies.


Pauan (approved):
LGTM!

PR #1138, opened 2025-12-02
[Feature] Add KeyStore interface for persistent Proving and Verifying Key storage

iamalwaysuncomfortable:
## Motivation

Aleo functions require a proving key and verifying key per function. On execution, if the keys do not currently exist, they are synthesized and stored in the local `Process` object. However This synthesis process takes a long time and execution time can be cut down drastically by inserting pre-synthesized keys. Normally on executors with filesystems, these keys are stored locally on disk automatically by `SnarkVM`. However `wasm` has no direct access to a filesystem, thus proving and verifying keys must be passed into `wasm` externally by JS. 

Because JS can operate in multiple contexts (such as the `browser`, `node.js` on a machine with a filesystem, or perhaps an enclave environment) key storage will necessarily need to be different depending on the operational context.

Currently there no convenient abstractions in the SDK for managing proving and verifying keys and it is not clear to many users how to do proper proving and verifying key management. This PR adds a `KeyStore` abstraction that provides methods for storing and retrieving proving and verifying keys and provides a default implementation called the `FileSystemKeyProvider` for NodeJS sdk users that have access to filesystems. 

This PR also adds the `KeyStore` object to the `FunctionKeyProvider` interface for convenient access.

## Test Plan

* Several unit tests should be written against the `FileSystemKeyProvider` to ensure the `KeyStore` interface operates as desired.
* Unit tests should test the new keyStore() method against the `OfflineKeyProvider` and `AleoKeyProvider` to ensure these interfaces work as desired.

---

> [!NOTE]
> **Medium Risk**
> Adds new filesystem-based persistence and integrity verification for cryptographic key material, which can impact execution correctness if verification/persistence is wrong. Changes also refactor public exports/import paths, increasing risk of integration/build breakage for downstream consumers.
> 
> **Overview**
> Adds a new persistent key storage layer via `KeyStore`/`KeyLocator` plus a Node-only `LocalFileKeyStore` that can `set`/`get` proving+verifying key bytes (or WASM `ProvingKey`/`VerifyingKey`), keep per-key fingerprint metadata on disk, and verify integrity when reading (or when an expected fingerprint is provided).
> 
[...]

PR #1165, opened 2026-01-21
[feat] IndexedDB interface for KeyStore

kpandl:
This PR adds an IndexedDB interface for KeyStore.

I thought it may be good if the app is able to have two IndexedDBs, one for mainnet, and one for testnet. Thus, it can persist proving keys even after switching the network.
This feature added quite some lines of code to the diff, I'm not sure if this is the best way to implement it. Happy to discuss or to also just have the PR without two separate DBs for the two networks.

## Test Plan

No tests ran yet. Happy to discuss the best ways to test this PR.

## Related PRs

https://github.com/ProvableHQ/sdk/pull/1138


iamalwaysuncomfortable:
Closing this in favor of an application specific implementation.

PR #1172, opened 2026-01-29
[DPS] Add functionality for encrypting ProvingRequests

iamalwaysuncomfortable:
## Motivation

This PR adds tooling to the SDK to encrypt ProvingRequests before being sent to the DPS service. It does so by encrypting the ProvingRequest with a cryptobox key and sending it to the DPS endpoint. Further tooling for encrypting view keys and individual authorizations are included as well.

## Test Plan
Tests against a staging service are included.

### DIFF
--- a/sdk/src/models/provingResponse.ts
+++ b/sdk/src/models/provingResponse.ts
@@ -2,14 +2,14 @@ import { TransactionJSON } from "./transaction/transactionJSON.js";
 
 /** HTTP status and optional message from snarkOS broadcast (Accepted/Rejected variants). */
 export interface BroadcastResponse {
-    status_code: bigint;
+    status_code: bigint | number;
     message?: string;
 }
 
 /** Result of the optional broadcast step. Discriminated by `status`. */
 export type BroadcastResult =
-    | { status: "Accepted"; status_code: number; message?: string }
-    | { status: "Rejected"; status_code: number; message?: string }
+    | { status: "Accepted"; status_code: bigint | number; message?: string }
+    | { status: "Rejected"; status_code: bigint | number; message?: string }
     | { status: "Failed"; message: string }
     | { status: "Skipped" };
 
@@ -26,7 +26,7 @@ export interface ProveApiErrorBody {
 
 /** Error thrown on prove API failure; `status` is set for retry logic (e.g. retryWithBackoff checks error.status >= 500). */
 export interface ProvingRequestError extends Error {
-    status?: number;
+    status?: bigint | number;
 }
 
 /** Success variant of a proving request result. */
@@ -38,11 +38,11 @@ export interface ProvingSuccess {
 /** Failure variant of a proving request result (HTTP 400, 500, 503). */
 export interface ProvingFailure {
     ok: false;
-    status: number;
+    status: bigint | number;
     error: ProveApiErrorBody;
 }
 
-/** Result of a proving request. Use this when you want to handle status yourself instead of throwing. */
+/** Result of a proving request. Type used to give callers the ability to self-handle errors. */
 export type ProvingResult = ProvingSuccess | ProvingFailure;
 
 /** Type guard: value is a ProvingResponse. */
--- a/sdk/src/models/provingResponse.ts
+++ b/sdk/src/models/provingResponse.ts
@@ -1,11 +1,12 @@
 import { TransactionJSON } from "./transaction/transactionJSON";
 
-export interface BroadcastResult {
+export interface BroadcastResponse {
     status_code: bigint,
     status?: string
 }
 
 export interface ProvingResponse {
-    transaction: TransactionJSON,
-    broadcast_result: BroadcastResult,
+    message?: string, // Potential error message.
+    transaction?: TransactionJSON, // Transaction if successful.
+    broadcast_result?: BroadcastResponse, // Broadcast result if successful.
 }
--- a/sdk/src/models/provingResponse.ts
+++ b/sdk/src/models/provingResponse.ts
@@ -1,6 +1,11 @@
 import { TransactionJSON } from "./transaction/transactionJSON";
 
+export interface BroadcastResult {
+    status_code: bigint,
+    status?: string
+}
+
 export interface ProvingResponse {
     transaction: TransactionJSON,
-    broadcast?: boolean,
+    broadcast_result: BroadcastResult,
 }
--- a/sdk/src/models/provingResponse.ts
+++ b/sdk/src/models/provingResponse.ts
@@ -1,12 +1,66 @@
-import { TransactionJSON } from "./transaction/transactionJSON";
+import { TransactionJSON } from "./transaction/transactionJSON.js";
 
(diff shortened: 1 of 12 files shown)


sdk/src/models/provingResponse.ts
```diff
@@ -1,6 +1,11 @@
 import { TransactionJSON } from "./transaction/transactionJSON";
 
+export interface BroadcastResult {
```
niklaslong: [Linking](https://github.com/ProvableHQ/delegated-proving-service/blob/ef4a5675cdcfbe746918e1b8b72c9f1a22381ea9/utils/src/types.rs#L185) the broadcast result on the prover for reference.

sdk/src/network-client.ts
```diff
@@ -1716,9 +1785,44 @@ class AleoNetworkClient {
...
         }
 
+        // If private DPS is enabled, send an encrypted payload.
+        if (options.dpsPrivacy) {
+            try {
+                // One pubkey per request is required, so the entire flow is wrapped in a retry.
+                const response = await retryWithBackoff(async () => {
+                    // First, get the service pubkey.
+                    const pubKeyResponse = await get(proverUri + '/pubkey', {
+                        headers
```
niklaslong: Just noting that sticky routing relies on the cookie set by Kong for a particular key. This cookie should be read from every request to `/pubkey` and included in the call to `prove/encrypted`. Based on this code it seems to me this is handled correctly?

yarn.lock
```diff
@@ -6424,6 +6424,18 @@ levn@^0.4.1:
     prelude-ls "^1.2.1"
     type-check "~0.4.0"
 
+libsodium-wrappers@^0.8.2:
```
niklaslong, line 6427: The features we're using have been stabilised for a long time but we should make sure we keep the underlying `libsodium` versions reasonably in sync.

niklaslong (approved):
LGTM overall, left a few questions and notes.

sdk/src/security.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @returns {string} the encrypted ProvingRequest in RFC 4648 standard Base64.
+ */
+export function encryptProvingRequest(publicKey: string, provingRequest: ProvingRequest): string {
+    return encryptMessage(publicKey, provingRequest.toBytesLe());
+}
+
+/**
+ * Encrypt a view key with a libsodium cryptobox public key.
+ *
+ * @param {Uint8Array} publicKey The cryptobox X25519 public key to encrypt with (encoded in RFC 4648 standard Base64).
```
marshacb: Looks like the JSDoc says `Uint8Array` but the fn signature uses `string`. Do we intend them to be different?

sdk/tests/program-manager.test.ts
```diff
@@ -160,6 +163,41 @@ describe('Program Manager', async () => {
                 expect(request.feeAuthorization()).equal(undefined);
             }
         });
+
+        it.only('Should execute an encrypted proving request', async function() {
```
marshacb: I see `it.only` is in this test (possibly leftover from debugging). Should we remove it so the full test suite runs?

marshacb (approved):
LGTM. Left a couple of small comments/questions, but nothing blocking 🙂

sdk/src/security.ts
iamalwaysuncomfortable: %100, thanks for the comment, fixed.

sdk/tests/program-manager.test.ts
iamalwaysuncomfortable: Also good call, removed.

yarn.lock
iamalwaysuncomfortable: This is the JS-based library which is the stable JS api for libsodium and the best bet to ensure we're using the well audited libsodium impl for JS. 

I think what you might be suggesting is we should consider running the encryption in Rust the same way we do for the RSS/DPS, but it'd require us setting up our own wasm toolchain to first export the C code to wasm and then also the wrapping Rust code that calls the C code. 

This would be more erroenous than using the official JS library.

sdk/src/network-client.ts
iamalwaysuncomfortable: Indeed, it's handled earlier in the method!

sdk/src/models/provingResponse.ts
iamalwaysuncomfortable: This type basically needs to handle all possible shapes returned from the API (errors and all) there's no a single concrete Rust type that handles all shapes on both proving error and success. I could rename it perhaps.

niklaslong (approved):

marshacb (approved):

sdk/tests/program-manager.test.ts
marshacb: looks like the `it.only` got re-introduced

sdk/src/network-client.ts
```diff
@@ -1677,33 +1745,82 @@ class AleoNetworkClient {
...
+        const isExpired = jwtData && Date.now() >= jwtData.expiration;
         if (!jwtData || isExpired) {
             if (options.apiKey && options.consumerId) {
                 jwtData = await this.refreshJwt(apiKey!, consumerId!);
-                // Update both the class and the options with the new JWT
                 this.jwtData = jwtData;
                 options.jwtData = jwtData;
             } else {
-                throw new Error('JWT or both apiKey and consumerId are required');
+                console.warn('JWT or both apiKey and consumerId are required when using the Provable API');
```
marshacb, line 1833: I see we previously threw an error here. Whats the intended behavior when credentials are missing?

marshacb (commented):
lgtm. Just a test nit and a question

sdk/src/network-client.ts
marshacb: Ahh, I see this is intended for the caller to handle the error themselves.

sdk/src/network-client.ts
```diff
@@ -1677,33 +1745,82 @@ class AleoNetworkClient {
...
             : options.provingRequest;
 
         const apiKey = options.apiKey ?? this.apiKey;
-        const consumerId = options.consumerId ?? this.consumerId;    
-        let jwtData = options.jwtData ?? this.jwtData
+        const consumerId = options.consumerId ?? this.consumerId;
+        let jwtData = options.jwtData ?? this.jwtData;
 
-        // Check if JWT is expired or missing
-        const bufferTime = FIVE_MINUTES; // 5 minutes buffer
```
marshacb, line 1697: Is the removal of the 5-minute JWT buffer intentional?

sdk/tests/program-manager.test.ts
iamalwaysuncomfortable: So it did, changed.

sdk/src/network-client.ts
iamalwaysuncomfortable: Yes, also the authentication method might change this provides flexibility for calling clients without auth in test environments or calling DPS provers with a different authorization format.

PR #1173, opened 2026-01-30
feat(wasm): add isValid static method to Address for validating addre…

marshacb:
## Motivation

Adds a dedicated method to validate addresses without throwing exceptions. Useful for input validation in forms and APIs.

## Test Plan

- Added Rust tests in `wasm/src/account/address.rs` - run with `yarn test` in `wasm/`
- Added TypeScript tests in `sdk/tests/wasm.test.ts` - run with `yarn test` in `sdk/`

Tests cover: valid/invalid address strings, valid/invalid byte arrays, and incorrect input types.

### DIFF
--- a/wasm/src/account/address.rs
+++ b/wasm/src/account/address.rs
@@ -30,7 +30,7 @@ use snarkvm_console::prelude::{FromBits, FromBytes, FromFields, Network, ToBits,
 
 use core::{convert::TryFrom, fmt, ops::Deref, str::FromStr};
 use js_sys::{Array, Uint8Array};
-use wasm_bindgen::{convert::TryFromJsValue, prelude::*};
+use wasm_bindgen::{JsCast, convert::TryFromJsValue, prelude::*};
 
 /// Public address of an Aleo account
 #[wasm_bindgen]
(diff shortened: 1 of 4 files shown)


Roee-87 (approved):
all tests pass -- LGTM.

wasm/src/account/address.rs
```diff
@@ -185,6 +185,28 @@ impl Address {
 
         Ok(Address::from(group))
     }
+
+    /// Check if the input is a valid Aleo address.
+    ///
+    /// @param {string | Uint8Array} address - Either a string representation of an address
+    ///        or a Uint8Array of bytes in little-endian format.
+    /// @returns {boolean} True if the input is a valid address, false otherwise.
+    #[wasm_bindgen(js_name = "isValid")]
```
iamalwaysuncomfortable, line 194: The one thing that should be also be done is add a method to the Account object in the JS SDK that uses this under the hood!

iamalwaysuncomfortable (commented):
@Roee-87 - can you give this a review. We were originally thinking about

wasm/src/account/address.rs
marshacb: Great point! Added a static `isValidAddress` method to the Account object that uses `Address.isValid` under the hood.

iamalwaysuncomfortable (approved):

PR #1182, opened 2026-02-06
[Feature] Productionize Record Scanner Interface

iamalwaysuncomfortable:
## Motivation

The record scanner is currently adding encrypted registration flows to ensure sensitive material is protected before it gets to the scanner enclave, this PR adds encrypted registration flows to the RecordScanner object.

## Test Plan

This PR adds tests to the Record Scanner to ensure these encrypted flows work.

### DIFF
--- a/sdk/src/models/record-scanner/encryptedRecordsResult.ts
+++ b/sdk/src/models/record-scanner/encryptedRecordsResult.ts
@@ -1,0 +1,16 @@
+import type { RecordScannerErrorBody } from "./error.js";
+import type { EncryptedRecord } from "../record-provider/encryptedRecord.js";
+
+export interface EncryptedRecordsSuccess {
+    ok: true;
+    data: EncryptedRecord[];
+}
+
+export interface EncryptedRecordsFailure {
+    ok: false;
+    status: number;
+    error: RecordScannerErrorBody;
+}
+
+export type EncryptedRecordsResult = EncryptedRecordsSuccess | EncryptedRecordsFailure;
+
--- a/sdk/src/models/record-scanner/encryptedRecordsResult.ts
+++ b/sdk/src/models/record-scanner/encryptedRecordsResult.ts
@@ -1,4 +1,4 @@
-import type { RecordScannerErrorBody } from "./error.js";
+import type { RecordScannerFailure } from "./error.js";
 import type { EncryptedRecord } from "../record-provider/encryptedRecord.js";
 
 export interface EncryptedRecordsSuccess {
@@ -6,11 +6,5 @@ export interface EncryptedRecordsSuccess {
     data: EncryptedRecord[];
 }
 
-export interface EncryptedRecordsFailure {
-    ok: false;
-    status: number;
-    error: RecordScannerErrorBody;
-}
-
-export type EncryptedRecordsResult = EncryptedRecordsSuccess | EncryptedRecordsFailure;
+export type EncryptedRecordsResult = EncryptedRecordsSuccess | RecordScannerFailure;
 
--- a/sdk/src/models/record-scanner/error.ts
+++ b/sdk/src/models/record-scanner/error.ts
@@ -37,3 +37,13 @@ export interface RecordScannerErrorBody {
     status: number;
 }
 
+/**
+ * Failure variant shared by record-scanner result types.
+ * Use with a success interface to form a discriminated union (e.g. Success | RecordScannerFailure).
+ */
+export interface RecordScannerFailure {
+    ok: false;
+    status: number;
+    error: RecordScannerErrorBody;
+}
+
--- a/sdk/src/models/record-scanner/error.ts
+++ b/sdk/src/models/record-scanner/error.ts
@@ -15,6 +15,44 @@ export class RecordScannerRequestError extends Error {
     }
 }
 
+/** Error thrown when findCreditsRecord or findCreditsRecords is called but decryption is not enabled on the record scanner. */
+export class DecryptionNotEnabledError extends Error {
+    readonly filter?: OwnedFilter;
+
+    constructor(message: string, filter?: OwnedFilter) {
+        super(message);
+        this.name = "DecryptionNotEnabledError";
+        this.filter = filter;
+        Object.setPrototypeOf(this, DecryptionNotEnabledError.prototype);
+    }
+}
+
+/** Error thrown when findCreditsRecord or findCreditsRecords is called but no view key for the UUID is stored in viewKeys or the account. */
+export class ViewKeyNotStoredError extends Error {
+    readonly uuid?: string;
+    readonly filter?: OwnedFilter;
+
+    constructor(message: string, uuid?: string, filter?: OwnedFilter) {
+        super(message);
+        this.name = "ViewKeyNotStoredError";
+        this.uuid = uuid;
+        this.filter = filter;
+        Object.setPrototypeOf(this, ViewKeyNotStoredError.prototype);
+    }
+}
+
+/** Error thrown when no record matches the supplied search filter (e.g. findCreditsRecord / findCreditsRecords). */
+export class RecordNotFoundError extends Error {
+    readonly filter?: OwnedFilter;
+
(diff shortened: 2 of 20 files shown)


sdk/src/models/record-scanner/error.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/** Error thrown when a record scanner request fails due to an invalid response. */
+export class UUIDError extends Error {
+    readonly uuid?: string;
+    readonly filter?: OwnedFilter;
+
+    constructor(message: string, uuid?: string, filter?: OwnedFilter) {
+        super(message);
+        this.name = "InvalidResponseError";
```
marshacb, line 63: Should we change the name here to "UUIDError" for consistency?

sdk/src/record-scanner.ts
```diff
@@ -52,74 +84,291 @@ type RecordScannerOptions = {
...
+                this.uuid = data.uuid;
+            }
+
+            // Add the view key to the local scanner's view keys if configured to do so.
+            if (this.cacheViewKeysOnRegister) {
+                this.addViewKey(viewKey);
+            }
+            return { ok: true, data };
+        } catch (err) {
+            if (err instanceof RecordScannerRequestError) {
```
marshacb: Nit: I see we use this error pattern a number of times. We could add a method to RecordScanner that could be reused for these cases.

e.g.
```
private handleRequestError(err: unknown): { ok: false; status: number; error: RecordScannerErrorBody } {
    if (err instanceof RecordScannerRequestError) {
        return {
            ok: false,
            status: err.status,
            error: { message: err.message, status: err.status },
        };
    }
    throw err; 
}
```

So then we could just have one line each such as:
```
} catch (err) {
    return this.handleRequestError(err);
}
```

sdk/tests/program-manager.test.ts
```diff
@@ -217,6 +217,9 @@ describe('Program Manager', async () => {
                         },
                     );
 
+                if (!safeResult.ok) {
+                    console.log(safeResult);
```
marshacb, line 221: Are we keeping this log for debugging or should we remove it before merging?

sdk/src/record-scanner.ts
```diff
@@ -151,21 +418,40 @@ class RecordScanner implements RecordProvider {
...
+    /**
+     * Check if tags appear in any record inputs on-chain, indicating that the records they belong to have been spent. This is a safe variant of /records/tags that returns a result instead of throwing on HTTP error.
+     *
+     * *
      * @param {string[]} tags The tags to check.
-     * @returns {Promise<Record<string, boolean>>} Map of Aleo Record tags and whether they appeared in any inputs on chain. If boolean corresponding to the tag has a true value, that Record is considered spent by the Aleo Network.
+     * @returns {Promise<TagsResult>} Map of Aleo Record tags and whether they appeared in any inputs on chain. If a boolean corresponding to the tag has a true value, that Record is considered spent by the Aleo Network.
      */
-    async checkTags(tags: string[]): Promise<Record<string, boolean>> {
+    async tags(tags: string[]): Promise<TagsResult> {
```
marshacb, line 518: I noticed the safe vs throwing method naming isn't consistent (e.g., `encrypted()` vs `encryptedRecords()`, `owned()` vs `findRecords()`). Is the current naming intentional or should we align on a pattern? I was wondering if using suffixes could be a good option (e.g. `registerSafe()`, `findRecordsOrThrow()`, etc) but we could also choose to document them as is.

sdk/src/record-scanner.ts
```diff
@@ -52,74 +84,291 @@ type RecordScannerOptions = {
...
+    }
+
+    /**
+     * Refreshes the JWT by making a POST request to /jwts/{consumer_id}. Used when authentication is required.
      */
-    async setUuid(uuidOrViewKey: Field | ViewKey): Promise<void> {
-        this.uuid = uuidOrViewKey instanceof ViewKey ? this.computeUUID(uuidOrViewKey) : uuidOrViewKey;
+    private async refreshJwt(apiKey: string, consumerId: string): Promise<RecordScannerJWTData> {
+        const response = await post(
+            `https://api.provable.com/jwts/${consumerId}`,
```
marshacb: Is the hardcoded `https://api.provable.com/jwts/` intentional? Safe to assume JWT refresh always will go to the production auth service or do we need to possibly account for different environments?

sdk/src/models/record-scanner/encryptedRecordsResult.ts
```diff
@@ -0,0 +1,16 @@
...
+import type { EncryptedRecord } from "../record-provider/encryptedRecord.js";
+
+export interface EncryptedRecordsSuccess {
+    ok: true;
+    data: EncryptedRecord[];
+}
+
+export interface EncryptedRecordsFailure {
+    ok: false;
+    status: number;
```
marshacb: The new result types seem to use `status: number` but the SDK convention (per `provingResponse.ts` and others) is `status: bigint | number`. Should we align with the existing pattern for consistency or is this difference intentional?

marshacb (commented):
Looking great! Left a few questions/minor suggestions on naming conventions and type consistency. Let me know your thoughts :)

sdk/src/models/record-scanner/error.ts
iamalwaysuncomfortable: Indeed

sdk/src/record-scanner.ts
iamalwaysuncomfortable: I like this, adding it.

sdk/tests/program-manager.test.ts
iamalwaysuncomfortable: I like to keep the debug logs in the tests where they make sense, because it allows to read the CI and understand what the issue is.

sdk/src/record-scanner.ts
iamalwaysuncomfortable: The RecordProvider interface was shipped with the specific names like `encryptedRecords()` and `findRecords()` and given we have increasingly more users I'd like to keep the main `RecordProvider` interface the same. 

The goal with naming them as they're named instead of `safe` as the names is that I want them to match the final suffix of the uri endpoints they're hitting. It's a specific choice, but I think it makes sense.

These are a really good set of points though.

sdk/src/record-scanner.ts
iamalwaysuncomfortable: This is a great callout actually, we should NOT have this hardcoded and instead have it be `${this.url}/jwts/${consumerId}`

sdk/src/models/record-scanner/encryptedRecordsResult.ts
iamalwaysuncomfortable: Indeed, this should be corrected. Likely the bigint in provingResponse.ts should NOT accept bigints.

sdk/tests/record-scanner.test.ts
```diff
@@ -97,26 +114,29 @@ describe("RecordScanner", () => {
...
         const expectedBody = JSON.stringify({ view_key: defaultAccount.viewKey().to_string(), start: 0 });
         expect(body).to.equal(expectedBody);
         
-        expect(registrationResponse.uuid).equal("test-uuid");
-        expect(registrationResponse.job_id).equal("test-job-id");
-        expect(registrationResponse.status).equal("pending");
+        expect(result.ok).to.equal(true);
+        if (result.ok) {
+            expect(result.data.uuid).equal("test-uuid");
+            expect(result.data.job_id).equal("test-job-id");
```
niklaslong: I don't think we still return a job_id in the registration response?

niklaslong (approved):
LGTM, we'll likely need to implement the revoke endpoint as well for completeness but I don't think that's a blocker for this changeset 🚀

sdk/tests/record-scanner.test.ts
iamalwaysuncomfortable: I'l double-tap on this.

PR #1186, opened 2026-02-10
[Fix] RecordPlaintext.owner() and OwnedRecord property names for private records

marshacb:
## Motivation

Private transfer integration tests were failing when attempting to use records created via transfer_public_to_private.

Two bugs were identified:

RecordPlaintext.owner() only handled public owner visibility: 
- Records created by `transfer_public_to_private` store the owner with private visibility. The owner() method returned "Record is not public" instead of extracting the address from the decrypted plaintext.

Property name mismatch in NetworkRecordProvider:
- findCreditsRecords() returned camelCase properties (`recordPlaintext`, `programName`) but `OwnedRecord` type and consumers expected snake_case (`record_plaintext`, `program_name`).

These issues prevented the SDK from finding and using shielded records for private transfers.

## Test Plan

- Added 2 unit tests for private owner extraction
- Verified tests fail without fix, pass with fix
- Ran transfer-private integration test. Confirmed transfer-private test runs completely

### DIFF
--- a/wasm/src/record/record_plaintext.rs
+++ b/wasm/src/record/record_plaintext.rs
@@ -32,6 +32,7 @@ use crate::{
             CurrentNetwork,
             FieldNative,
             IdentifierNative,
+            LiteralNative,
             PlaintextEntryNative,
             PlaintextNative,
             ProgramIDNative,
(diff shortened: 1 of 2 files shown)


wasm/src/record/record_plaintext.rs
```diff
@@ -108,7 +109,11 @@ impl RecordPlaintext {
     pub fn owner(&self) -> Result<Address, String> {
         match self.0.owner() {
             Owner::<CurrentNetwork, PlaintextNative>::Public(owner) => Ok(Address::from(*owner)),
-            _ => Err("Record is not public".to_string()),
+            // For decrypted records, extract the address from the plaintext literal
+            Owner::<CurrentNetwork, PlaintextNative>::Private(plaintext) => match plaintext {
+                PlaintextNative::Literal(LiteralNative::Address(address), _) => Ok(Address::from(*address)),
+                _ => Err("Private owner is not an address literal".to_string()),
+            },
```
iamalwaysuncomfortable, line 116: Thanks this is actually essential for record scanning in general.

iamalwaysuncomfortable (approved):

PR #1191, opened 2026-02-12
[Feature] Enable option to use the FeeMaster when building proving requests.

iamalwaysuncomfortable:
## Motivation

Currently the `buildProvingRequest` function automatically calculates the fee for a function and will always send in a ProvingRequest which requires the user to pay the fee. This PR enables users to specify a useFeeMaster flag when building proving requests that enables them to use the `FeeMaster` and avoid paying fees. 

Behaviorally this simply avoid building a `Fee Authorization` within wasm and sends the request without it prompting the Delegated Proving Service to pay using the feemaster.

Secondly, this PR also deprecates the `baseFee` parameter for ProvingRequests as this fee is now calculated on behalf of the user.

## Test Plan

* JS and Wasm tests that ensure Fee Authorizations are not built when the useFeeMaster flag is set have been added to ensure this behavior is consistent.


Pauan (approved):
LGTM

marshacb (approved):

PR #1196, opened 2026-02-18
[Examples] create leo large fn example

marshacb:
## Motivation

The SDK lacked an end-to-end example demonstrating larger function execution, record lifecycle (create → consume → discover), and integration with DPS and RSS. This PR adds a loyalty program template that exercises all of these: multi-program execution with record inputs/outputs, on-chain mapping reads, record scanning via RSS, and delegated proving via DPS which serve as both an integration test and a reference for developers building with the SDK.

## Test Plan

- [x] `npm run local`  runs full flow with local proving (mint → add points → redeem → use voucher)
- [x] `npm run delegated`  runs full flow with DPS (verifies on-chain record creation and cross-program execution)
- [x] `npm run scanner`  discovers on-chain records via RSS (verifies JWT fetch, record decryption, and parsing)
- [x] CI job `create-leo-app loyalty-program (delegated + scanner)` passes with `ALEO_CONSUMER_ID` and `ALEO_DPS_API_KEY` secrets

Screenshot 2026-02-17 at 9 28 25 PM

---

> [!NOTE]
> **Medium Risk**
> Mostly additive example/template code, but it introduces new CI paths that broadcast transactions and depend on external DPS/RSS services and secrets, which can increase flakiness and maintenance risk.
> 
> **Overview**
> Adds a new end-to-end **loyalty program** example to `create-leo-app`, including a Node.js template (CLI demo) and a React template, both bundling two Leo programs (`loyalty_token.aleo` + `loyalty_rewards.aleo`) to exercise record lifecycles, cross-program execution, and tier/voucher flows.
> 
> The Node template includes a large TypeScript demo (`src/index.ts`) that supports **local proving**, **delegated proving via DPS** (broadcasting txs and polling confirmation), and **record discovery via RSS** (JWT fetch + decrypt/parse), with accompanying `.env.example`, build config, and scripts.
> 
> CI (`.github/workflows/sdk.yml`) is extended with new jobs to run the loyalty program template in `local`, `delegated`, and `scanner` modes using repository secrets for DPS/RSS configuration.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 30f2ccefc0dbf7971f2a80e11f379f350064fc91. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


sdk/src/network-client.ts
```diff
@@ -90,10 +92,14 @@ class AleoNetworkClient {
     jwtData?: JWTData;
     proverUri?: string;
     recordScannerUri?: string;
+    /** API host for program/edition lookups (defaults to https://api.provable.com/v2) */
+    apiHost: string;
```
iamalwaysuncomfortable: Was this to get around the CORS issue as it looks like this *might* end up duplicating what the `host` is doing.

sdk/src/network-client.ts
```diff
@@ -284,6 +305,29 @@ class AleoNetworkClient {
         }
     }
 
+    /**
+     * Fetches data from the Provable API (v2) for program/edition lookups.
+     * Uses apiHost which defaults to https://api.provable.com/v2.
+     *
+     * @param url The API endpoint path
+     */
```
iamalwaysuncomfortable: Basically same question here, is why do we need the separate endpoint.

create-leo-app/template-node-loyalty-program-ts/_gitignore
```diff
@@ -0,0 +1,24 @@
+# Logs
+logs
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+pnpm-debug.log*
+lerna-debug.log*
```
iamalwaysuncomfortable, line 8: Duplicate file perhaps?

create-leo-app/template-node-loyalty-program-ts/loyalty_rewards/src/main.leo
```diff
@@ -0,0 +1,124 @@
...
+        owner: address,
+        voucher_id: field,
+        reward_type: u8,       // 1=discount, 2=freebie, 3=upgrade
+        amount: u64
+    }
+
+    mapping voucher_exists: field => bool;
+    mapping voucher_used: field => bool;
+    mapping redemptions_by_type: field => u64;
+    mapping approved_upgrades: field => bool;
```
iamalwaysuncomfortable, line 31: The mappings and structs should have doc comments.

create-leo-app/template-node-loyalty-program-ts/loyalty_token/src/main.leo
```diff
@@ -0,0 +1,184 @@
...
+    async function finalize_approve_upgrade(checksum_hash: field) {
+        approved_upgrades.set(checksum_hash, true);
+    }
+
+    // Calculate tier: Gold (10000+), Silver (1000-9999), Bronze (0-999)
+    inline calculate_tier(points: u64) -> u8 {
+        return points >= 10000u64 ? 2u8 : (points >= 1000u64 ? 1u8 : 0u8);
+    }
+
+    async transition mint_card(
```
iamalwaysuncomfortable, line 59: Add doc comments here.

create-leo-app/template-node-loyalty-program-ts/loyalty_token/src/main.leo
```diff
@@ -0,0 +1,184 @@
...
+        owner: address,
+        card_id: field,
+        points: u64,
+        tier: u8  // 0=bronze, 1=silver, 2=gold
+    }
+
+    mapping card_exists: field => bool;
+    mapping total_cards: field => u64;
+    mapping total_points_issued: field => u64;
+    mapping approved_upgrades: field => bool;
```
iamalwaysuncomfortable, line 29: Add documentation to the structs and mappings

create-leo-app/template-node-loyalty-program-ts/loyalty_token/src/main.leo
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    record LoyaltyCard {
+        owner: address,
+        card_id: field,
+        points: u64,
+        tier: u8  // 0=bronze, 1=silver, 2=gold
+    }
+
+    mapping card_exists: field => bool;
```
iamalwaysuncomfortable, line 23: We can optionally get rid of this field as each function takes a record and those or gaurunteed to be unique, so if a record is minted, its validity is guarunteed by the chain.

create-leo-app/template-node-loyalty-program-ts/loyalty_token/src/main.leo
```diff
@@ -0,0 +1,184 @@
...
+
+        let total: u64 = Mapping::get_or_use(total_points_issued, 0field, 0u64);
+        Mapping::set(total_points_issued, 0field, total + points);
+    }
+
+    transition check_points(card: LoyaltyCard) -> (LoyaltyCard, u64) {
+        return (card, card.points);
+    }
+
+    async transition transfer_card(
```
iamalwaysuncomfortable, line 122: Generally add doc comment methods to these transitions.

create-leo-app/template-node-loyalty-program-ts/loyalty_token/src/main.leo
```diff
@@ -0,0 +1,184 @@
...
+
+    async function finalize_transfer_card(card_id: field) {
+        let exists: bool = Mapping::get(card_exists, card_id);
+        assert(exists);
+    }
+
+    async transition split_card(
+        card: LoyaltyCard,
+        points_to_keep: u64,
+        new_nonce: field
```
iamalwaysuncomfortable: The `new_nonce` field relies on the owner of the card to put in their own randomness which is potentially game-able, this nonce should probably be based on the hash of the record as the record has its own hidden `_nonce` formation of a record incorporates inherent randomness. This would also allow us to up the constraint count of the function.

This works like so.

```rust
let new_nonce: field = BHP256::hash_to_field(card);
```

create-leo-app/template-node-loyalty-program-ts/loyalty_rewards/src/main.leo
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    record RewardVoucher {
+        owner: address,
+        voucher_id: field,
+        reward_type: u8,       // 1=discount, 2=freebie, 3=upgrade
+        amount: u64
+    }
+
+    mapping voucher_exists: field => bool;
```
iamalwaysuncomfortable, line 25: This field may be redundant. It's used as an existence check but the records themselves guaruntee uniqueness.

create-leo-app/template-node-loyalty-program-ts/loyalty_rewards/src/main.leo
```diff
@@ -0,0 +1,124 @@
...
+    async transition approve_upgrade(checksum: [u8; 32]) -> Future {
+        assert_eq(self.caller, ADMIN);
+        return finalize_approve_upgrade(BHP256::hash_to_field(checksum));
+    }
+
+    async function finalize_approve_upgrade(checksum_hash: field) {
+        approved_upgrades.set(checksum_hash, true);
+    }
+
+    async transition redeem_points_for_voucher(
```
iamalwaysuncomfortable, line 56: transitions should have doc comments.

create-leo-app/template-node-loyalty-program-ts/loyalty_rewards/src/main.leo
```diff
@@ -0,0 +1,124 @@
...
+        };
+
+        return (updated_card, voucher, finalize_redeem_card(voucher_id, reward_type));
+    }
+
+    async function finalize_redeem_card(voucher_id: field, reward_type: u8) {
+        let exists: bool = Mapping::contains(voucher_exists, voucher_id);
+        assert(!exists);
+
+        Mapping::set(voucher_exists, voucher_id, true);
```
iamalwaysuncomfortable, line 85: This is likely redundant due to the record itself being unique.

create-leo-app/template-node-loyalty-program-ts/loyalty_rewards/src/main.leo
```diff
@@ -0,0 +1,124 @@
...
+
+        let voucher_id: field = BHP256::hash_to_field(
+            VoucherInput { card_id: card.card_id, reward_type: reward_type, cost: points_to_spend }
+        );
+
+        let voucher: RewardVoucher = RewardVoucher {
+            owner: card.owner,
+            voucher_id: voucher_id,
+            reward_type: reward_type,
+            amount: points_to_spend / 10u64
```
iamalwaysuncomfortable, line 75: Note that this works under integer division rules (i.e. the remainder is discarded: 99/10 = 9)

create-leo-app/template-node-loyalty-program-ts/src/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+    private readonly REWARDS_PROGRAM_ID = "loyalty_rewards.aleo";
+
+    // Feature configuration
+    private _provingMode: ProvingMode = ProvingMode.Local;
+    private _recordScanner: RecordScanner | null = null;
+    private _dpsUrl?: string;
+    private _dpsApiKey?: string;
+    private _consumerId?: string;
+    private _dpsPrivacy: boolean = false;
+    private _rssPrivacy: boolean = false;
```
iamalwaysuncomfortable, line 232: I like this, we should consider this a broader configuration object for all contracts.

create-leo-app/template-node-loyalty-program-ts/src/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+// Types and Interfaces
+// ============================================================================
+
+/**
+ * Proving mode for transactions
+ */
+export enum ProvingMode {
+    Local = "local",
+    Delegated = "delegated",
+}
```
iamalwaysuncomfortable, line 30: We'll eventually want to move this to the `types` package of the aleo-dev-toolkit.

create-leo-app/template-node-loyalty-program-ts/src/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+ */
+export interface RedeemResult {
+    card: LoyaltyCard;
+    voucher: RewardVoucher;
+}
+
+/**
+ * Configuration for the LoyaltyProgram.
+ */
+export interface LoyaltyProgramConfig {
```
iamalwaysuncomfortable: This is almost general enough to be considered just `ProgramConfig`

```suggestion
export interface ProgramConfig {
```

create-leo-app/template-node-loyalty-program-ts/src/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+    private readonly TX_POLL_INTERVAL_MS = 5000;
+    private readonly TX_TIMEOUT_MS = 300000; // 5 minutes
+
+    /**
+     * Create a new LoyaltyProgram instance.
+     *
+     * @param account - The Aleo account to use for transactions
+     * @param config - Optional configuration for proving mode and record scanner
+     */
+    constructor(account: Account, config?: LoyaltyProgramConfig) {
```
iamalwaysuncomfortable:
```suggestion
    constructor(account: Account, config?: ProgramConfig) {
```

create-leo-app/template-node-loyalty-program-ts/src/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+        if (!registerResponse.ok) {
+            throw new Error(`Failed to register with scanner: ${registerResponse.status}`);
+        }
+
+        const result = await registerResponse.json();
+        console.log(`   ${COLORS.dim}├─ Registered with UUID: ${result.uuid?.slice(0, 20)}...${COLORS.reset}`);
+
+        // Set the UUID on the scanner for subsequent queries.
+        await this._recordScanner.setUuid(this._account.viewKey());
+    }
```
iamalwaysuncomfortable: This can likely delegated directly to the record scanner method?

create-leo-app/template-node-loyalty-program-ts/src/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+     *
+     * @param startHeight - The block height to start scanning from.
+     * @param endHeight - Optional end block height (defaults to latest).
+     * @returns Array of LoyaltyCard records.
+     *
+     * @example
+     * const cards = await loyalty.findMyCards(0);
+     * console.log(`Found ${cards.length} cards`);
+     */
+    async findMyCards(startHeight: number = 0, endHeight?: number): Promise<LoyaltyCard[]> {
```
iamalwaysuncomfortable, line 368: This is an interesting take on how to find the records. I think this works for the purpose of this contract, but we should generally find a way to have a records.fetch function on a contract.

[This is how Puzzle's SDK does it for instance](https://docs.puzzle.online/functions/records), but we could likely do better than that. 

Something like: 
- `contract.records.[recordName]` that gets auto-generated in the build step. 
- `contract.records({recordName: string, filter: RecordFilter})` + `contracts.records()` which emit the records and their schema.

Let's debate in the coming days.

create-leo-app/template-node-loyalty-program-ts/src/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+        if (!this._recordScanner) {
+            throw new Error("No record scanner configured. Use setRecordScanner() to configure.");
+        }
```
iamalwaysuncomfortable, line 371: Should we make this mandatory on any contract? A record finder should likely be necessary.

In the case of using the SDK in node, it'd require either a local record scanner OR a remote record scanner. In the case of the wallet, you'd simply ask the wallet for records (because the wallet manages the concern).

create-leo-app/template-node-loyalty-program-ts/src/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+     * @returns The same card (for chaining) and points value.
+     */
+    async checkPoints(card: LoyaltyCard): Promise<{ card: LoyaltyCard; points: number }> {
+        const inputs = [card.raw];
+
+        const outputs = await this.execute(
+            this.tokenProgram,
+            "check_points",
+            inputs
+        );
```
iamalwaysuncomfortable, line 529: This serves as a good manual impl of the contract abstraction.

create-leo-app/template-node-loyalty-program-ts/src/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+
+        return this.parseVoucher(outputs[0]);
+    }
+
+    // ==========================================================================
+    // Private Execution Methods
+    // ==========================================================================
+
+    private async execute(
```
iamalwaysuncomfortable, line 706: We probably shouldn't deliniate between execute and executeWithImports and unify them.

create-leo-app/template-node-loyalty-program-ts/src/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+            outputs = await this.executeDelegatedWithImports(program, functionName, inputs, imports);
+        } else {
+            outputs = await this.executeLocalWithImports(program, functionName, inputs, imports);
+        }
+
+        logOperationComplete(functionName, Date.now() - start, this._provingMode);
+        return outputs;
+    }
+
+    private async executeLocal(
```
iamalwaysuncomfortable, line 727: I think the deliniation should just be executeLocal and executeDelegated to keep things simple.

create-leo-app/template-node-loyalty-program-ts/src/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+            for (const transition of transaction.execution.transitions) {
+                if (transition.outputs) {
+                    for (const output of transition.outputs) {
+                        if (output.value) {
+                            // Check if this is an encrypted record (starts with "record1").
+                            if (output.type === "record" && output.value.startsWith("record1")) {
+                                // Decrypt the record using the account's view key.
+                                const ciphertext = RecordCiphertext.fromString(output.value);
+                                const plaintext = ciphertext.decrypt(this._account.viewKey());
+                                outputs.push(plaintext.toString());
```
iamalwaysuncomfortable: This will only work when the transaction contains the record in question and the record is owned by the `address` it was sent to.

create-leo-app/template-node-loyalty-program-ts/README.md
```diff
@@ -0,0 +1,127 @@
+# Loyalty Program - Node.js Example
```
iamalwaysuncomfortable, line 1: We should turn this into a tutorial and perhaps as a follow-on PR add a small config script to help with the configuration. This would help us make a prototype of the initial create-leo-app cli.

iamalwaysuncomfortable (commented):
This is excellent work, had a few small nits and some comments about how things might be organized, but this helps really frame how the `create-leo-app` cli and a contract abstraction should work. 

Let's go over this in detail tomorrow.

iamalwaysuncomfortable:
@cursor review

sdk/src/network-client.ts
marshacb: Good catch! I originally added `apiHost` thinking the `program/edition` endpoints had moved exclusively to the v2 API. After validating, both `/program/{id}` and `/program/{id}/latest_edition` return identical results on the explorer v1 API with the same CORS policy. The separation isn't needed so I'll revert to using `host` for program lookups.

create-leo-app/template-node-loyalty-program-ts/_gitignore
marshacb: Good call out. IIUC every template uses `_gitignore` which gets renamed to `.gitignore` when a project is scaffolded. I do see 2 `.gitignore` that seem like duplicates in this case so Ill remove them and add .env to `_gitignore` instead.

create-leo-app/template-node-loyalty-program-ts/loyalty_rewards/src/main.leo
marshacb: Adding doc comments to all structs, records, and mappings across both programs.

create-leo-app/template-node-loyalty-program-ts/loyalty_token/src/main.leo
marshacb: Good point! Since records are unique on-chain this is technically redundant for validity. It does prevent minting duplicate card IDs (same holder/nonce/points hash), which is a minor application-level safeguard. I was considering that without this check there could be unique records that have the same card id (low chance unless the same inputs were intentionally used I assume) which could be confusing at the application layer. This is deployed so not removable IIUC but if we want to redeploy entirely and remove this we can.

create-leo-app/template-node-loyalty-program-ts/loyalty_rewards/src/main.leo
marshacb: Good catch. `voucher_exists` is kind of redundant since record uniqueness is guaranteed by the hidden `_nonce`. Thinking of leaving it as is for now unless you feel strongly about it since we've already deployed and mappings can't be removed per upgrade constraints.

create-leo-app/template-node-loyalty-program-ts/loyalty_rewards/src/main.leo
marshacb: Same as above. The mapping can't be removed due to upgrade constraints. We could stop writing to it in a future upgrade but since the checks are harmless we also could leave them in. Happy to remove and take a different approach if you recommend.

create-leo-app/template-node-loyalty-program-ts/src/index.ts
marshacb: Agreed! Replaced the hand-rolled encrypted registration flow with the SDK's RecordScanner.registerEncrypted(viewKey, startHeight) in both templates and removed the encryptRegistrationRequest import.

create-leo-app/template-node-loyalty-program-ts/src/index.ts
marshacb: Agree that record discovery is essential for production use. I kept it optional here because the local proving flow runs entirely offline via `OfflineQuery`. No transactions are broadcast and no records land on-chain, so there's nothing to scan. Every operation returns its output records directly, so the caller always has them in hand. The scanner is specifically for re-discovering records on-chain, which only matters once you're broadcasting transactions in delegated mode. This makes it so users can layer on scanning when they move to delegated proving and on-chain execution. We could remove local and require scanning if that makes this example more clear/focused though.

create-leo-app/template-node-loyalty-program-ts/src/index.ts
marshacb: Good point! This assumes all record outputs belong to the executing account, which holds true for the self-operations in this demo. For transfers or minting to other addresses, the decrypt would fail since the record is encrypted to the new owner's view key. Should we handle this now or when we build out the contract abstraction layer? (something like try/catch with index aware output mapping).

create-leo-app/template-node-loyalty-program-ts/README.md
marshacb: Love the idea. A tutorial format with step-by-step progression (local proving first, then layering on delegated + scanning) would match how the template is already structured. For the config script, we could start with a simple interactive prompt that generates the .env file by asking for proving mode, DPS/RSS credentials, etc. That would double as a prototype for the create-leo-app CLI's project setup flow. Happy to tackle both as a follow-on PR.

create-leo-app/template-node-loyalty-program-ts/src/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+                                outputs.push(output.value);
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        return outputs;
+    }
```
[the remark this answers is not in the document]
marshacb: Looks like a possible false positive. CI runs this exact path (delegated full_flow including redeemForVoucher) and produces correct results (3100 points, Upgrade voucher with value 200). The inner cross-program transition outputs don't appear as separate entries in the transaction's transitions array the way this analysis assumes.

create-leo-app/template-node-loyalty-program-ts/loyalty_token/src/main.leo
marshacb: Good call! Added `split_card_v2` to cover this. It uses `BHP256::hash_to_field(card)` as the nonce. The original `split_card` is kept as a legacy variant with the caller-provided nonce parameter. `split_card_v2` is the version used in both the node and react templates.

create-leo-app/template-react-loyalty-program-ts/src/workers/worker.ts
```diff
@@ -0,0 +1,1592 @@
...
+      this.emitStatus({
+        type: "error",
+        operation: functionName,
+        error: error.message,
+      });
+      throw error;
+    }
+  }
+
+  private async executeWithImports(
```
iamalwaysuncomfortable: Same thing with this example we probably want to unify in the same way that was done in the node.js example.

iamalwaysuncomfortable (commented):
One final comment on unifying the execute methods in the react example the same way they were with node, after this is done this can go in!

create-leo-app/template-node-loyalty-program-ts/src/index.ts
```diff
@@ -0,0 +1,1279 @@
...
```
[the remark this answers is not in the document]
marshacb: These functions aren't called in any of our scripts and the loyalty templates are example code, not production. Deferring for now.

iamalwaysuncomfortable (approved):

PR #1197, opened 2026-02-18
[Fix](wasm): normalize address casing and reject non-lowercase

marshacb:
## Motivation

Aleo addresses use `bech32m` encoding which requires uniform casing. The network rejects uppercase addresses at transaction submission, but the SDK previously accepted them in `from_string` and `isValid` without any enforcement. This led to confusing failures downstream. This PR normalizes casing in `from_string` (auto-lowercases) and accepts non-lowercase input in `isValid` for better UX.

## Test Plan

- Added wasm-level tests (address.rs) verifying `is_valid` rejects uppercase and mixed case strings, and `from_string` auto-lowercases to the correct address
- Added TypeScript-level tests (wasm.test.ts) covering the same behaviors through the JS bindings
- Existing tests for `from_private_key`, `from_view_key`, `from_program_id`, signature verification, and byte round-trips continue to pass unchanged

### DIFF
--- a/wasm/src/account/address.rs
+++ b/wasm/src/account/address.rs
@@ -188,7 +188,7 @@ impl Address {
     }
 
     /// Check if the input is a valid Aleo address.
-    /// String addresses must be lowercase to be considered valid.
+    /// String addresses are automatically lowercased before validation.
     ///
     /// @param {string | Uint8Array} address - Either a string representation of an address
     ///        or a Uint8Array of bytes in little-endian format.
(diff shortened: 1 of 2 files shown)


wasm/src/account/address.rs
```diff
@@ -195,6 +197,9 @@ impl Address {
     pub fn is_valid(address: JsValue) -> bool {
         // Try parsing as string first
         if let Some(address_str) = address.as_string() {
+            if address_str != address_str.to_lowercase() {
+                return false;
+            }
```
iamalwaysuncomfortable: Let's double-confirm that we want to go forward with UX. If we do, this is good to proceed.

iamalwaysuncomfortable (approved):
This is approved!

PR #1202, opened 2026-02-24
[Feature] Add revoke method to RSS object

iamalwaysuncomfortable:
## Motivation

The RSS includes `revoke` functionality allowing UUIDs to be removed from the RSS, this PR adds this method to the `RecordScanner` object.

The PR also makes a few more minor changes.
1. Revokes any UUIDs it registers.
2. Fixes a small naming error in UUIDError
3. Throws a UUIDError in the `owned` method if the UUID is improperly configured.

## Test Plan

The following tests are added.
* Stub tests for the revoke flow
* Addition of revoke invocations after testing `register` and `registerEncrypted` methods.,

### DIFF
--- a/sdk/tests/record-scanner.test.ts
+++ b/sdk/tests/record-scanner.test.ts
@@ -6,6 +6,7 @@ import {
     DecryptionNotEnabledError,
     RecordNotFoundError,
     RecordScannerRequestError,
+    UUIDError,
     ViewKeyNotStoredError,
 } from "../src/models/record-scanner/error";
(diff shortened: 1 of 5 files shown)


iamalwaysuncomfortable:
@cursor review

niklaslong (approved):

sdk/tests/record-scanner.test.ts
niklaslong, line 893: Should this case also call revoke?

marshacb (approved):
lgtm!

PR #1208, opened 2026-02-26
[Feature] Add dynamic dispatch import resolution across WASM entry points

marshacb:
## Motivation

Programs using `call.dynamic` reference their targets via field-encoded identifiers at runtime, not through static import declarations. The existing `resolve_imports` only loads statically declared imports, so dynamically dispatched targets would never be added to the process which would cause authorization and execution to fail for any program using `call.dynamic`.

This PR:

- Unifies import resolution into a single `resolve_imports_or_builder` dispatcher across all WASM entry points (`execute`, `authorize`, proving request, and deploy paths)
- Introduces `ProgramImportsBuilder`, a zero-cost WASM builder that carries `proving`/`verifying` keys alongside program source, with `programNames()`/`getProgram()` for lightweight enumeration and `toObject()`/`fromObject()` for full round-trip serialization including keys as Uint8Array
- Adds a `KeyStore` interface for persistent key storage, with automatic load-before-execution (`loadKeysFromStore`) and save-after-execution (`persistExtractedKeys`) so the second execution of a program skips key synthesis entirely
- Includes IndexedDBKeyStore, a browser-native KeyStore implementation for the React template
- Handles the new `Literal::Identifier` variant from the latest `snarkVM` rev
### WASM tests (Rust):
- `test_resolve_dynamic_imports` - verifies resolve_imports does NOT load dynamic targets, and `resolve_dynamic_imports` does
- `test_resolve_multiple_dynamic_imports` - multiple dynamic targets loaded from a single imports object
- `test_authorize_dynamic_dispatch` - `process.authorize()` succeeds for a `call.dynamic` program after import resolution
- `test_program_imports_program_names` - lightweight enumeration without key serialization
- `test_program_imports_get_program` - source retrieval for known/unknown programs
- `test_program_imports_add_key_bytes_*` - byte-level key round-trip, validation of missing program, and invalid bytes
- `test_program_imports_to_object_from_object_roundtrip_keys` - full structured format round-trip with real VK bytes

### SDK tests (TypeScript):
- Dynamic dispatch tests - fixture parsing, offline execution with single/multiple dynamic targets, authorization with serialization round-trip, delegated proving path
- KeyStore integration tests -  `buildProgramImports` with static/dynamic imports and network merge, `loadKeysFromStore` positive/negative/error paths, `persistExtractedKeys` with top-level keys, `resolveTopLevelKeys` fallback chain, `synthesizeKeys` auto-persist, `run()` end-to-end with auto-built builder and post-execution persistence

### Manual testing:
- Node template (`template-node-ts`) updated to use `LocalFileKeyStore` with `setKeyStore()` for persistent key caching on disk
- React template (`template-react-ts`) with `IndexedDBKeyStore` - verified keys persist to `IndexedDB` after first execution, second execution logs `Inserting externally provided proving and verifying keys` and skips synthesis

---

> [!NOTE]
> **Medium Risk**
> Touches core WASM `ProgramManager` execution/authorization/deploy/proving-request paths; incorrect import loading or clone semantics could break transaction building for existing consumers.
> 
> **Overview**
> Adds `ProgramManager::resolve_dynamic_imports` to load *non-statically-imported* programs from the user-provided `imports` map, enabling programs that use `call.dynamic` to authorize/execute successfully.
> 
> Wires this dynamic import resolution (and associated `imports.clone()` adjustments) into the WASM `ProgramManager` entry points for authorization, execution (offline/on-chain/devnode/fee estimation), deployment/upgrade, and proving requests.
> 
> Expands e2e and SDK test coverage with simple `call.dynamic` programs (single + multiple targets) and new tests for `ProgramManager.run()`, `buildAuthorization`/`buildAuthorizationUnchecked`, `provingRequest`, plus WASM unit tests validating dynamic import loading behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit af672cd96f081e53f0d1e50f4e685ebd80dd5464. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


marshacb:
@cursor review

wasm/src/programs/manager/mod.rs
```diff
@@ -125,37 +133,69 @@ impl ProgramManager {
...
+            }
+            let value = Reflect::get(imports, &program_id.as_str().into()).ok();
+            if let Some(source) = extract_source(&value) {
+                let import = ProgramNative::from_str(&source).map_err(|err| err.to_string())?;
+                if !process.contains_program(import.id()) {
+                    log(&format!("Importing program: {program_id}"));
+                    Self::resolve_program_imports(process, &import, imports)?;
+                    log(&format!("Adding {program_id} to the process"));
+                    process.add_program_with_edition(&import, 1).map_err(|err| err.to_string())?;
+                }
```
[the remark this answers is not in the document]
marshacb: Pre-existing behavior. The original `if let Some(import_string) = ...as_string()` also silently skipped missing/non-string imports. Worth addressing but may be out of scope for this PR.

wasm/src/programs/manager/mod.rs
```diff
@@ -125,37 +133,69 @@ impl ProgramManager {
...
-        imports: Option<Object>,
-    ) -> Result<(), String> {
+    /// Resolve all imports from the imports object and load them into the process.
+    /// Accepts plain string values (`{ "name.aleo": "source" }`) or objects with a
+    /// `source` property (`{ "name.aleo": { source: "..." } }`).
+    ///
+    /// To provide proving/verifying keys for imported programs, use `ProgramImports`
+    /// (the builder) instead of this legacy Object path.
+    pub(crate) fn resolve_imports(process: &mut ProcessNative, imports: Option<Object>) -> Result<(), String> {
         if let Some(imports) = imports {
```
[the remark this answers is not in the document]
marshacb: Updated PR description. Removed reference to previous `resolve_dynamic_imports` fn

sdk/src/program-manager.ts
```diff
@@ -3188,9 +3223,9 @@ class ProgramManager {
         const programImports = imports ? imports : await this.networkClient.getProgramImports(programSource);
         console.log(JSON.stringify(programImports));
         if (Object.keys(programImports)) {
-            return WasmProgramManager.estimateFeeForAuthorization(authorization, programSource, programImports, edition);
+            return WasmProgramManager.estimateFeeForAuthorization(authorization, programSource, programImports, edition, programImportsBuilder);
         }
-        return WasmProgramManager.estimateFeeForAuthorization(authorization, programSource, imports, edition);
+        return WasmProgramManager.estimateFeeForAuthorization(authorization, programSource, imports, edition, programImportsBuilder);
```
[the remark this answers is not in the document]
marshacb: Removed the console.log and changed Object.keys(programImports) to Object.keys(programImports).length > 0 in both occurrences.

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,275 @@
...
+            for (fn_name, vk) in &entry.verifying_keys {
+                let fn_id = IdentifierNative::from_str(fn_name).map_err(|e| e.to_string())?;
+                let has_vk = process.get_stack(&program_id).map_or(false, |stack| stack.contains_verifying_key(&fn_id));
+                if has_vk {
+                    log(&format!("Verifying key already exists for {name}/{fn_name}, skipping"));
+                    continue;
+                }
+                log(&format!("Inserting verifying key for {name}/{fn_name}"));
+                process.insert_verifying_key(&program_id, &fn_id, vk.clone()).map_err(|e| e.to_string())?;
+            }
```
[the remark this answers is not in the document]
marshacb: Added a `process.contains_program(&program_id)` guard at the top of phase 2's loop to skip key insertion when a corresponding program isnt in the process already.

wasm/src/programs/manager/mod.rs
```diff
@@ -125,37 +133,69 @@ impl ProgramManager {
...
-                            process.add_program_with_edition(&import, 1).map_err(|err| err.to_string())?;
-                        }
+            let keys = Object::keys(&imports);
+            for i in 0..keys.length() {
+                let key = keys.get(i).as_string().unwrap_or_default();
+                if key == "credits.aleo" {
+                    continue;
+                }
+                let value = Reflect::get(&imports, &key.as_str().into()).ok();
+                if let Some(source) = extract_source(&value) {
```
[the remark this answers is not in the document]
marshacb: Changed to `let Some(key) = keys.get(i).as_string() else { continue; };` to skip non-string keys explicitly instead of defaulting to empty string.

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,279 @@
...
+
+use super::ProgramManager;
+
+/// Internal storage for a single imported program's source and optional keys.
+#[derive(Clone)]
+struct ProgramEntry {
+    source: String,
+    proving_keys: HashMap<String, ProvingKeyNative>,
+    verifying_keys: HashMap<String, VerifyingKeyNative>,
+}
```
iamalwaysuncomfortable, line 58:
```suggestion
/// Internal storage for a single imported program's source and optional keys.
#[derive(Clone)]
struct ProgramEntry {
    program: ProgramNative,
    edition: U16Native,
    proving_keys: HashMap<IdentifierNative, ProvingKeyNative>,
    verifying_keys: HashMap<IdentifierNative, VerifyingKeyNative>,
}

impl ProgramEntry {
    pub fn id(&self) -> &ProgramIDNative {
        self.program.id()
    }
    
    // Get the program stored.
    pub fn program(&self) -> &Program
}
```

Couple suggested modifications. Since this is an internal object, we probably want to keep things in the Rust representation so we don't have to serialize back and forth too often.

Also:
* We probably want to keep the serialized object as the `Program` object so it's confirmed as valid and we can get its data so anything inserted can be verified.
* We should store the edition because multiple editions will each have their own pks/vks.

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,279 @@
...
+/// ```js
+/// const imports = new ProgramImports();
+/// imports.addProgram("my_program.aleo", programSource);
+/// imports.addProvingKey("my_program.aleo", "my_function", provingKey);
+/// imports.addVerifyingKey("my_program.aleo", "my_function", verifyingKey);
+/// ```
+#[wasm_bindgen]
+#[derive(Clone)]
+pub struct ProgramImports {
+    entries: HashMap<String, ProgramEntry>,
```
iamalwaysuncomfortable:
```suggestion
    entries: HashMap<ProgramIDNative, ProgramEntry>,
```

We want to store the ProgramID here instead I think.

sdk/src/models/imports.ts
```diff
@@ -6,4 +6,4 @@ interface ImportedPrograms {
     [key: string]: string; // This allows for arbitrary keys with any type values
 }
 
-export { ImportedVerifyingKeys, ImportedPrograms }
\ No newline at end of file
+export { ImportedVerifyingKeys, ImportedPrograms }
```
iamalwaysuncomfortable, line 9: Here we might consider creating a new typescript interface for imports, while maintaining the legacy imports interface under a deprecation warning that enables people to pass in proving keys and verifying keys.


```suggestion
interface Import {
   program: string,
   provingKey: Uint8Array?,
   verifyingKey: Uint8Array?,
}

interface Imports {
   [key: string]: Import
}

export { Import, Imports, ImportedVerifyingKeys, ImportedPrograms }
```

sdk/src/program-manager.ts
```diff
@@ -95,6 +109,7 @@ interface ExecuteOptions {
     program?: string | Program;
     imports?: ProgramImports;
     edition?: number,
+    programImportsBuilder?: ProgramImportsBuilder;
```
iamalwaysuncomfortable, line 116: We could allow this as an option parameter for flexibility, but MOSTLY construct keys inside the ProgramManager functions.

sdk/src/program-manager.ts
```diff
@@ -3179,18 +3213,18 @@ class ProgramManager {
...
         if (!authorization) {
             throw new Error("Authorization must be provided if estimating fee for Authorization.")
         }
         const programSource = program ? program.toString() : await this.networkClient.getProgram(programName, edition);
         const programImports = imports ? imports : await this.networkClient.getProgramImports(programSource);
-        console.log(JSON.stringify(programImports));
-        if (Object.keys(programImports)) {
-            return WasmProgramManager.estimateFeeForAuthorization(authorization, programSource, programImports, edition);
+        if (Object.keys(programImports).length > 0) {
+            return WasmProgramManager.estimateFeeForAuthorization(authorization, programSource, programImports, edition, programImportsBuilder);
```
iamalwaysuncomfortable, line 3485: So instead of only relying on outside callers having to construct the ProgramImportsBuilder, we should mostly handle this inside of the methods themselves.

The source for the proving keys and verifying keys should either be:
- The KeyProvider
- Any imports object people pass in from outside (per the interface shown above)

wasm/src/programs/manager/authorize.rs
```diff
@@ -90,14 +95,15 @@ impl ProgramManager {
...
+        program_imports: Option<ProgramImports>,
     ) -> Result<Authorization, String> {
         let mut process_native = ProcessNative::load_web().map_err(|err| err.to_string())?;
         let process = &mut process_native;
 
         log("Check program imports are valid and add them to the process");
         let program_native = ProgramNative::from_str(program).map_err(|e| e.to_string())?;
         log(&format!("Creating proving request for {}:{function_name}", program_native.id()));
-        ProgramManager::resolve_imports(process, &program_native, imports)?;
+        ProgramManager::resolve_imports_or_builder(process, imports, program_imports.as_ref())?;
```
iamalwaysuncomfortable, line 106: We may consider accepting only the builder here and construct that object ourselves from the outside.

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// ```js
+    /// { "my_program.aleo": "program source..." }
+    /// // or
+    /// { "my_program.aleo": { source: "program source..." } }
+    /// ```
+    ///
+    /// @param {Object} object A plain JavaScript object mapping program names to source code.
+    /// @returns {ProgramImports}
+    #[wasm_bindgen(js_name = "fromObject")]
+    pub fn from_object(object: Object) -> Self {
```
iamalwaysuncomfortable, line 120: Here we should check to see if the object also has the Proving and Verifying keys as well, via the new suggested object shown above.

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// @param {ProvingKey} key The proving key.
+    #[wasm_bindgen(js_name = "addProvingKey")]
+    pub fn add_proving_key(&mut self, program_name: &str, identifier: &str, key: ProvingKey) {
+        let entry = self.entries.entry(program_name.to_string()).or_insert_with(|| ProgramEntry {
+            source: String::new(),
+            proving_keys: HashMap::new(),
+            verifying_keys: HashMap::new(),
+        });
+        entry.proving_keys.insert(identifier.to_string(), ProvingKeyNative::from(key));
+    }
```
iamalwaysuncomfortable: We should add an add_proving_key_bytes method which attempts to construct a proving key from a bytes array here.

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// @param {VerifyingKey} key The verifying key.
+    #[wasm_bindgen(js_name = "addVerifyingKey")]
+    pub fn add_verifying_key(&mut self, program_name: &str, identifier: &str, key: VerifyingKey) {
+        let entry = self.entries.entry(program_name.to_string()).or_insert_with(|| ProgramEntry {
+            source: String::new(),
+            proving_keys: HashMap::new(),
+            verifying_keys: HashMap::new(),
+        });
+        entry.verifying_keys.insert(identifier.to_string(), VerifyingKeyNative::from(key));
+    }
```
iamalwaysuncomfortable, line 204: Same here, we should add an `add_verifying_key_bytes` method here.

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// The key is transferred directly from the WASM `ProvingKey` type with no
+    /// serialization overhead.
+    ///
+    /// @param {string} program_name The program name (e.g., "my_program.aleo").
+    /// @param {string} identifier The function name or record name the key belongs to.
+    /// @param {ProvingKey} key The proving key.
+    #[wasm_bindgen(js_name = "addProvingKey")]
+    pub fn add_proving_key(&mut self, program_name: &str, identifier: &str, key: ProvingKey) {
+        let entry = self.entries.entry(program_name.to_string()).or_insert_with(|| ProgramEntry {
+            source: String::new(),
```
iamalwaysuncomfortable: This should probably error if the sources aren't yet added

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,279 @@
...
+    #[wasm_bindgen(js_name = "toObject")]
+    pub fn to_object(&self) -> Object {
+        let obj = Object::new();
+        for (name, entry) in &self.entries {
+            if !entry.source.is_empty() {
+                Reflect::set(&obj, &name.as_str().into(), &entry.source.as_str().into()).unwrap();
+            }
+        }
+        obj
+    }
```
iamalwaysuncomfortable, line 353: We should output the proving and verifying keys here as bytes as well.

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    /// Check whether any programs have been added to this builder.
+    ///
+    /// @returns {boolean}
+    #[wasm_bindgen(js_name = "hasPrograms")]
+    pub fn has_programs(&self) -> bool {
+        !self.entries.is_empty()
+    }
```
iamalwaysuncomfortable, line 386: We should stick the rust naming here I think.

```suggestion
    /// Check whether any programs have been added to this builder.
    ///
    /// @returns {boolean}
    #[wasm_bindgen(js_name = "isEmpty")]
    pub fn is_empty(&self) -> bool {
        !self.entries.is_empty()
    }
```

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    /// Check whether a specific program has been added.
+    ///
+    /// @param {string} name The program name.
+    /// @returns {boolean}
+    #[wasm_bindgen(js_name = "hasProgram")]
+    pub fn has_program(&self, name: &str) -> bool {
+        self.entries.contains_key(name)
+    }
+}
```
iamalwaysuncomfortable, line 396: Similarly here

```suggestion
    /// Check whether a specific program has been added.
    ///
    /// @param {string} name The program name.
    /// @returns {boolean}
    #[wasm_bindgen(js_name = "contains")]
    pub fn contains(&self, name: &str) -> bool {
        self.entries.contains_key(name)
    }
}
```

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,279 @@
...
+/// (`{ source: "source code", ... }`).
+pub(crate) fn extract_source(value: &Option<wasm_bindgen::JsValue>) -> Option<String> {
+    let value = value.as_ref()?;
+    // Plain string format: "source code"
+    if let Some(source) = value.as_string() {
+        return Some(source);
+    }
+    // Structured object format: { source: "source code", ... }
+    Reflect::get(value, &"source".into()).ok().and_then(|v| v.as_string())
+}
```
iamalwaysuncomfortable: Rename to program just to stick with existing conventions

```suggestion
/// Extract source code from a JS value (plain string or object with `.program`).
///
/// Supports both legacy format (`"source code"`) and structured format
/// (`{ program: "source code", ... }`).
pub(crate) fn extract_source(value: &Option<wasm_bindgen::JsValue>) -> Option<String> {
    let value = value.as_ref()?;
    // Plain string format: "source code"
    if let Some(program) = value.as_string() {
        return Some(program);
    }
    // Structured object format: { program: "source code", ... }
    Reflect::get(value, &"source".into()).ok().and_then(|v| v.as_string())
}
```

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,279 @@
...
+// Internal methods (not exported to JS).
+impl ProgramImports {
+    /// Resolve all programs and their keys into the given process.
+    ///
+    /// This performs the following steps:
+    /// 1. Iterates all entries and loads program source code into the process
+    ///    (respecting transitive static imports via depth-first resolution).
+    /// 2. Inserts any pre-provided proving and verifying keys directly into
+    ///    the process, avoiding expensive on-demand key synthesis.
+    pub(crate) fn resolve_into(&self, process: &mut ProcessNative) -> Result<(), String> {
```
iamalwaysuncomfortable: Maybe we rename this as load_programs?

```suggestion
    pub(crate) fn load_programs(&self, process: &mut ProcessNative) -> Result<(), String> {
```

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,279 @@
...
+        // Phase 1: Load all programs into the process.
+        for (name, entry) in &self.entries {
+            if name == "credits.aleo" || entry.source.is_empty() {
+                continue;
+            }
+            let program = ProgramNative::from_str(&entry.source).map_err(|e| e.to_string())?;
+            if !process.contains_program(program.id()) {
+                log(&format!("Importing program: {name}"));
+                // Resolve transitive static imports first.
+                self.resolve_program_imports(process, &program)?;
```
iamalwaysuncomfortable: It looks like this method might skip adding the program keys, perhaps we want to ensure that when we're calling this, any keys stored are also added?

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,279 @@
...
+                log(&format!("Inserting verifying key for {name}/{fn_name}"));
+                process.insert_verifying_key(&program_id, &fn_id, vk.clone()).map_err(|e| e.to_string())?;
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Recursively resolve a program's static imports in depth-first order.
+    fn resolve_program_imports(&self, process: &mut ProcessNative, program: &ProgramNative) -> Result<(), String> {
```
iamalwaysuncomfortable, line 504: We might want to add any program proving and verifying keys here or consider just unifying this method with the method that calls it, because I believe as is, if programs get added here, it wouldn't add any of the proving or verifying keys.

iamalwaysuncomfortable (changes requested):
This is a great first start, but I believe we need to add a few more things here.

Namely:
1. We want to add an option at the top level for people to send in program imports that include the proving and verifying keys into the top level of calls of the program manager.
2. We want to add the ability for keys to be resolved via any object implementing the `KeyProvider` interface so if keys have been stored, it can retrieve them prior to any call.
3. In resolve_program imports recursion we probably want to ensure we're adding in any proving and verifying keys stored.

sdk/src/program-manager.ts
```diff
@@ -199,6 +227,7 @@ interface ProvingRequestOptions {
     unchecked?: boolean;
     edition?: number;
     useFeeMaster?: boolean;
+    programImportsBuilder?: ProgramImportsBuilder;
```
```suggestion
```

iamalwaysuncomfortable:
Successfully merged in subsequent PRs branched from this one, so closing out former work. Thanks for the contributions as always @marshacb

PR #1216, opened 2026-03-02
[Feature] Add snarkVerify and snarkVerifyBatch for standalone SNARK proof verification

marshacb:
## Motivation

The `snark.verify` opcode is being added to `snarkVM` alongside `dynamic dispatch`, enabling Leo programs to verify SNARK proofs from programs that are not deployed on-chain. This PR exposes the same capability to the SDK, allowing client-side proof verification via `snarkVerify` and `snarkVerifyBatch`. These functions take a verifying key, public inputs, and a proof, and return whether the proof is valid, matching the opcode's interface for JS/TS consumers.
## Test Plan

- [x] Rust unit tests (wasm-pack test --node): `Proof` type roundtrip (string and bytes), native `VerifyingKey::verify` with valid/invalid inputs, proof fixture roundtrip
- [x] TypeScript SDK tests (npm test): `Proof.fromString/toBytes` roundtrip, `snarkVerify` with valid inputs returns true, `snarkVerify` with wrong inputs returns false, `snarkVerifyBatch` with valid batch returns true, `snarkVerifyBatch` with mismatched key/input counts throws
- [x] All tests run on both testnet and mainnet network targets (fixtures are network-agnostic)
- [x] create-leo-app templates (`template-node-ts`, `template-react-ts`) updated with working examples

---

> [!NOTE]
> **Medium Risk**
> Introduces new WASM-exposed SNARK verification primitives and new `ProgramManager` verification entrypoints; while mostly additive, it touches cryptographic verification surfaces and input parsing/field conversion paths.
> 
> **Overview**
> Adds standalone SNARK proof verification support to the SDK by exposing WASM functions `snarkVerify`/`snarkVerifyBatch` plus a new `Proof` type, and re-exporting them through the browser SDK surface.
> 
> Extends `ProgramManager` with `verifyProof` and `verifyBatchProof` (and exported `VerificationOptions` types), including automatic conversion of Aleo-typed public inputs to field elements via `Plaintext.toFields()`.
> 
> Updates create-leo-app Node/React templates to demonstrate proof verification (including a React UI + worker method), and adds SDK/wasm tests and fixtures to validate valid/invalid and batch verification behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 369140c84af5c90424c4a39d9eda2477c7037df2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


marshacb:
@cursor review

wasm/src/programs/execution.rs
```diff
@@ -195,3 +196,143 @@ pub fn verify_function_execution(
...
+/// @param {Array<string>} inputs Array of field element strings representing public inputs (e.g. ["1field", "2field"])
+/// @param {Proof} proof The proof to verify
+/// @returns {boolean} True if the proof is valid, false otherwise
+#[wasm_bindgen(js_name = "snarkVerify")]
+pub fn snark_verify(verifying_key: &VerifyingKey, inputs: Array, proof: &Proof) -> Result<bool, String> {
+    // Parse the input field elements from strings.
+    let raw_inputs = parse_field_inputs(&inputs)?;
+
+    // Verify the proof using VarunaVersion::V2 (current standard).
+    let is_valid = verifying_key.verify("snark_verify", VarunaVersion::V2, &raw_inputs, proof);
```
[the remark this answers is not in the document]
marshacb: The `function_name` param is unused in snarkVM (_function_name). Primarily diagnostic and doesn't affect verification. No change needed.

wasm/src/programs/proof.rs
```diff
@@ -0,0 +1,107 @@
...
+
+impl From<ProofNative> for Proof {
+    fn from(proof: ProofNative) -> Proof {
+        Proof(proof)
+    }
+}
+
+impl PartialEq for Proof {
+    fn eq(&self, other: &Self) -> bool {
+        *self.0 == *other.0
```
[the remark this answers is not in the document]
marshacb: This is functionally correct. `*self.0` and `self.0` are equivalent here since `ProofNative` derives `PartialEq`. This matches the existing convention in `VerifyingKey` and `ProvingKey`. No change needed.

sdk/src/program-manager.ts
```diff
@@ -3086,6 +3089,64 @@ class ProgramManager {
...
+     * const isValid = programManager.verifyProof(verifyingKey, ["1field", "2field"], proof);
+     */
+    verifyProof(verifyingKey: VerifyingKey, inputs: string[], proof: Proof): boolean {
+        try {
+            return snarkVerify(verifyingKey, inputs, proof);
+        } catch (e) {
+            console.warn(`Proof verification failed: ${e}`);
+            return false;
+        }
+    }
```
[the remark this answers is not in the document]
marshacb: Removed the try/catch. Input errors (e.g. malformed field strings) now propagate to the caller instead of being silently returned as false. Invalid proofs still return false since VerifyingKey::verify returns a bool, not a Result

create-leo-app/template-node-ts/src/index.ts
```diff
@@ -73,6 +83,33 @@ async function localProgramExecution(program: string, programName: string, aleoF
     }
 }
 
+// Demonstrate standalone SNARK proof verification. This verifies a proof from a circuit that is
+// not necessarily an Aleo program execution. Useful for verifying proofs received from external
+// sources (e.g. another party proved a computation and you want to verify it).
+function snarkProofVerification() {
```
iamalwaysuncomfortable, line 89: Can we make a version of this for the react example that allows people to put their own proof and verifying key in?

sdk/src/program-manager.ts
```diff
@@ -3086,6 +3089,54 @@ class ProgramManager {
         }
     }
 
+    /**
+     * Verify a SNARK proof against a verifying key and public inputs.
```
```suggestion
     * Verify an Aleo zkSnark proof against a verifying key and public inputs.
```

wasm/src/programs/execution.rs
```diff
@@ -195,3 +196,143 @@ pub fn verify_function_execution(
...
+///
+/// @param {Array<string>} verifying_keys Array of verifying key strings, one per circuit
+/// @param {Array<Array<Array<string>>>} inputs 3D array of field element strings [circuit_idx][instance_idx][field_idx]
+/// @param {Proof} proof The batch proof to verify
+/// @returns {boolean} True if the batch proof is valid, false otherwise
+#[wasm_bindgen(js_name = "snarkVerifyBatch")]
+pub fn snark_verify_batch(verifying_keys: Array, inputs: Array, proof: &Proof) -> Result<bool, String> {
+    if verifying_keys.length() != inputs.length() {
+        return Err(format!(
+            "Mismatch: {} verifying keys but {} input groups provided",
```
iamalwaysuncomfortable:
```suggestion
            "Mismatch: {} verifying keys but {} input groups provided. # of input groups must match # of verifying keys.",
```

sdk/src/program-manager.ts
```diff
@@ -3086,6 +3089,54 @@ class ProgramManager {
...
+     *
+     * This verifies a proof produced by an Aleo program that may not be deployed on chain.
+     * It directly invokes the Varuna proof verification from snarkVM.
+     *
+     * @param {VerifyingKey} verifyingKey The verifying key for the circuit
+     * @param {string[]} inputs Array of field element strings representing public inputs (e.g. ["1field", "2field"])
```
iamalwaysuncomfortable: So although we DO want to support fields, we also want to support passing the string representations of the types here in the SDK instead of the field representations, then internally we can call toFields() on the inputs. 

So this means the string representation of normal aleo types like:
- ["1u32", "2u32"]
- ["{ is_open: true, commision: 20u8}"]

We can then parse to see if they're correct aleo types and call `to_fields()` internally. Users should have the option to pass both the Field representation of the inputs AND the serialized aleo type representations. We can do this via a parameter object.

sdk/src/program-manager.ts
```diff
@@ -3086,6 +3089,54 @@ class ProgramManager {
...
+     *
+     * @example
+     * import { ProgramManager, Proof, VerifyingKey } from "@provablehq/sdk/mainnet.js";
+     *
+     * const programManager = new ProgramManager();
+     * const verifyingKey = VerifyingKey.fromString("verifier1...");
+     * const proof = Proof.fromString("proof1...");
+     * const isValid = programManager.verifyProof(verifyingKey, ["1field", "2field"], proof);
+     */
+    verifyProof(verifyingKey: VerifyingKey, inputs: string[], proof: Proof): boolean {
```
iamalwaysuncomfortable: We want to move to object parameters in general in new JS methods, so here we'd do this.
 
```suggestion
    verifyProof(VerificationOptions): boolean {
```

sdk/src/program-manager.ts
```diff
@@ -3086,6 +3089,54 @@ class ProgramManager {
...
+        return snarkVerify(verifyingKey, inputs, proof);
+    }
+
+    /**
+     * Verify a batch SNARK proof against multiple verifying keys and their corresponding public inputs.
+     *
+     * Each verifying key is paired with one or more sets of public inputs (instances).
+     *
+     * @param {string[]} verifyingKeys Array of verifying key strings, one per circuit
+     * @param {string[][][]} inputs 3D array of field element strings [circuit_idx][instance_idx][field_idx]
```
iamalwaysuncomfortable: Same comment here, we want users to be able to put it in the plaintext string representation of the inputs.

sdk/src/program-manager.ts
iamalwaysuncomfortable: > We can then parse to see if they're correct aleo types and call to_fields() internally. Users should have the option to pass both the Field representation of the inputs AND the serialized aleo type representations. We can do this via a parameter object.

Technically we could do it in this JS side because we have the `toFields()` tooling to do it here.

wasm/src/programs/execution.rs
```diff
@@ -195,3 +196,143 @@ pub fn verify_function_execution(
...
+///
+/// This function verifies a batch proof produced by Aleo programs that may not be deployed on chain.
+/// Each verifying key is paired with one or more sets of public inputs (instances).
+///
+/// @param {Array<string>} verifying_keys Array of verifying key strings, one per circuit
+/// @param {Array<Array<Array<string>>>} inputs 3D array of field element strings [circuit_idx][instance_idx][field_idx]
+/// @param {Proof} proof The batch proof to verify
+/// @returns {boolean} True if the batch proof is valid, false otherwise
+#[wasm_bindgen(js_name = "snarkVerifyBatch")]
+pub fn snark_verify_batch(verifying_keys: Array, inputs: Array, proof: &Proof) -> Result<bool, String> {
```
iamalwaysuncomfortable, line 230: We might consider giving the user the option to give either field inputs OR the string serialization.

sdk/src/program-manager.ts
```diff
@@ -3086,6 +3089,54 @@ class ProgramManager {
...
 
+    /**
+     * Verify a SNARK proof against a verifying key and public inputs.
+     *
+     * This verifies a proof produced by an Aleo program that may not be deployed on chain.
+     * It directly invokes the Varuna proof verification from snarkVM.
+     *
+     * @param {VerifyingKey} verifyingKey The verifying key for the circuit
+     * @param {string[]} inputs Array of field element strings representing public inputs (e.g. ["1field", "2field"])
+     * @param {Proof} proof The proof to verify
```
iamalwaysuncomfortable: This should likely just be the proof string and not the `wasm` object.
```suggestion
     * @param {string} proof The proof to verify
```

wasm/src/programs/proof.rs
```diff
@@ -0,0 +1,107 @@
...
+use crate::types::native::ProofNative;
+use snarkvm_wasm::utilities::{FromBytes, ToBytes};
+
+use std::{ops::Deref, str::FromStr};
+use wasm_bindgen::prelude::wasm_bindgen;
+
+/// SNARK proof for verification of program execution
+#[wasm_bindgen]
+#[derive(Clone, Debug)]
+pub struct Proof(ProofNative);
```
iamalwaysuncomfortable, line 26: I don't think we need to export this object as it doesn't add much. The main reason for exporting an object to wasm is when it provides computations done in Rust that would be hard/error-prone/hard-to-maintain if re-implemented in JS/TS.  In methods that use proofs we can just construct the Rust object via the string representation.

iamalwaysuncomfortable (changes requested):
This is great! A couple of nits + a few main change requests:
* On the JS side, users should be able to specify either
   * The field representation of the inputs (as this is what leo/aleo functions take to do snark.verify) for people who know how to handle this.
   * However we should also be able accept the string serialization representation of the inputs (i.e. ["1u32", "2u32"]) and perform the conversion on behalf of the user for users who don't know how to handle it.
* We likely don't need the proof object exported.
* The create-leo-app example should allow people to paste in their own proofs and inputs.

Thanks for putting this together, I'd say this is almost over the line.

create-leo-app/template-node-ts/src/index.ts
marshacb: Great idea! Updated the React template to accept user provided verifying key, proof, and public inputs via text fields. Pre-filled with sample fixtures so it works out of the box.

wasm/src/programs/execution.rs
marshacb: Handled at the SDK layer.` ProgramManager.verifyProof()` and `verifyBatchProof()` will now both run `inputsToFields()` which accepts either raw field strings or Aleo type strings (e.g. "1u32", "true", structs) and automatically converts them to fields via `Plaintext.toFields()`. The wasm functions are kept as thin wrappers over the snarkVM native API, while ProgramManager provides the user-friendly interface.

sdk/src/program-manager.ts
marshacb: Good call! Updated to accept proof and verifyingKey as plain strings. The methods now handle `Proof.fromString()` / `VerifyingKey.fromString()` conversion internally so users don't need to import or construct wasm objects.

wasm/src/programs/proof.rs
marshacb: Agreed! Updated to accept proof: string (and verifyingKey: string). The ProgramManager methods handle Proof.fromString() / VerifyingKey.fromString() internally, so users never need to import or construct the Proof wasm object.

iamalwaysuncomfortable:
This feature has been merged via #1263 therefore closing this!

Issue #1195, opened 2026-02-16
[Feature] Expose helper to compute publicly computable function inputs


vicsn:
## 🚀 Feature

An MPC wallet building on Aleo needs an end-user to be able to supply publicly computable inputs for the internal `Request::sign` function.

In terrible pseudocode, the function signature of this new exposed SDK utility could be:
`function compute_public_message_payload(program_id: String, function_name: String, inputs: [ValueType<N>]) -> (Field<N>, Field<N>, Field<N>, [(Field<N>, [Field<N>])])`

Where the output consists of:
- `function_id: Field<N>` // as computed by snarkVM's `compute_function_id`
- `is_root: Field<N>` // whether the transition is a root transition or not, serialized as field. Can hardcode to `true` for now.
- `program_checksum: Option<Field<N>>` // only needed if the deployed program contains a `constructor`
- For each input to the VM function call:
  - `index: Field<N>` // The u16 index of the input, converted to a field
  - `serialized_input_as_fields: [Field<N>]` // as computed by snarkVM's `input.to_fields()`

iamalwaysuncomfortable:
@vicsn This wouldn't be so difficult. 

In terms of organization, generally what we're moving towards in regards to vendor or wallet-specific needs like this is putting these in the `integrations` directory and providing an integrations guide, so this would be a good candidate for this feature.

We'll do this on our end, but would be keen to discuss about providing an integrations guide for community/ecosystem tools and work.

iamalwaysuncomfortable:
@marshacb I asked copilot to implement this as an experiment, but if it doesn't do a faithful job, could you implement this?

iamalwaysuncomfortable:
@vicsn this is underway, can you confirm that what's requested in the issue above is exactly what's needed by the potential consumers of this?

vicsn:
> @vicsn this is underway, can you confirm that what's requested in the issue above is exactly what's needed by the potential consumers of this?

Thank you! I'll pass the PR on, I have a feeling the partners need to test first before we can land on the final design

PR #1217, opened 2026-03-03
[Feature] Compute MPC inputs for computing a multi-party Request::sign computation

Roee-87:
## Motivation

This PR implements the feature requested in issue [1195](https://github.com/ProvableHQ/sdk/issues/1195).

## Test Plan

Unit test added to SDK test subdirectory.


sdk/src/program-manager.ts
```diff
@@ -3713,6 +3728,24 @@ class ProgramManager {
             imports,
         );
     }
+
+    async signMessage(options: PublicMessagePayload): Promise<[
```
iamalwaysuncomfortable: This method isn't actually signing anything, this probably should be renamed computeMPCPayload

```suggestion
    async computeMPCPayload(options: PublicMessagePayload): Promise<[
```

sdk/src/program-manager.ts
```diff
@@ -3713,6 +3728,24 @@ class ProgramManager {
...
+        Field,
+        Field | null,
+        Array<[Field, Field[]]>
+        ]> {
+        const { programName, functionName, inputs, isRoot, checksum} = options;
+        if (!this.account) {
+            throw new Error("No account set in ProgramManager. Please set an account to sign messages.");
+        }
+        try {
+            return ExecutionRequest.computePublicMessagePayload(programName, functionName, inputs, isRoot, checksum ?? null) as [Field, Field, Field | null, Array<[Field, Field[]]>];
```
iamalwaysuncomfortable: Suggested rename to be more explicit.

```suggestion
            return ExecutionRequest.computeMPCInputs(programName, functionName, inputs, isRoot, checksum ?? null) as [Field, Field, Field | null, Array<[Field, Field[]]>];
```

sdk/src/program-manager.ts
```diff
@@ -3713,6 +3728,24 @@ class ProgramManager {
             imports,
         );
     }
+
+    async signMessage(options: PublicMessagePayload): Promise<[
+        Field,
+        Field,
+        Field | null,
+        Array<[Field, Field[]]>
+        ]> {
```
iamalwaysuncomfortable: This should likely be changed to `computeMPCInputs` since it's just translating types into the their `Field` reprs.  No actual signing is being done here.

```suggestion
    async computeMPCInputs(options: MPCOptions): Promise<MPCInputs> {
```

sdk/src/program-manager.ts
```diff
@@ -220,6 +219,22 @@ interface FeeEstimateOptions {
...
 
+/**
+ * 
+ * @property {string} programName - The name of the program containing the function to execute.
+ * @property {string} functionName - The name of the function to execute within the program.
+ * @property {string[]} inputs - The inputs to the function being executed.
+ * @property {boolean} isRoot - Whether this transition is the first transition being executed in a transaction.
+ * @property {string} [checksum] - The optional checksum of the program, used to verify the program source code on the network matches the program source code used to generate the proof.
+*/
+interface PublicMessagePayload {
```
iamalwaysuncomfortable: Naming should be explicit here like `MpcOptions`

```suggestion
interface MPCOptions {
```

sdk/tests/program-manager.test.ts
```diff
@@ -430,4 +431,28 @@ describe('Program Manager', async () => {
             }
         });
     });
+
+    describe('Program Manager public message signing', () => {
+        it('Should sign a public message and return the expected fields', async () => {
+            const signedMessage = await programManager.signMessage({
```
iamalwaysuncomfortable:
```suggestion
            const signedMessage = await programManager.computeMPCInputs({
```

wasm/src/programs/request.rs
```diff
@@ -219,6 +224,69 @@ impl ExecutionRequest {
...
+    ///
+    /// Returns an array of the form:
+    ///   `[function_id, is_root, program_checksum | null, [[index, [field,...]], ...]]`
+    ///
+    /// @param {string} program_id The id of the program.
+    /// @param {string} function_name The function name.
+    /// @param {string[]} inputs The inputs to the function as strings.
+    /// @param {boolean} is_root Flag to indicate if this is the top level function in the call graph.
+    /// @param {Field | undefined} program_checksum The program checksum (required if the program has a constructor).
+    #[wasm_bindgen(js_name = "computePublicMessagePayload")]
```
iamalwaysuncomfortable:
```suggestion
    #[wasm_bindgen(js_name = "computeMPCInputs)]
```

wasm/src/programs/request.rs
```diff
@@ -219,6 +224,69 @@ impl ExecutionRequest {
...
+    pub fn compute_public_message_payload(
```
iamalwaysuncomfortable:
```suggestion
    pub fn compute_mpc_inputs(
```

iamalwaysuncomfortable (changes requested):
A couple of naming change nits, but that's all, overall this is great

wasm/src/programs/request.rs
Roee-87: done [e30db93](https://github.com/ProvableHQ/sdk/pull/1217/commits/e30db93c06b13542e78f26dd138389f1bddcca17)

wasm/src/programs/request.rs
Roee-87: done [e30db93](https://github.com/ProvableHQ/sdk/pull/1217/commits/e30db93c06b13542e78f26dd138389f1bddcca17)

sdk/tests/program-manager.test.ts
Roee-87: done [e30db93](https://github.com/ProvableHQ/sdk/pull/1217/commits/e30db93c06b13542e78f26dd138389f1bddcca17)

sdk/src/program-manager.ts
Roee-87: done [e30db93](https://github.com/ProvableHQ/sdk/pull/1217/commits/e30db93c06b13542e78f26dd138389f1bddcca17)

sdk/src/program-manager.ts
Roee-87: done.  Defined a `MPCInput` type:
[e30db93](https://github.com/ProvableHQ/sdk/pull/1217/commits/e30db93c06b13542e78f26dd138389f1bddcca17)

sdk/src/program-manager.ts
Roee-87: done [e30db93](https://github.com/ProvableHQ/sdk/pull/1217/commits/e30db93c06b13542e78f26dd138389f1bddcca17)

sdk/tests/program-manager.test.ts
```diff
@@ -305,6 +307,44 @@ describe('Program Manager', async () => {
...
+                true
+            );
+            
+            // Ensure the execution request is valid.
+            const provingRequest = await programManager.provingRequest({
+                programName: "credits.aleo",
+                functionName: "transfer_public",
+                priorityFee: 0,
+                privateFee: false,
+                inputs,
```
vicsn: We shouldn't have to pass in `inputs`, they are encoded in the `executionRequest`. We definitely should not have to pass in the `privateKey` because in the MPC context we don't have access to it.

sdk/src/program-manager.ts
```diff
@@ -220,6 +223,37 @@ interface FeeEstimateOptions {
...
+ * 0 → Field representation of the function ID
+ * 1 → Field representation of whether this is the root transition (will be 1field or 0field)
+ * 2 → optional checksum
+ * 3 → array of [index value of the input, [array of field elements representing the input data]]
+ */
+type MPCInputs = [
+    Field,
+    Field,
+    Field | null,
+    Array<[Field, Field[]]>
```
vicsn: What about changing this to a strongly typed and more readable enum per Input Type (Public/Private/Record to start with).

wasm/src/programs/request.rs
```diff
@@ -219,6 +224,69 @@ impl ExecutionRequest {
...
+
+        // Compute the function ID.
+        let function_id =
+            Field::from(compute_function_id(&network_id, &program_id, &function_name).map_err(|e| e.to_string())?);
+
+        // Compute 'is_root' as a field element.
+        let is_root_field = Field::from(if is_root { FieldNative::one() } else { FieldNative::zero() });
+
+        // Build the per-input data array.
+        let input_data = Array::new();
```
vicsn: A few more requests:
- records, can we also let this function output h and tag? See: https://github.com/ProvableHQ/snarkVM/blob/mainnet/console/program/src/request/sign.rs#L189
- private, can we let this function output input_hash = N::hash_psd8(&ciphertext.to_fields()?)?; See: https://github.com/ProvableHQ/snarkVM/blob/mainnet/console/program/src/request/sign.rs#L165

sdk/tests/program-manager.test.ts
iamalwaysuncomfortable: Yeah the test needs to be revised here, but you don't actually need to pass these things in. We'll update the test.

sdk/src/program-manager.ts
iamalwaysuncomfortable: I don't actually understand the context of this comment. This is already strongly typed typescript already. The issue specified this as the required output type:

> In terrible pseudocode, the function signature of this new exposed SDK utility could be:
function compute_public_message_payload(program_id: String, function_name: String, inputs: [ValueType<N>]) -> (Field<N>, Field<N>, Field<N>, [(Field<N>, [Field<N>])])

This is what the `computeMPCInputs` function emits, the issue explicitly asked for exactly this type signature. Are we iterating on what that type signature should look like?

wasm/src/programs/request.rs
iamalwaysuncomfortable: Sure!

sdk/src/program-manager.ts
iamalwaysuncomfortable: Inferring a bit, it seems like this is the closer to the type signature you're imagining. 

I.E. a structured object that names the fields instead of release an array with no naming as was specified in the original issue.

```typescript
interface RequestOutputs {
    outputType: "i8" | "i16" | "i32" | "i64" | "i128" | "u8" | "u16" | "u32" | "u64" | "u128" | "signature" | "field" | "address" | "group" | "struct" | "array" | "record";
    index: Field;
    data: [Field];
    input_hash: Field;
    name?: string; // the record or struct name.
    h?: Field // optional h (only used in case of a record.)
    tag?: Field // optional tag (only used in case of a record.)
}

interface MpcInput:
{
  function_id: Field;
  is_root: Field;
  outputs: RequestOutputs[]
  program_checksum?: Field
} 
```

If so this is a different type signature than the original ask and it should be confirmed with those asking for it if this is the type signature they want to see. 

If it's confirmed, this is certainly do-able.

wasm/src/programs/request.rs
iamalwaysuncomfortable: See the suggested type signature above to see if the placement of `h` and `tag` is what's desired.

sdk/src/program-manager.ts
vicsn: > Are we iterating on what that type signature should look like?

Indeed, to make it clearer :) That looks already more readable than the typeless array. My only nit is that I would rename `RequestOutputs` to `RequestSignInputs` or something, because they are still metadata about **inputs** to the function call.

vicsn: Update: `input_hash` is no longer needed

wasm/src/programs/request.rs
vicsn: Update: `input_hash` is no longer needed.

sdk/src/program-manager.ts
iamalwaysuncomfortable: Then the above is what we'll do!

wasm/src/programs/request.rs
Roee-87: Calculating h will require the user submit a View Key.  I think in our implementation we said that this would be available to all of the MPC participants, correct?

vicsn: Correct

sdk/src/models/ExternalSigningInputs.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Type surrounding the input to a MPC request.
+ *
+ * @property {string} outputType - The type of output being requested.
+ * @property {string} index - The index of the output being requested represented as an Ed/BLS-377 base field element.
+ * @property {string[]} data - The data being requested represented as Ed/BLS-377 base field elements.
+ * @property {string} [name] - The name of the record being requested.
+ * @property {string} [h] - The h value of the record being requested represented as an Ed/BLS-377 base field element.
+ * @property {string} [tag] - The tag of the record being requested represented as an Ed/BLS-377 base field element.
+ */
+export interface RequestSignInput {
```
vicsn: Is it possible to use a `Field` representation for all components (except `outputType`)?

sdk/src/models/MPCInputs.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Type representing the input to a MPC request.
+ *
+ * @property {string} functionId - The ID of the function the request is for serialized to Ed/BLS-377 base field elements.
+ * @property {string} isRoot - Field representation of a boolean indicating whether this is a top-level transition (i.e. "0field" or "1field").
+ * @property {RequestSignInput[]} requestInputs - The inputs to the function being executed.
+ * @property {string} [checksum] - The Ed/BLS-377 base field representation of the program checksum (for programs that contain constructors).
+ */
+export interface MPCInput {
```
vicsn: The functionality in this PR may also be needed when the private key lives in e.g. a minimal TEE.

Can we adjust the naming of all relevant new functions/structures/comments to e.g. `ExternallySigned`/`ExternalSigning` instead of using `MPC`?

vicsn: Done: 94947b87

sdk/src/models/ExternalSigningInputs.ts
```diff
@@ -0,0 +1,59 @@
+import { Address, Field, ViewKey } from "@provablehq/wasm";
+
+/**
+ * Type surrounding the input to an external signing request.
+ *
+ * @property {string} outputType - The type of output being requested.
```
vicsn: All instances of `outputType` in this PR should be `inputType`

wasm/src/programs/request.rs
```diff
@@ -219,6 +420,178 @@ impl ExecutionRequest {
...
+    #[wasm_bindgen(js_name = "computeExternalSigningInputs")]
+    pub fn compute_external_signing_inputs(
+        program_id: String,
+        function_name: String,
+        inputs: Array,
+        input_types: Array,
+        is_root: bool,
+        program_checksum: Option<Field>,
+        view_key: Option<ViewKey>,
+    ) -> Result<Object, String> {
```
vicsn, line 573: TODO: we also need to return the `commitment`

wasm/src/programs/request.rs
```diff
@@ -206,6 +218,195 @@ impl ExecutionRequest {
...
+    pub fn from_externally_signed_data(
+        program_id: String,
+        function_name: String,
+        inputs: Array,
+        input_types: Array,
+        signature: Signature,
+        tvk: Field,
+        signer: Address,
+        sk_tag: Field,
+        view_key: Option<ViewKey>,
```
vicsn: As discussed: we should:
- remove the `view_key`
- make `gammas` required (and not an option)

We can always create a different methods which does depend on this.

iamalwaysuncomfortable (commented):
@vicsn please review the latest changes and let us know if anything needs to change.

sdk/src/models/mpc.ts
```diff
@@ -0,0 +1,211 @@
...
+ *
+ * @property {string} outputType - The type of output being requested.
+ * @property {FieldOutput<F>} index - The index of the output represented as an Ed/BLS-377 base field element.
+ * @property {FieldOutput<F>[]} data - The data represented as Ed/BLS-377 base field elements.
+ * @property {string} [name] - The name of the record being requested.
+ * @property {FieldOutput<F>} [h] - The h value of the record represented as an Ed/BLS-377 base field element.
+ * @property {FieldOutput<F>} [tag] - The tag of the record represented as an Ed/BLS-377 base field element.
+ */
+export interface RequestSignInput<F extends OutputFormat = "string"> {
+    outputType: "constant" | "public" | "private" | "record" | "external_records";
```
iamalwaysuncomfortable:
```suggestion
    outputType: "constant" | "public" | "private" | "record" | "external_record";
```

sdk/src/models/external-signing.ts
```diff
@@ -0,0 +1,213 @@
+import {
```
vicsn, line 1: Can we not name any files or objects "mpc" but "externalSigning" or "externallySigned"

sdk/src/models/mpc.ts
```diff
@@ -0,0 +1,213 @@
...
+ * @property {string} outputType - The type of output being requested.
+ * @property {FieldOutput<F>} index - The index of the output represented as an Ed/BLS-377 base field element.
+ * @property {FieldOutput<F>[]} data - The data represented as Ed/BLS-377 base field elements.
+ * @property {string} [name] - The name of the record being requested.
+ * @property {FieldOutput<F>} [h] - The h value of the record represented as an Ed/BLS-377 base field element.
+ * @property {FieldOutput<F>} [tag] - The tag of the record represented as an Ed/BLS-377 base field element.
+ * @property {FieldOutput<F>} [recordViewKey] - The record view key (present for record inputs when viewKey was provided).
+ */
+export interface RequestSignInput<F extends OutputFormat = "string"> {
+    outputType: "constant" | "public" | "private" | "record" | "external_record";
```
vicsn: All instances of `outputType` in this PR should be some derivation of `inputType`.

iamalwaysuncomfortable: I'm assuming you mean because we want to distinguish from actual transition outputs - something like `signingInputType`.

vicsn: Indeed!

sdk/src/models/external-signing.ts
```diff
@@ -0,0 +1,213 @@
...
+ * @property {FieldOutput<F>} [checksum] - The Ed/BLS-377 base field representation of the program checksum.
+ * @property {string} [signer] - The signer address string (present when viewKey was provided).
+ * @property {string} [skTag] - The tag secret key string (present when viewKey was provided).
+ */
+export interface ExternalSigningInput<F extends OutputFormat = "string"> {
+    functionId: FieldOutput<F>;
+    isRoot: boolean;
+    requestInputs: RequestSignInput<F>[];
+    checksum?: FieldOutput<F>;
+    signer?: string;
```
vicsn, line 131: `signer` has to be removed.

It is not used by current MPC libraries - any external signer has the private key and therefore knows who the signer is supposed to be.. More importantly, I learned today that if they would use a given `signer`, it would be a potential vulnerability. So we better remove it for clarity and security.

iamalwaysuncomfortable (approved):

PR #1227, opened 2026-03-09
[Feature] Return proving/verifying keys from execution methods

marshacb:
## Motivation

SnarkVM assumes proving/verifying keys persist in long-running processes, but SDK users typically run ephemeral processes where synthesized keys are discarded after each execution. Since key synthesis is the most expensive part of execution (30s+ for complex functions), this meant users were paying that cost repeatedly with no way to retrieve or reuse keys.

This PR adds `cacheKeys: true` to `ExecuteOptions`, enabling:

- **Key extraction**: `buildExecutionTransaction()` and `run()` return the synthesized proving/verifying keys alongside execution results
- **Automatic persistence**: Keys are auto-persisted to the configured `KeyStore` (e.g., `LocalFileKeyStore` from #1207)
- **Transparent reuse**: Subsequent executions automatically check memory cache → KeyStore → network/synthesis, so keys are reused with zero user intervention

See `docs/design/key-caching.md` for the full design spec.

## What changed

- **SDK (`program-manager.ts`)**: Added `cacheKeys` option to `ExecuteOptions`, `ExecutionTransactionResult` return type with overloaded `buildExecutionTransaction()`, memory + KeyStore caching in both `buildExecutionTransaction()` and `run()`
- **SDK (`keys/provider/memory.ts`)**: `AleoKeyProvider` now accepts an optional `KeyStore` via constructor, `functionKeys()` falls back to KeyStore when memory cache misses
- **WASM (`execute.rs`, `response.rs`)**: New `ExecutionTransactionResponse` type and `buildExecutionTransactionWithKeys` method that extracts proving/verifying keys from the process before it's dropped

## Follow-up

- `run()` positional parameter fragility — should migrate to an options object (existing tech debt)

## Test plan

- 7 unit tests in `key-provider.test.ts`: KeyStore constructor wiring, `setKeyStore()`, memory → KeyStore fallback chain, memory cache priority, error paths — all passing
- 1 integration test in `program-manager.test.ts`: `run()` with `cacheKeys: true` end-to-end — synthesizes keys via WASM, persists to `LocalFileKeyStore`, verifies keys on disk
- 1 skipped integration test: `buildExecutionTransaction` with `cacheKeys: true` — requires a funded account and network access, to be enabled when test infrastructure supports it
- Full regression: 41/41 key-provider tests passing, 0 new failures in program-manager suite

<details>
<summary>How to run</summary>

```bash
# Full build (WASM must be built first for Rust changes)
yarn build:all

# Run tests
cd sdk && yarn test

### DIFF
--- a/docs/design/key-caching.md
+++ b/docs/design/key-caching.md
@@ -1,0 +1,215 @@
+---
+id: key-caching
+title: "Design: Execution Key Caching & Persistence"
+---
+
+# Execution Key Caching & Persistence
+
+## Problem
+
+SnarkVM assumes proving and verifying keys persist in a long-running process. In practice, SDK users run ephemeral
+processes (serverless functions, CLI tools, web workers) where the VM is created, used, and destroyed. When a process
+exits, all synthesized keys are lost.
+
+Key synthesis is the most expensive part of program execution -- often 30+ seconds for complex functions. Without key
+persistence, every execution pays this cost, even for the same program and function.
+
+Users faced three problems:
+1. No way to retrieve keys from execution methods for custom storage
+2. No automatic persistence -- keys were discarded after each execution
+3. Manual key management required understanding synthesis, caching, and locator conventions
+
+## Design Goals
+
+- **Zero-config improvement**: Users who add `cacheKeys: true` get automatic key persistence with no other changes
+- **Backward compatible**: All existing code works identically; `cacheKeys` defaults to `false`
+- **Layered access**: Users can use automatic persistence, manual key extraction, or both
+- **No new dependencies**: Built on the existing `KeyStore` interface (PR #1207) and `AleoKeyProvider`
+
+## Architecture
+
+```
+                        buildExecutionTransaction({ cacheKeys: true })
+                                        |
+                                        v
+                    +-------------------------------------------+
+                    |           TypeScript SDK Layer             |
+                    |  ProgramManager.buildExecutionTransaction  |
+                    +-------------------------------------------+
+                            |                       |
+                            v                       v
+            +------------------------+    +-------------------+
+            |      WASM Layer        |    |  persistKeysToStore()
+            | buildExecutionTx-      |    |  (auto-persist to  |
+            |   WithKeys()           |    |   KeyStore)        |
+            +------------------------+    +-------------------+
+                            |                       |
+                            v                       v
+            +------------------------+    +-------------------+
+            | ExecutionTransaction-   |    |    KeyStore       |
+            |   Response             |    | (LocalFileKeyStore |
+            | { tx, pk?, vk? }       |    |  or custom impl)  |
+            +------------------------+    +-------------------+
+
+            Subsequent executions:
+            AleoKeyProvider.functionKeys()
+                1. Memory cache  -->  hit? return
+                2. KeyStore      -->  hit? cache in memory, return
+                3. Network/synth -->  fallback
+```
+
+## API Surface
+
+### New option: `cacheKeys`
+
+Added to `ExecuteOptions`, used by `buildExecutionTransaction()`, `execute()`, and `run()`.
+
+```typescript
+const result = await programManager.buildExecutionTransaction({
+    programName: "my_program.aleo",
+    functionName: "my_function",
+    inputs: ["1u32", "2u32"],
+    priorityFee: 0.0,
+    privateFee: false,
+    cacheKeys: true,  // opt-in to key return + auto-persistence
+});
+
+result.transaction;    // Transaction (always present)
+result.provingKey;     // ProvingKey | undefined
+result.verifyingKey;   // VerifyingKey | undefined
+```
+
+When `cacheKeys` is `false` (default), `buildExecutionTransaction` returns `Transaction` directly -- identical to
+existing behavior.
+
+TypeScript function overloads enforce this at the type level:
+
+```typescript
+// cacheKeys: true  --> returns ExecutionTransactionResult
+// cacheKeys: false --> returns Transaction
+```
+
+### Auto-persistence
+
+When a `KeyStore` is configured on the `AleoKeyProvider`, keys are automatically persisted after execution:
+
+```typescript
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: wasm/src/programs/manager/execute.rs [hunk 1]
```diff
@@ -295,6 +296,159 @@ impl ProgramManager {
...
+            let pk = process
+                .get_proving_key(program_native.id(), &function_name)
+                .map_err(|_| format!("Could not find proving key for {}/{}", program_native.id(), function))?;
+            let vk = process
+                .get_verifying_key(program_native.id(), &function_name)
+                .map_err(|_| format!("Could not find verifying key for {}/{}", program_native.id(), function))?;
+            (Some(pk), Some(vk))
+        } else {
+            (None, None)
+        };
```
iamalwaysuncomfortable, line 377: A couple of notes here:

1. In the case of imports, sometimes those are synthesized and added to the process, we want to extract those as well.
2. Perhaps instead of a new transaction type, we might want to pass an object into an existing method that collects the keys like the one you came up with in which can then be extracted later?

sdk/src/keys/provider/memory.ts
```diff
@@ -77,14 +78,24 @@ class AleoKeyProvider implements FunctionKeyProvider {
...
+        return this._keyStore;
+    }
+
+    /**
+     * Set the key store for persistent key storage
+     *
+     * @param {KeyStore | undefined} keyStore The key store to use for persistent storage
+     */
+    setKeyStore(keyStore: KeyStore | undefined) {
+        this._keyStore = keyStore;
```
iamalwaysuncomfortable, line 98: There's this implementation of adding the KeyStore to the key provider waiting in a PR: #1221 - let's see if we can merge the approaches tomorrow.

sdk/src/program-manager.ts
```diff
@@ -941,6 +1002,53 @@ class ProgramManager {
         }
 
         // Build an execution transaction
+        if (options.cacheKeys) {
+            const response: ExecutionTransactionResponse = await WasmProgramManager.buildExecutionTransactionWithKeys(
```
iamalwaysuncomfortable, line 1006: In this outer scope, one thing we might consider instead of a different `wasm` method is passing in the `wasm` object you defined and at the end, get the `pks/vks` out of it and then `destroy/free` it.

sdk/src/program-manager.ts
```diff
@@ -97,7 +100,22 @@ interface ExecuteOptions {
...
+ * Contains the transaction and the proving/verifying keys that were synthesized during execution.
+ *
+ * @property {Transaction} transaction - The execution transaction.
+ * @property {ProvingKey} provingKey - The proving key synthesized during execution.
+ * @property {VerifyingKey} verifyingKey - The verifying key synthesized during execution.
+ */
+interface ExecutionTransactionResult {
+    transaction: Transaction;
+    provingKey: ProvingKey;
+    verifyingKey: VerifyingKey;
```
iamalwaysuncomfortable, line 118: Since `wasm` is limited to 4gb total memory space, we probably want is to take these out of `wasm` and destroy them so as not to OOM the process. We also want to return the checksums so that people have a future way of verifying the integrity of the keys.

```suggestion
    provingKey: Uint8Array;
    provingKeyChecksum: string;
    verifyingKey: Uint8Array;
    verifyingKeyChecksum: string;
```

sdk/src/program-manager.ts
```diff
@@ -786,9 +841,15 @@ class ProgramManager {
...
+    buildExecutionTransaction(
+        options: ExecuteOptions & { cacheKeys: true },
+    ): Promise<ExecutionTransactionResult>;
+    buildExecutionTransaction(
+        options: ExecuteOptions,
+    ): Promise<Transaction>;
     async buildExecutionTransaction(
         options: ExecuteOptions,
-    ): Promise<Transaction> {
+    ): Promise<Transaction | ExecutionTransactionResult> {
```
iamalwaysuncomfortable, line 852: I originally though this signature might be good, but I think it might break a lot of downstream typescript implementors who will have to account for both types in further type lints.

I think instead we want to keep the original signature the same and allow callers to pass in an imports object which gets mutated and callers can extract the imports from there. OR if they have the keyprovider configured, they can extract it from there as well.

docs/design/key-caching.md
```diff
@@ -0,0 +1,215 @@
...
+
+const programManager = new ProgramManager(networkClient, keyProvider);
+programManager.setAccount(account);
+
+// First execution: synthesizes keys, persists to disk
+await programManager.execute({
+    programName: "my_program.aleo",
+    functionName: "my_function",
+    inputs: ["1u32", "2u32"],
+    cacheKeys: true,
```
iamalwaysuncomfortable, line 111: We might consider changing this to a KeyCache which gets populated with anything not currently existing.

wasm/src/programs/response.rs
```diff
@@ -148,3 +148,71 @@ impl Deref for ExecutionResponse {
...
 }
+
+/// Response from building an execution transaction, containing the transaction
+/// and optionally the proving/verifying keys that were synthesized or used.
+#[wasm_bindgen]
+pub struct ExecutionTransactionResponse {
+    transaction: TransactionNative,
+    proving_key: Option<ProvingKeyNative>,
+    verifying_key: Option<VerifyingKeyNative>,
+}
```
iamalwaysuncomfortable, line 159: I think here we want to go in favor of a typescript object instead that gets populated outside the wasm scope. But what we probably want is the [ProgramImports](https://github.com/ProvableHQ/sdk/blob/54b34eb736a5dff5c679bc02d09b753a27ebd14d/wasm/src/programs/manager/imports.rs#L64) object you created in PR #1208 and use that to collect the keys. Then when the execution terminates, we extract the keys out of it into either the KeyProvider or a separate typescript object.

iamalwaysuncomfortable (changes requested):
Some changes requested, and likely some unification needed with PR #1208. We can use the ProgramImports abstraction you built there to collect the imports and then store them in typescript.

marshacb:
covered in #1208

PR #1228, opened 2026-03-09
[Feature] Construct request from MPC outputs

iamalwaysuncomfortable:
## Motivation

This pull request introduces new functionality for constructing `ExecutionRequest`s from MPC (Multi-Party Computation) outputs, both in Rust and in the TypeScript SDK tests, and adds support for debug builds of the WASM package. The changes improve the ability to test and interoperate with MPC-generated signatures and related fields, especially for private transfers that involve record inputs.

### DIFF
--- a/sdk/tests/program-manager.test.ts
+++ b/sdk/tests/program-manager.test.ts
@@ -620,6 +620,7 @@ describe('Program Manager', async () => {
             const tcm = signedRequest.tcm().toString();
             const viewKeyString = viewKey.toString();
             console.log(2);
+            console.log(signature);
             const mpcRequest = ExecutionRequest.fromMPCWithStrings(
                 "credits.aleo",
                 "transfer_private",
@@ -631,7 +632,7 @@ describe('Program Manager', async () => {
                 tvk,
                 tcm,
                 viewKeyString,
-                undefined, //recordInputIdsJson,
+                recordInputIdsJson,
             );
             console.log(3);
             expect(mpcRequest.program_id()).to.equal(signedRequest.program_id());
--- a/sdk/tests/program-manager.test.ts
+++ b/sdk/tests/program-manager.test.ts
@@ -615,6 +615,11 @@ describe('Program Manager', async () => {
             if (recordId0 == null) throw new Error("Expected record input ID at index 0");
             const recordInputIdsJson = JSON.stringify([String(recordId0)]); // first input is the record
             console.log(recordInputIdsJson);
+            const signature = signedRequest.signature().toString();
+            const tvk = signedRequest.tvk().toString();
+            const tcm = signedRequest.tcm().toString();
+            const viewKeyString = viewKey.toString();
+            console.log(2);
             const mpcRequest = ExecutionRequest.fromMPCWithStrings(
                 "credits.aleo",
                 "transfer_private",
@@ -622,12 +627,13 @@ describe('Program Manager', async () => {
                 transferPrivateInputTypes,
                 true,
                 undefined,
-                signedRequest.signature().toString(),
-                signedRequest.tvk().toString(),
-                signedRequest.tcm().toString(),
-                viewKey.toString(),
-                recordInputIdsJson,
+                signature,
+                tvk,
+                tcm,
+                viewKeyString,
+                undefined, //recordInputIdsJson,
             );
+            console.log(3);
             expect(mpcRequest.program_id()).to.equal(signedRequest.program_id());
             expect(mpcRequest.function_name()).to.equal(signedRequest.function_name());
             expect(mpcRequest.inputs().length).to.equal(signedRequest.inputs().length);
--- a/sdk/tests/program-manager.test.ts
+++ b/sdk/tests/program-manager.test.ts
@@ -565,11 +565,8 @@ describe('Program Manager', async () => {
                 "transfer_public",
                 transferPublicInputs,
                 transferPublicInputTypes,
-                true,
-                undefined,
                 signedRequest.signature(),
                 signedRequest.tvk(),
-                signedRequest.tcm(),
                 viewKey,
                 undefined, // no record inputs
             );
@@ -583,8 +580,8 @@ describe('Program Manager', async () => {
         it('Should match ExecutionRequest.sign for transfer_private', async () => {
             // Use beacon private key - record is owned by beacon address
             const privateKey = PrivateKey.from_string(beaconPrivateKeyString);
-            const viewKey = ViewKey.from_private_key(privateKey);
-            const viewKeyString = viewKey.to_string(); // extract before viewKey is consumed by computeMPCInputs
+            const beaconRecord = RecordPlaintext.fromString(recordBeaconOwned);
+            const gamma = beaconRecord.gamma("credits.aleo", "credits", privateKey);
             const mpcInputs = await programManager.computeMPCInputs({
                 programName: "credits.aleo",
                 functionName: "transfer_private",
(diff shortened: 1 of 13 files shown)


wasm/src/programs/request.rs
```diff
@@ -215,6 +218,238 @@ impl ExecutionRequest {
...
+    /// @param {Field | undefined} program_checksum The program checksum (required if the program has a constructor).
+    /// @param {Signature} signature The MPC-computed signature.
+    /// @param {Field} tvk The transition view key.
+    /// @param {Field} tcm The transition commitment.
+    /// @param {ViewKey} view_key The view key of the signer.
+    /// @param {string | undefined} record_input_ids_json JSON string of record input ID array (one per record input, in order).
+    ///        Required when inputs include records. E.g. '["{\"type\":\"record\",...}"]'. Each element is the JSON representation of InputID::Record.
+    #[wasm_bindgen(js_name = "fromMPC")]
+    #[allow(clippy::too_many_arguments)]
+    pub fn from_mpc(
```
iamalwaysuncomfortable, line 239: @vicsn  Do we need two implementations? Or would the string impl suffice?

wasm/src/programs/request.rs
```diff
@@ -215,6 +218,238 @@ impl ExecutionRequest {
...
+                ValueTypeNative::from_str(&input_type.as_string().unwrap()).map_err(|e| e.to_string())
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        if input_types.len() != inputs.len() {
+            return Err("input_types and inputs must have the same length".to_string());
+        }
+
+        let signer = crate::types::native::AddressNative::try_from(&view_key_native).map_err(|e| e.to_string())?;
+        let sk_tag = crate::types::native::GraphKeyNative::try_from(view_key_native).map_err(|e| e.to_string())?.sk_tag();
```
iamalwaysuncomfortable: These should be imported at the top level.

wasm/src/programs/request.rs
```diff
@@ -215,6 +218,238 @@ impl ExecutionRequest {
...
+        function_name: String,
+        inputs: Array,
+        input_types: Array,
+        _is_root: bool,
+        _program_checksum: Option<Field>,
+        signature_native: SignatureNative,
+        tvk_native: FieldNative,
+        tcm_native: FieldNative,
+        view_key_native: ViewKeyNative,
+        record_input_ids_json: Option<String>,
```
iamalwaysuncomfortable: This should be deserialized at the JS level, not here.

wasm/src/programs/request.rs
```diff
@@ -215,6 +218,238 @@ impl ExecutionRequest {
...
+        let network_id = U16Native::new(CurrentNetwork::ID);
+        let function_id = compute_function_id(&network_id, &program_id, &function_name).map_err(|e| e.to_string())?;
+
+        let record_input_id_strs: Vec<String> = record_input_ids_json
+            .as_ref()
+            .map(|json| {
+                serde_json::from_str::<Vec<String>>(json).map_err(|e| format!("record_input_ids_json must be JSON array of strings: {e}"))
+            })
+            .transpose()?
+            .unwrap_or_default();
```
iamalwaysuncomfortable: Double-tapping here, these should be deserialized at the JS level.

wasm/src/programs/request.rs
iamalwaysuncomfortable: Also what's the motivation for this being a separate parameter?

wasm/src/programs/request.rs
```diff
@@ -215,6 +218,238 @@ impl ExecutionRequest {
...
+        )
+    }
+
+    fn from_mpc_impl(
+        program_id: String,
+        function_name: String,
+        inputs: Array,
+        input_types: Array,
+        _is_root: bool,
+        _program_checksum: Option<Field>,
```
iamalwaysuncomfortable: Do we need these?

sdk/tests/program-manager.test.ts
```diff
@@ -547,16 +558,91 @@ describe('Program Manager', async () => {
...
+                transferPublicInputs,
+                transferPublicInputTypes,
+                true,
+                undefined,
+                signedRequest.signature(),
+                signedRequest.tvk(),
+                signedRequest.tcm(),
+                viewKey,
+                undefined, // no record inputs
+            );
```
iamalwaysuncomfortable, line 572: Do we need `fromMPC` or will the string method suffice?

iamalwaysuncomfortable (commented):
@vicsn  A few questions about this impl.

iamalwaysuncomfortable:
@vicsn the human readable separator issue showing up in the tests is due to attempting to deserialize within Rust. Deserialize the JSON string with JS outside of rust first and then use an `Object` accessor from `js_sys` to do [ref](https://docs.rs/js-sys/latest/js_sys/struct.Object.html)

wasm/src/programs/request.rs
vicsn: A single one suffices, I have a slight preference for `fromMPC`

wasm/src/programs/request.rs
vicsn: No, the underscored variables can be removed.

wasm/src/programs/request.rs
```diff
@@ -215,6 +217,229 @@ impl ExecutionRequest {
...
+        signature_native: SignatureNative,
+        tvk_native: FieldNative,
+        tcm_native: FieldNative,
```
iamalwaysuncomfortable: No longer needed.

wasm/src/programs/request.rs
vicsn: Sure, the earlier it's made into the correct type the better I guess.

But indeed it's weird that we have to pass in `record_input_ids` separately, whereas all the other input_id types we compute in this function.

Currently I think the only thing we have to pass in manually which was computed in MPC is `gamma`. And then we can just compute the record input_id.

wasm/src/programs/request.rs
vicsn: Sure

sdk/tests/program-manager.test.ts
Roee-87: removed string method: [f208901](https://github.com/ProvableHQ/sdk/pull/1228/commits/f208901405a7cf1a812f5cb06052c7825e685723)

wasm/src/programs/request.rs
Roee-87: removed: [f208901](https://github.com/ProvableHQ/sdk/pull/1228/commits/f208901405a7cf1a812f5cb06052c7825e685723)

wasm/src/programs/request.rs
Roee-87: removed: [f208901](https://github.com/ProvableHQ/sdk/pull/1228/commits/f208901405a7cf1a812f5cb06052c7825e685723)

wasm/src/programs/request.rs
```diff
@@ -215,6 +218,167 @@ impl ExecutionRequest {
...
+    #[allow(clippy::too_many_arguments)]
+    pub fn from_mpc(
+        program_id: String,
+        function_name: String,
+        inputs: Array,
+        input_types: Array,
+        signature: Signature,
+        tvk: Field,
+        view_key: ViewKey,
+        gammas: Option<Array>,
```
iamalwaysuncomfortable, line 247: @vicsn we still might consider just accepting all string types here to avoid putting the burden of callers needing to do deserialization at this outer scope.

iamalwaysuncomfortable (commented):
@vicsn @Roee-87 PR implements the desired interface and is ready for re-review.

Roee-87 (approved):
wasm tests and sdk tests pass.

vicsn (approved):
This looks great.

PR #1236, opened 2026-03-12
[Feature] Re-export dynamic record from SnarkVM

iamalwaysuncomfortable:
## Motivation

This pull request introduces support for the `DynamicRecord` type in the SDK, enabling fixed-size representations of Aleo records with Merkle roots. It adds Rust and JS/TS bindings, exposes the type in the SDK API, and provides comprehensive tests and documentation. Additionally, there are minor documentation improvements and a bugfix for handling identifier literals.

This PR is required in order to convert records used in dynamic dispatch into dynamic record inputs.

## Test plan

This PR adds several JS and wasm tests


sdk/tests/wasm.test.ts
```diff
@@ -434,6 +434,71 @@ describe('WASM Objects', () => {
...
+            expect(fields.length).greaterThan(0);
+        });
+
+        it('returns a non-empty bit array', () => {
+            const bits = dynamicRecord.toBitsLe();
+            expect(bits).to.be.an('array');
+            expect(bits.length).greaterThan(0);
+        });
+
+        it('can convert back to a RecordPlaintext', () => {
```
marshacb: nit: The `toRecord` test checks the owner address but not the visibility suffix. Should we test the `owner_is_private` flag on both the JS and WASM sides? `toRecord(false)` would pass the same assertion as it currently is. We could consider asserting `.private` in the suffix and adding a `toRecord(false)` case asserting `.public`. Could also assert the nonce round trips. Not blocking.

marshacb (approved):
Looks great! Nice addition. Left one small nit on test coverage but nothing blocking 👍

sdk/tests/wasm.test.ts
iamalwaysuncomfortable: We should!

iamalwaysuncomfortable: Implemented!

PR #1237, opened 2026-03-12
[Feature] Create stringToField utility

iamalwaysuncomfortable:
## Motivation

Function signatures which use dynamic dispatch require dynamic program and function names to be converted to fields within their input signatures, this PR adds a utility for converting strings to a field in order to ensure that this conversion can be done within executions and authorizations.

## Test Plan

At test ensuring proper stringToField conversion has been added to ensure proper conversion.

### DIFF
--- a/sdk/src/browser.ts
+++ b/sdk/src/browser.ts
@@ -142,6 +142,7 @@ export {
     RecordPlaintext,
     Signature,
     Scalar,
+    stringToField,
     Transaction,
     Transition,
     U8,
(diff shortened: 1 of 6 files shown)


marshacb (approved):
Looks good! Two minor suggestions: a JS-side test in `wasm.test.ts` (for parity with other utilities) and a `///` doc comment on `string_to_field` so users know this is for dynamic dispatch identifier encoding. Neither are blocking.

PR #1240, opened 2026-03-13
[Refactor] Change KeyLocator from opaque string to KeyId tuple

marshacb:
## Motivation

The `KeyStore` interface used an opaque `locator: string` that didn't enforce any structure and callers could pass `credits.aleo/bond_public`, `transfer_private.prover.421e5a5`, or `inclusion` with no consistency. Keys for different program editions, networks, or key types could silently collide. This replaces the string with a discriminated union `KeyLocator = ProvingKeyLocator | VerifyingKeyLocator | TranslationKeyLocator` so key identity is explicit, type-safe, and collision-free. The `setKeys` method now enforces `prover/verifier` pairing at compile time. Filesystem validation is scoped to `LocalFileKeyStore` rather than the shared interface.

## Test Plan

- Factory functions (`provingKeyLocator`, `verifyingKeyLocator`, `translationKeyLocator`): verify correct fields, defaults for edition/amendment/network, and overrides
- Serialization format: contract tests pin expected output for `prover/verifier` and `translation` locators (includes `amendment` field)
- Store-level validation: raw KeyLocator objects (bypassing factories) with .., /, \, \0 in program, functionName, network, and recordName are caught by LocalFileKeyStore
- All existing `LocalFileKeyStore` tests updated and passing: `set/get/has/delete/clear`, `setKeys`, `getKeyMetadata`, checksum-from-disk restart, corruption detection, checksum match/mismatch, .aleo path behavior, clear() lifecycle


sdk/src/keys/keystore/interface.ts
```diff
@@ -5,13 +5,82 @@ import { KeyFingerprint } from "../verifier/interface.js";
...
+ * @example
+ * const proverId: KeyId = ["credits.aleo", "transfer_private_prover", 0, "mainnet"];
+ * const verifierId: KeyId = ["credits.aleo", "transfer_private_verifier", 0, "mainnet"];
+ */
+export type KeyId = readonly [
+    program: string,
+    func: string,
+    edition: number,
+    network: string,
+];
```
iamalwaysuncomfortable: Can we change this to an interface? Namely I think it's better in terms of keeping everything properly tracked.

```suggestion
type KeyType = "prover" | "verifier";

export interface KeyId {
    program: string;
    functionName: string;
    edition: number;
    network: string;
    keyType: KeyType;
};
```

sdk/src/keys/keystore/interface.ts
```diff
@@ -5,13 +5,82 @@ import { KeyFingerprint } from "../verifier/interface.js";
...
+ * store-level validation.
+ *
+ * @param {string} program - The program name (e.g. "credits.aleo").
+ * @param {string} func - The function name (e.g. "transfer_private").
+ * @param {number} [edition=0] - The program edition.
+ * @param {string} [network] - The network name. Defaults to the build-time network.
+ * @returns {KeyId}
+ * @throws {Error} If program or func contain unsafe characters.
+ */
+export function keyId(program: string, func: string, edition = 0, network = "%%NETWORK%%"): KeyId {
```
iamalwaysuncomfortable:
```suggestion
export function keyId(program: string, functionName: string, edition = 1, network = "%%NETWORK%%", keyType: KeyType): KeyId {
```

sdk/src/keys/keystore/interface.ts
```diff
@@ -5,13 +5,82 @@ import { KeyFingerprint } from "../verifier/interface.js";
 import { ProvingKey, VerifyingKey } from "../../wasm.js";
 
 /**
- * The key locator string and optional fingerprint to verify the integrity of the key.
+ * A 4-value tuple that definitively identifies a function's proving/verifying key.
  *
- * @property {string} locator - The unique identifier for the key.
+ * Callers distinguish prover vs verifier keys by encoding the role in the
+ * `func` component (e.g. `"transfer_private_prover"` / `"transfer_private_verifier"`).
```
iamalwaysuncomfortable: We should just specify what type of key it is via the KeyID interface, we want to keep the name the same as there are some functions out there in the Aleo world with the `_prover` or `_verifier` suffix already.

iamalwaysuncomfortable (changes requested):
This is really good!

Couple of nits:
* We probably want to change the KeyId type to an Interface and switch it to be an object so people or bots don't mess up on ordering.
* The interface should have a field to specify whether it's a proving or verifying key.
* Instead of `func` it should be `functionName` to be consistent with the other interface conventions.

sdk/src/keys/keystore/interface.ts
```diff
@@ -5,13 +5,85 @@ import { KeyFingerprint } from "../verifier/interface.js";
...
+ * @returns {string} A dot-delimited string safe for use as a filename.
  *
- * @property {string} locator - The unique identifier for the key.
+ * @example
+ * serializeKeyId({ program: "credits.aleo", functionName: "transfer_private", edition: 1, network: "mainnet", keyType: "prover" })
+ * // => "credits.aleo.transfer_private.e1.mainnet.prover"
+ */
+export function serializeKeyId(id: KeyId): string {
+    return `${id.program}.${id.functionName}.e${id.edition}.${id.network}.${id.keyType}`;
+}
```
iamalwaysuncomfortable, line 87: Let's move these to the file.ts as they're specific to that implementation.

sdk/src/keys/keystore/interface.ts
```diff
@@ -5,13 +5,85 @@ import { KeyFingerprint } from "../verifier/interface.js";
 import { ProvingKey, VerifyingKey } from "../../wasm.js";
 
 /**
- * The key locator string and optional fingerprint to verify the integrity of the key.
+ * Discriminates whether a {@link KeyId} refers to a proving key or a verifying key.
+ */
+export type KeyType = "prover" | "verifier";
```
iamalwaysuncomfortable:
```suggestion
export type KeyType = "prover" | "verifier" | "translation";
```

iamalwaysuncomfortable: We should probably add a third type here, although how to address the extended TranslationKey interface will be a little bit harder.

sdk/src/keys/keystore/interface.ts
```diff
@@ -5,13 +5,85 @@ import { KeyFingerprint } from "../verifier/interface.js";
...
+ * const proverId: KeyId = { program: "credits.aleo", functionName: "transfer_private", edition: 1, network: "mainnet", keyType: "prover" };
+ * const verifierId: KeyId = { program: "credits.aleo", functionName: "transfer_private", edition: 1, network: "mainnet", keyType: "verifier" };
+ */
+export interface KeyId {
+    program: string;
+    functionName: string;
+    edition: number;
+    network: string;
+    keyType: KeyType;
+}
```
iamalwaysuncomfortable, line 29: Maybe migrate to this?

```typescript
interface BaseFunctionKeyLocator {
      program: string;
      functionName: string;
      edition: number;
      amendment: number;
      network: string;
      checksum?: string;
  }

  interface ProvingKeyLocator extends BaseFunctionKeyLocator {
      keyType: "prover";
  }

  interface VerifyingKeyLocator extends BaseFunctionKeyLocator {
      keyType: "verifier";
  }

  interface TranslationKeyLocator extends BaseFunctionKeyLocator {
      keyType: "translation";
      recordName: string;
      recordInputPosition: number;
  }

  type KeyLocator = ProvingKeyLocator | VerifyingKeyLocator | TranslationKeyLocator;
```

iamalwaysuncomfortable (changes requested):
A few more changes requested based on dynamic dispatch, can you implement these + some tests?

sdk/src/keys/keystore/interface.ts
```diff
@@ -5,19 +5,132 @@ import { KeyFingerprint } from "../verifier/interface.js";
...
+ * @param {number} [amendment=0] - The program amendment.
+ * @param {string} [network] - The network name. Defaults to the build-time network.
+ * @returns {VerifyingKeyLocator}
+ */
+export function verifyingKeyLocator(
+    program: string,
+    functionName: string,
+    edition = 1,
+    amendment = 0,
+    network = "%%NETWORK%%",
```
iamalwaysuncomfortable, line 105: There should be an optional checksum here and in the ProvingKeyLocator and TranslationKeyLocator functions

iamalwaysuncomfortable (commented):
One final comment about building a constructor which accepts an optional checksum

sdk/src/keys/keystore/file.ts
```diff
@@ -30,41 +30,71 @@ export class LocalFileKeyStore implements KeyStore {
...
+     * @throws {InvalidLocatorError} If any component contains unsafe characters.
+     */
+    private serializeLocator(locator: KeyLocator): string {
+        this.validateComponent(locator.program, "program");
+        this.validateComponent(locator.functionName, "functionName");
+        this.validateComponent(locator.network, "network");
+        const base = `${locator.program}.${locator.functionName}.e${locator.edition}.a${locator.amendment}.${locator.network}.${locator.keyType}`;
+        if (locator.keyType === "translation") {
+            this.validateComponent(locator.recordName, "recordName");
+            return `${base}.${locator.recordName}.${locator.recordInputPosition}`;
```
iamalwaysuncomfortable, line 106: For the checksum, if it exists, put it at the end of Proving Keys, Verifying Keys, and Translation Keys

sdk/src/browser.ts
```diff
@@ -77,10 +77,18 @@ import {
...
     KeyLocator,
     KeyStore,
+    KeyType,
+    ProvingKeyLocator,
+    TranslationKeyLocator,
+    VerifyingKeyLocator,
+    provingKeyLocator,
+    verifyingKeyLocator,
+    translationKeyLocator,
 } from "./keys/keystore/interface.js";
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Will resolve generally in a follow-on PR.

sdk/src/browser.ts
```diff
@@ -210,14 +218,22 @@ export {
...
+    KeySearchParams,
     KeyStore,
+    KeyType,
     KeyVerificationError,
+    ProvingKeyLocator,
+    provingKeyLocator,
+    TranslationKeyLocator,
+    translationKeyLocator,
+    VerifyingKeyLocator,
+    verifyingKeyLocator,
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Rollup handles type erasures, false positive

sdk/src/keys/keystore/file.ts
```diff
@@ -30,41 +30,95 @@ export class LocalFileKeyStore implements KeyStore {
...
+     * @throws {InvalidLocatorError} If the value is negative.
+     */
+    private validateNonNegative(value: number, label: string): void {
+        if (value < 0) {
+            throw new InvalidLocatorError(
+                `KeyLocator ${label} must not be negative (got ${value})`,
+                String(value),
+                "negative_value"
+            );
+        }
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Addressed

sdk/src/keys/keystore/interface.ts
```diff
@@ -37,58 +155,51 @@ export interface KeyStore {
      * Returns the `VerifyingKey` for a given locator.
      *
      * @param {KeyLocator} locator The unique locator for the desired `VerifyingKey`.
-     *
      * @returns {Promise<VerifyingKey | null>} Returns the `VerifyingKey` for the given locator if it exists or null if it does not exist.
      */
     getVerifyingKey(locator: KeyLocator): Promise<VerifyingKey | null>;
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Addressed

sdk/tests/key-provider.test.ts
```diff
@@ -185,85 +262,119 @@ describe("KeyStore (file) – LocalFileKeyStore", () => {
...
             try {
-                await keystore.has("dir\\key.prover");
+                await keystore.getKeyBytes(bad);
                 expect.fail("should throw");
             } catch (e) {
                 expect(e).instanceof(InvalidLocatorError);
-                expect((e as InvalidLocatorError).reason).equal("path_separator");
+                expect((e as InvalidLocatorError).reason).equal("path_traversal");
             }
         });
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Addressed

sdk/src/keys/keystore/file.ts
```diff
@@ -30,41 +30,95 @@ export class LocalFileKeyStore implements KeyStore {
...
-                locator,
+                `KeyLocator ${label} must not contain ".." (got "${value}")`,
+                value,
                 "path_traversal"
             );
         }
-
-        // Block path separators and null byte
-        if (locator.includes("/") || locator.includes("\\") || locator.includes("\0")) {
+        if (value.includes("/") || value.includes("\\") || value.includes("\0")) {
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Addressed

iamalwaysuncomfortable (approved):
Added a few further type verifications + narrowed the interface for getProvingKey and getVerifyingKey.

@marshacb we'll probably want to add further methods on the interface to address the translation keys, but we should only do when the mainnet dynamic dispatch release lands (as we won't be able to support the corresponding `wasm` types until that point. So this further strengthens that we should wait to use the KeyStore method of resolving multiple ProvingKeys and VerifyingKeys (+ their translation keys) until that point so that no-one uses this interface until the release of dynamic dispatch on mainnet.

PR #1252, opened 2026-03-19
[Fix] Refactor JWT refresh logic to use local variables

iamalwaysuncomfortable:
This fixes a small bug in the SDK which doesn't correctly refresh the JWT if the API key and Consumer ID are set in the network client itself.


miazn (approved):
lgtm

PR #1248, opened 2026-03-17
RecordScanner: fix JWT refresh endpoint

melonges:
Fixes #1247.

## Motivation

`RecordScanner.refreshJwt` currently posts to the wrong JWT endpoint (`/scanner/:network/jwts/:consumerId`), which returns 401 and prevents authentication. This PR updates the request URL to the correct JWT endpoint.

## Test Plan

- Instantiate `RecordScanner` with an API key and consumer ID.
- Call `refreshJwt` (or trigger the auth flow) and verify the request goes to `https://api.provable.com/jwts/:consumerId`.
- Confirm the response includes the `authorization` header and subsequent RecordScanner operations succeed.

## Related PRs

None.

---

> [!NOTE]
> **Low Risk**
> Low risk: a small change to the JWT refresh URL and request header formatting, affecting only the authentication refresh flow.
> 
> **Overview**
> Fixes `RecordScanner.refreshJwt` to POST to the correct Provable JWT endpoint (`https://api.provable.com/jwts/:consumerId`) instead of deriving it from the configured scanner base URL, preventing 404s during auth.
> 
> Also normalizes the API key header literal formatting in that request.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 58f9a7b321dab6434c2720c2a3d77c2f1236ee47. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #1256, opened 2026-03-23
[Fix] JWT refresh URL in RecordScanner and AleoNetworkClient

### LINKED ISSUE #1247: [Bug] RecordScanner auth fails due to wrong JWT endpoint path
## 🐛 Bug Report

RecordScanner fails to authenticate because it uses the wrong JWT endpoint.

**Correct endpoint:** `https://api.provable.com/jwts/:consumerId`

**Current endpoint used by RecordScanner:** `https://api.provable.com/scanner/:network/jwts/:consumerId`

The wrong path appears in `sdk/src/record-scanner.ts` here:
https://github.com/ProvableHQ/sdk/blob/9ea80c3401ad10dce530276c21c869c8ee342b7b/sdk/src/record-scanner.ts#L277

### Steps to Reproduce
1. Instantiate `RecordScanner`.
2. Attempt to issue a JWT / authenticate.
3. The request goes to the `scanner/:network/jwts/:consumerId` path and fails (HTTP 404).

### Expected Behavior
RecordScanner should call the non-networked JWT endpoint `https://api.provable.com/jwts/:consumerId`, allowing authentication to succeed.

marshacb:
## Motivation

`RecordScanner.refreshJwt()` appended `/jwts/{consumerId}` to `this.url`, which already included the scanner path (e.g. `https://api.provable.com/scanner/testnet`), sending JWT refresh requests to the wrong endpoint. `AleoNetworkClient.refreshJwt()` had a similar issue and since it hardcoded `https://api.provable.com/jwts/` instead of deriving from the user-configured host, breaking self-hosted and staging environments.

Both are fixed by extracting `baseUrl = new URL(host).origin` at construction time and using it for JWT refresh.

Additionally, CI was missing the KONG_API_KEY, CONSUMER_ID, and RECORD_SCANNER_URL env vars needed for the record scanner integration tests to authenticate, and the real API tests now register the view key in a before hook so they pass on both testnet and mainnet builds.

Fixes #1247

## Test Plan

- Added unit test for RecordScanner: verifies JWT refresh hits `{origin}/jwts/{consumerId}`, not `{origin}/scanner/{network}/jwts/{consumerId}`
- Added unit test for AleoNetworkClient: verifies JWT refresh derives URL from the configured host, not from a hardcoded `api.provable.com`
- [x] Both tests pass
## Related PRs

#1248

### DIFF
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -12,6 +12,8 @@ env:
   PUZZLE_PK: ${{ secrets.PUZZLE_PK }}
   PUZZLE_VK: ${{ secrets.PUZZLE_VK }}
   RECORD_SCANNER_URL: ${{ secrets.RECORD_SCANNER_URL }}
+  KONG_API_KEY: ${{ secrets.ALEO_DPS_API_KEY }}
+  CONSUMER_ID: ${{ secrets.ALEO_CONSUMER_ID }}
 
 
 jobs:
--- a/sdk/src/record-scanner.ts
+++ b/sdk/src/record-scanner.ts
@@ -111,6 +111,7 @@ export interface RecordScannerOptions {
 class RecordScanner implements RecordProvider {
     readonly cacheViewKeysOnRegister?: boolean;
     readonly url: string;
+    private readonly baseUrl: string;
     private apiKey?: { header: string, value: string };
     private consumerId?: string;
     private jwtData?: RecordScannerJWTData;
@@ -133,6 +134,9 @@ class RecordScanner implements RecordProvider {
         }
 
         // Configure the url to use the network the SDK is using.
+        // baseUrl is the API root (origin only) — used for JWT refresh which
+        // lives at /jwts/{consumerId} on the API root, not under /scanner.
+        this.baseUrl = new URL(options.url).origin;
         this.url = options.url + network;
 
         // Get any view keys passed in the options.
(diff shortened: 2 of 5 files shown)


sdk/src/record-scanner.ts
```diff
@@ -133,6 +134,9 @@ class RecordScanner implements RecordProvider {
         }
 
         // Configure the url to use the network the SDK is using.
+        // baseUrl is the API root (origin only) — used for JWT refresh which
+        // lives at /jwts/{consumerId} on the API root, not under /scanner.
+        this.baseUrl = new URL(options.url).origin;
```
iamalwaysuncomfortable, line 139: This *should* work, but write some edge cases around it just to make sure for both RSS and the Delegated Proving method.

marshacb: Makes sense. Ill add edge case tests for both RSS and the Delegated Proving method covering standard production URLs (custom hosts, ports, trailing slashes, and localhost)

.github/workflows/sdk.yml
```diff
@@ -12,6 +12,8 @@ env:
   PUZZLE_PK: ${{ secrets.PUZZLE_PK }}
   PUZZLE_VK: ${{ secrets.PUZZLE_VK }}
   RECORD_SCANNER_URL: ${{ secrets.RECORD_SCANNER_URL }}
+  KONG_API_KEY: ${{ secrets.ALEO_DPS_API_KEY }}
+  CONSUMER_ID: ${{ secrets.ALEO_CONSUMER_ID }}
```
iamalwaysuncomfortable, line 16: We need to add these to dependabot as well

marshacb: I should see dependabot under repo settings near `actions` right? (for adding secrets to it. If so I dont seem to have permissions for it)

iamalwaysuncomfortable: You're right, I'll add the secrets!

iamalwaysuncomfortable (approved):

PR #1258, opened 2026-03-25
[Feature] Export Value type with serialization methods

marshacb:
## Motivation

Re-export the `Value` type from `snarkVM` through the WASM layer to the TypeScript SDK. `Value` is the fundamental enum representing program function `inputs/outputs` (`Plaintext`, `Record`, `Future`). Previously only available internally as `ValueNative`. This exposes it to JS consumers with full serialization support.

## Test Plan

 - Rust wasm_bindgen_test tests: string/bytes roundtrip for all 3 variants, variant type checking, Plaintext/Record extraction and construction, error paths, bits/fields serialization
 - TS tests: same coverage across the WASM bridge including fromPlaintext/fromRecordPlaintext construction paths
 - [x] yarn build:all compiles cleanly
 - [x] All Value tests pass on both testnet and mainnet network variants

### DIFF
--- a/sdk/src/browser.ts
+++ b/sdk/src/browser.ts
@@ -149,6 +149,7 @@ export {
     U32,
     U64,
     U128,
+    Value,
     VerifyingKey,
     ViewKey,
     initThreadPool,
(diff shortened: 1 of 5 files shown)


iamalwaysuncomfortable (approved):

iamalwaysuncomfortable:
Excellent work! This is good to merge.

PR #1262, opened 2026-03-25
[Feature] Add ProgramImportsBuilder for zero-cost proving/verifying

### LINKED ISSUE #1212: [Feature] Allow extraction of proving and verifying keys during execution
## Motivation

SnarkVM is built in a way that assumes that Proving Keys and Verifying Keys will be stored in a long running process. However in most execution contexts called by chain users, the VM is launched in an ephemeral process and destroyed.

Thus if the user does NOT pass in the Proving Key, execution is much slower because the key must be synthesized.

In many cases
- Users forgo passing the Proving Keys
- Users do NOT know how to generate proving keys

To alleviate this disconnect, the Proving Keys should be returned by the SDK. Users should have the option to do the following.
- Get a proving key and verifying keys from the execution methods manually in case they have their own custom key storage method
- Use a configured KeyProvider/Keystore object in the SDK to automatically collect and store the keys (requiring little to no user intervention)

## Implementation

1. The buildExecutionTransaction method in both the `wasm` and `js` version of the SDK should optionally return the ProvingKey and VerifyingKey.
2. There should be configuration options for the SDK that insert the keys into whatever `KeyProvider` is configured so that users automatically get the keys without needing to use any explicit imperative logic.

marshacb:
## Motivation

Programs that import other programs (e.g., `double_test.aleo` importing `multiply_test.aleo`) currently require the SDK to re-synthesize proving and verifying keys for every imported function on each execution. Key synthesis dominates execution time and for programs with import chains, the first execution can take 20-30s while a cache-hit execution takes 1-3s.

`ProgramImportsBuilder` (TS) / `ProgramImports` (Rust WASM) solves this with a `ProcessNative`-backed builder that owns a single `snarkVM` `Process` instance. Programs and their keys are loaded into this process once, then the entire process is handed to the WASM execution layer allowing for no duplicate key storage and no re-synthesis. The builder flows through execution and is returned enriched with any keys that were synthesized, enabling a `cache-on-first-use` pattern.

On the TS side, `ProgramManager` integrates this with `KeyStore` (filesystem via `LocalFileKeyStore`, browser via the new `IndexedDBKeyStore`). `buildProgramImports` resolves the full import graph via BFS, loads cached keys from the store, and passes the populated builder to WASM. After execution, `persistExtractedKeys` writes any newly synthesized keys back to the store. On subsequent calls, all keys load from cache and synthesis is skipped entirely.

The node loyalty program template demonstrates the full pipeline: mint_card drops from ~8s to ~1s, cross-program redeem from ~22s to ~3s on second execution.

## Test Plan

- WASM builder CRUD tests: new(), addProgram, contains, getProgram, programNames, isEmpty, fromObject (plain + structured), toObject
- BFS import resolution tests: no imports, user-provided, network fetch, merge (user wins), network error resilience, transitive resolution (3-level chain), BFS dedup, parallel sibling fetching via getProgram
- Call chain filtering tests: only loads keys for called functions, empty programs, multi-program calls
- loadKeysFromStore tests: KeyStore queries with KeyLocator, no-op without store, _keyStore priority over keyProvider.keyStore(), error swallowing, mid-iteration failure recovery, edition pass-through (including edition=0)
- `persistExtractedKeys` tests: write to store, skip existing keys, no-op without store, error swallowing, multi-program with partial failure, partial key availability (PK without VK)
- `resolveTopLevelKeys` tests: KeyStore priority, KeyProvider fallback, graceful failure, edition pass-through
- `setKeyStore` tests: field assignment, precedence over provider
- `KeyLocator` factories tests: correct construction for proving/verifying locators, edition defaults to 1
- Key bytes API tests: error paths (program not added, invalid bytes), real execution round-trip with synthesized keys, toObject/fromObject round-trip, addProvingKeyBytes/addVerifyingKeyBytes with real bytes, idempotent insertion
- `OfflineExecutionResult` take semantic tests: getResponse() throws on second call, getProgramImports() returns undefined without builder, returns undefined on second call
- `getImportNames` regex tests: single/none/multiple imports, cross-validation against WASM Program.fromString().getImports()
- Edition handling tests: pass-through, defaults, 0-is-valid
- CI: Added `yarn start:caching` step to the loyalty program local job to validate the full `ProgramImportsBuilder` + `LocalFileKeyStore` pipeline in CI.

 
## Related PRs

#1208 

---

> [!NOTE]
> **High Risk**
> High risk because it refactors `ProgramManager` execution/authorization/proving-request flows to route import resolution and proving/verifying key lookup/persistence through a new `ProgramImportsBuilder` + `KeyStore` pipeline, plus adds new network calls for edition/amendment metadata.
> 
> **Overview**
> Introduces a new **import-aware key-caching pipeline** in `ProgramManager`: it now builds a `ProgramImportsBuilder` for a program’s full transitive import graph, preloads any cached proving/verifying keys from a `KeyStore`, passes the builder into WASM execution paths, and then persists any newly synthesized keys back to the store.
> 
> Adds browser persistence via a new `IndexedDBKeyStore` implementation and exposes it through `sdk/src/browser.ts`; Node and React templates are updated to demonstrate/enable caching (`LocalFileKeyStore` / `IndexedDBKeyStore`), with `.aleo` ignored.
> 
> Extends `AleoNetworkClient` with `getProgramAmendmentCount` and updates edition handling throughout `ProgramManager` to resolve `{edition, amendment}` and use them when locating/storing keys. CI is updated to run a new key-caching demo for the loyalty template and adds a new dynamic-dispatch/deep-chain integration test template job.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 014002e0788f161715761c4dc3c4dbb40ec6aa53. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sdk/src/program-manager.ts
```diff
@@ -162,6 +169,7 @@ interface ExecuteAuthorizationOptions {
     offlineQuery?: OfflineQuery;
     program?: string | Program;
     imports?: ProgramImports;
+    programImportsBuilder?: ProgramImportsBuilder;
```
```suggestion
```

We should handle this inside the ProgramManager

sdk/src/program-manager.ts
```diff
@@ -304,6 +313,184 @@ class ProgramManager {
...
+                }
+                if (await keyStore.has(vkLocator)) {
+                    const vk = await keyStore.getVerifyingKey(vkLocator);
+                    if (vk) builder.addVerifyingKey(programName, fnName, vk);
+                }
+            } catch (e) {
+                console.debug(`Failed to load keys for ${programName}/${fnName}: ${e}`);
+            }
+        }
+    }
```
iamalwaysuncomfortable, line 646: We don't want to add all the proving/verifying keys from a program as we'd likely run out of memory.

We only want to add the keys functions we know are in the call chain.

sdk/src/program-manager.ts
```diff
@@ -304,6 +313,184 @@ class ProgramManager {
...
+
+        const programNames: string[] = Array.from(builder.programNames());
+        for (const name of programNames) {
+            const source = builder.getProgram(name);
+            if (!source) continue;
+
+            const program = Program.fromString(source);
+            const functions: string[] = Array.from(program.getFunctions());
+
+            for (const fnName of functions) {
```
iamalwaysuncomfortable: Basically same thing here, we only want to extract the functions we've derived proving and verifying keys for.

sdk/src/program-manager.ts
```diff
@@ -879,29 +1066,18 @@ class ProgramManager {
         }
         const [feeProvingKey, feeVerifyingKey] = feeKeys;
 
-        // If the function proving and verifying keys are not provided, attempt to find them using the key provider
+        // Build the ProgramImportsBuilder if not provided.
+        const programImportsBuilder = options.programImportsBuilder
+            ?? await this.buildProgramImports(program, imports);
```
iamalwaysuncomfortable: We generally shouldn't allow people to pass this in from the outside.

iamalwaysuncomfortable (changes requested):
I think for the immediate release we want to hold off on merging this changeset as it's large, but we should target to have this in by March 31st.

sdk/src/keys/keystore/indexeddb.ts
```diff
@@ -0,0 +1,255 @@
...
+                const db = request.result;
+                if (!db.objectStoreNames.contains(this.storeName)) {
+                    db.createObjectStore(this.storeName, { keyPath: "locator" });
+                }
+            };
+            request.onsuccess = () => resolve(request.result);
+            request.onerror = () => reject(request.error);
+        });
+        return this.dbPromise;
+    }
```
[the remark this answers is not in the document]
marshacb: updated to clear cached promise on rejection for retry

sdk/src/program-manager.ts
```diff
@@ -332,6 +344,284 @@ class ProgramManager {
...
+                    const pk = builder.getProvingKey(programName, fnName);
+                    const vk = builder.getVerifyingKey(programName, fnName);
+                    if (pk && vk) {
+                        await keyStore.setKeys(pkLocator, vkLocator, [pk, vk]);
+                    }
+                }
+            } catch (e) {
+                console.debug(`Failed to persist keys for ${programName}: ${e}`);
+            }
+        }
```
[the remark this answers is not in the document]
marshacb: moved try/catch to per-function instead of per-program

sdk/src/program-manager.ts
```diff
@@ -332,6 +344,284 @@ class ProgramManager {
...
+                if (collected.has(dep)) visit(dep);
+            }
+            sorted.push([name, source]);
+        };
+        for (const [name] of collected) visit(name);
+
+        for (const [name, source] of sorted) {
+            if (builder.contains(name)) continue;
+            try {
+                builder.addProgram(name, source, edition);
```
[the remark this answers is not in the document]
marshacb: Per-import edition resolution would require N additional network calls for no practical benefit: edition doesn't affect key synthesis or proof correctness, and the only impact is key cache locator accuracy for non-zero edition imports. Choosing not to file now. Can revisit if multi-edition programs become common.

sdk/src/program-manager.ts
```diff
@@ -332,6 +344,290 @@ class ProgramManager {
...
+                const subImports = ProgramManager.getImportNames(source);
+                if (subImports.length > 0) {
+                    const transitiveCalls = ProgramManager.getCalledFunctions(source);
+                    for (const [tProgram, tFns] of transitiveCalls) {
+                        if (!calledFunctions.has(tProgram)) {
+                            calledFunctions.set(tProgram, tFns);
+                        } else {
+                            for (const fn of tFns) calledFunctions.get(tProgram)!.add(fn);
+                        }
+                    }
```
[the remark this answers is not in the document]
marshacb: Leaf programs have no import statements, so they can't have call program.aleo/function instructions. Aleo requires imports before external calls. The guard is correct. The top-level program's calls are scanned separately before the BFS loop.

sdk/src/program-manager.ts
```diff
@@ -332,6 +344,290 @@ class ProgramManager {
...
+        edition?: number,
+    ): Promise<void> {
+        const keyStore = await this.resolveKeyStore();
+        if (!keyStore) return;
+
+        const functions = functionNames ?? Array.from(Program.fromString(programSource).getFunctions());
+
+        for (const fnName of functions) {
+            const pkLocator = provingKeyLocator(programName, fnName, edition);
+            const vkLocator = verifyingKeyLocator(programName, fnName, edition);
```
iamalwaysuncomfortable: This uses the default edition and amendment, these should check what the current edition and amendment are before searching for the proving key.

Here's the endpoint that will return both the edition and amendment count.
https://api.provable.com/v2/mainnet/programs/:programID/amendment_count

wasm/src/programs/execution.rs
```diff
@@ -126,17 +126,26 @@ pub fn verify_function_execution(
...
+    // Use ProgramImports' internal process when available, otherwise load fresh.
+    let mut owned_process;
+    let mut program_imports = program_imports;
+    let process: &mut ProcessNative = match program_imports.as_mut() {
+        Some(pi) => pi.process_mut(),
+        None => {
+            owned_process = ProcessNative::load_web().map_err(|e| e.to_string())?;
+            ProgramManager::resolve_imports(&mut owned_process, imports, Some(program_id.as_str()))?;
+            &mut owned_process
+        }
```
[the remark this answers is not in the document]
marshacb: By design `ProgramImportsBuilder` is pre-populated with all imports by `buildProgramImports()` on the TS side before reaching WASM. Running `resolve_imports` on it would be redundant and risk double-adding programs. The imports Object is the legacy fallback path, only used when no builder is provided.

sdk/src/program-manager.ts
```diff
@@ -332,6 +344,290 @@ class ProgramManager {
...
+            if (builder.contains(name)) continue;
+            try {
+                builder.addProgram(name, source, edition);
+            } catch (e) {
+                console.warn(`Failed to add import ${name} to builder: ${e}`);
+                continue;
+            }
+
+            const calledFns = calledFunctions.get(name);
+            await this.loadKeysFromStore(builder, name, source, calledFns ? [...calledFns] : [], edition);
```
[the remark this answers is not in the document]
marshacb: Intentional since `[]` here means "no functions are called from this import, so load zero keys." Loading all keys for uncalled imports wastes memory. If the functions are actually needed at execution time, WASM synthesizes them on demand.

sdk/src/program-manager.ts
```diff
@@ -332,6 +344,290 @@ class ProgramManager {
...
+            try {
+                fns = Array.from(builder.functionKeysAvailable(programName));
+            } catch (e) {
+                console.debug(`Failed to query keys for ${programName}: ${e}`);
+                continue;
+            }
+            for (const fnName of fns) {
+                try {
+                    const pkLocator = provingKeyLocator(programName, fnName, edition);
+                    const vkLocator = verifyingKeyLocator(programName, fnName, edition);
```
iamalwaysuncomfortable: We have to account for amendment here too.

sdk/src/program-manager.ts
```diff
@@ -332,6 +344,290 @@ class ProgramManager {
...
+        return builder;
+    }
+
+    /**
+     * Extract `import program_name.aleo;` names from program source via regex.
+     * Avoids a WASM round-trip compared to Program.fromString + getImports.
+     */
+    private static getImportNames(programSource: string): string[] {
+        const result: string[] = [];
+        const importPattern = /^import\s+([a-zA-Z_][a-zA-Z0-9_]*\.aleo)\s*;/gm;
```
iamalwaysuncomfortable: Maybe this would be more robust to allow whitespace ahead of it.

`/^\s*import\s+([a-zA-Z_][a-zA-Z0-9_]*\.aleo)\s*;/gm`

sdk/src/program-manager.ts
```diff
@@ -332,6 +344,290 @@ class ProgramManager {
...
+        builder: ProgramImportsBuilder,
+        programName: string,
+        programSource: string,
+        functionNames?: string[],
+        edition?: number,
+    ): Promise<void> {
+        const keyStore = await this.resolveKeyStore();
+        if (!keyStore) return;
+
+        const functions = functionNames ?? Array.from(Program.fromString(programSource).getFunctions());
```
iamalwaysuncomfortable: We don't want to blanket add proving and verifying keys for all functions that we have, the calling logic should be able to construct the call non-dynamic-dispatch call graph.
```suggestion
        const functions = functionNames;
```

iamalwaysuncomfortable: I understand in dynamic dispatch functions, we don't know the call graph and without this we have to rely on synthesis, but I think this is ultimately more stable behavior to allow synthesis rather than potentially loading many very large proving keys.

Figuring out the exact dynamic call graph is something we should address in a more deterministic fashion later.

wasm/src/programs/manager/deploy.rs
```diff
@@ -81,17 +81,26 @@ impl ProgramManager {
...
+
+        let process: &mut ProcessNative = match program_imports.as_mut() {
+            Some(pi) => pi.process_mut(),
+            None => {
+                owned_process = ProcessNative::load_web().map_err(|err| err.to_string())?;
+                ProgramManager::resolve_imports(&mut owned_process, imports, Some(program_id.as_str()))?;
+                &mut owned_process
+            }
+        };
+
```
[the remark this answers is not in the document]
marshacb: same as above

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,528 @@
...
+    /// ```
+    ///
+    /// Programs created via this method default to edition 0 (initial deployment).
+    ///
+    /// @param {Object} object A plain JavaScript object mapping program names to source code
+    ///   and optional keys.
+    /// @returns {ProgramImports}
+    #[wasm_bindgen(js_name = "fromObject")]
+    pub fn from_object(object: Object) -> Result<ProgramImports, String> {
+        let mut imports = Self::new()?;
```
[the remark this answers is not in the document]
marshacb: Doc updated. fromObject defaults to edition 1 (via add_program(..., None) → unwrap_or(1)). The structured format doesn't include an edition field by design. Edition is managed by the caller via addProgram(name, source, edition). Adding edition to the serialized format is a potential future enhancement.

wasm/src/programs/manager/execute.rs
```diff
@@ -96,19 +148,29 @@ impl ProgramManager {
...
+                pi.process_mut()
+            }
+            None => {
+                owned_process = ProcessNative::load_web().map_err(|err| err.to_string())?;
+                log("Check program imports are valid and add them to the process");
+                program_native = ProgramNative::from_str(program).map_err(|e| e.to_string())?;
+                ProgramManager::resolve_imports(&mut owned_process, imports, Some(&program_native.id().to_string()))?;
+                &mut owned_process
+            }
+        };
```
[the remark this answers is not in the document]
marshacb: same as above

create-leo-app/template-node-loyalty-program-ts/.gitignore
```diff
@@ -0,0 +1 @@
+.aleo
```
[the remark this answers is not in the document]
marshacb: Acknowledged. This is the only template with both files. It's needed here because CI runs yarn start:caching which generates .aleo/ cache files in the template directory. Other templates don't produce build artifacts during CI. The `_gitignore` handles scaffolded projects; the `.gitignore` protects the SDK repo.

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,528 @@
...
+
+        if !self.process.contains_program(&program_id) {
+            // Resolve static imports depth-first before adding this program.
+            self.resolve_program_imports(&program)?;
+            log(&format!("Adding {program_id} to the process"));
+            self.process.add_program_with_edition(&program, edition).map_err(|e| e.to_string())?;
+        }
+
+        self.sources.insert(program_id.clone(), source.to_string());
+        self.editions.insert(program_id, edition);
```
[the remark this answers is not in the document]
marshacb: Fixed. `addProgram` now validates `program.id() == program_id` and rejects mismatches.

wasm/src/programs/manager/mod.rs
```diff
@@ -181,31 +187,25 @@ impl ProgramManager {
...
+                }
+                if !process.contains_program(import.id()) {
+                    log(&format!("Importing program: {import_id_str}"));
+                    Self::resolve_program_imports(process, &import, imports)?;
+                    log(&format!("Adding {import_id_str} to the process"));
+                    // Edition 1 is required: since ConsensusVersion::V8, Authorization::check_valid_edition
+                    // rejects edition 0 for non-constructor programs.
+                    process.add_program_with_edition(&import, 1).map_err(|err| err.to_string())?;
+                }
             }
```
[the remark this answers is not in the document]
marshacb: Fixed. Missing imports now return an explicit error instead of silently skipping.

wasm/src/programs/manager/execute.rs
```diff
@@ -96,19 +148,29 @@ impl ProgramManager {
...
+
+        // Use ProgramImports' internal process when available, otherwise load fresh.
+        let mut owned_process;
+        let program_native;
+        let process: &mut ProcessNative = match program_imports.as_mut() {
+            Some(pi) => {
+                program_native = pi.ensure_program(program, edition)?;
+                pi.process_mut()
+            }
+            None => {
```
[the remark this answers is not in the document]
marshacb: Fixed so all edition defaults are now `unwrap_or(1)`, aligned with mainnet's `check_valid_edition` requirements.

sdk/src/program-manager.ts
```diff
@@ -972,23 +1256,39 @@ class ProgramManager {
...
             provingKey,
             verifyingKey,
             feeProvingKey,
             feeVerifyingKey,
             offlineQuery,
-            edition
+            edition,
+            programImportsBuilder,
         );
+        const transaction = result.getTransaction();
```
iamalwaysuncomfortable: We want to to avoid changing the return signature here, some people depend on the raw `wasm` package so we don't want to break that API for them.

I know WHY this is done (because we're passing ownership and getting an object back out) - but we should avoid changing the return signature of the core wasm function.

Instead what we could do is make the interior of ProgramImports wrapped with interior mutability, i.e. we wrap the functions with something like Arc<RwLock<Process>>, etc. - that way we can
1. Clone it, send it into the execution
2. During execution and have the interior data structure be modified
3. Allow the cloned object to be cleaned up, and our outer copy now contains all the keys we need.

Proceed as normal as you do below!

sdk/src/program-manager.ts
iamalwaysuncomfortable: Response is 
```typescript
interface AmendmentResponse {
  program_id: string;
  edition: number;
  amendment_count: number; // The amendment count for the specific edition.
}
```

iamalwaysuncomfortable (changes requested):
Thanks for the work on this 

Going through this a few things stood out.
1. We shouldn't change the return signature of buildExecutionTransaction, we can avoid doing so by using internal mutability in the ProgramImports object in rust
2. There shouldn't really be a world we attempt to import all proving keys for a function, I understand right now this leads us to a situation where dynamic imports will need to be synthesized, but that's okay for now as the tradeoff of importing all proving keys could cause some issues.
3. Need to account for the amendments in the calls.
4. Minor nit on import parsing.

sdk/src/program-manager.ts
marshacb: Agreed. I removed the `getFunctions()` fallback. `loadKeysFromStore` now only loads keys for explicitly identified functions in the call graph so dynamic dispatch targets synthesize on demand.

wasm/src/programs/manager/execute.rs
```diff
@@ -198,102 +243,45 @@ impl ProgramManager {
...
-
-        log("Check program imports are valid and add them to the process");
-        let program_native = ProgramNative::from_str(program).map_err(|e| e.to_string())?;
-        let program_id = program_native.id().to_string();
-        ProgramManager::resolve_imports(process, imports, Some(program_id.as_str()))?;
-        let rng = &mut StdRng::from_entropy();
-
-        log(&format!("Executing function: {program_id}/{function} on-chain"));
+        program_imports: Option<ProgramImports>,
+    ) -> Result<ExecutionResult, String> {
```
iamalwaysuncomfortable: We want to keep this signature the same (it should return the transaction type)

iamalwaysuncomfortable (commented):
## Edition doc/code mismatch

The code correctly defaults edition to `1` everywhere (`edition.unwrap_or(1)`), since edition 0 is rejected by `Authorization::check_valid_edition` since ConsensusVersion::V8 for non-constructor programs. But several doc comments say "Defaults to 0".

Two of the four are in diff context lines that can't receive inline suggestions, so noting them here:

- **`execute.rs:240`** (`buildExecutionTransaction`): doc says "or 0 if the program does not exist" → should say "or 1"
- **`execute.rs:369`** (`buildDevnodeExecutionTransaction`): same fix
- **`execute.rs:493`** (`estimateExecutionFee`): doc says "Defaults to 0" → should say "Defaults to 1"

The other two (`execute.rs:305` and `imports.rs:111`) have inline suggestions below.

## `fromObject` error handling

See inline comment on `imports.rs:111`.

wasm/src/programs/manager/execute.rs
```diff
@@ -316,154 +305,45 @@ impl ProgramManager {
         imports: Option<Object>,
```
iamalwaysuncomfortable, line 323: Doc says "defaults to 0" but code uses `edition.unwrap_or(1)` at line 315.

```suggestion
    /// @param edition The program edition (defaults to 1).
```

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,572 @@
...
+    /// Accepts three formats:
+    /// ```js
+    /// // 1. Plain string — source code only.
+    /// { "my_program.aleo": "program source..." }
+    ///
+    /// // 2. Structured — program source with optional edition.
+    /// { "my_program.aleo": { program: "program source..." } }
+    ///
+    /// // 3. Structured with keys — program source plus proving/verifying keys per function.
+    /// {
```
iamalwaysuncomfortable, line 111: If `add_program` fails here, the error is silently swallowed — the caller gets back a `ProgramImports` with missing programs and no indication anything went wrong. For a key caching feature where correctness matters, this could lead to missing keys and unexpected re-synthesis.

Suggested fix — propagate the error:

```suggestion
            imports.add_program(&name, &source, None).map_err(|e| {
                format!("Failed to add program '{}' from object: {}", name, e)
            })?;
```

wasm/src/programs/manager/execute.rs
```diff
@@ -96,20 +132,33 @@ impl ProgramManager {
         url: Option<String>,
         offline_query: Option<OfflineQuery>,
         edition: Option<u16>,
-    ) -> Result<ExecutionResponse, String> {
+        program_imports: Option<ProgramImports>,
+    ) -> Result<OfflineExecutionResult, String> {
```
iamalwaysuncomfortable: The return type here should not change.

wasm/src/programs/manager/execute.rs
```diff
@@ -155,7 +204,7 @@ impl ProgramManager {
             execution_response.add_proving_key(process, function, program_native.id())?;
         }
 
-        Ok(execution_response)
+        Ok(OfflineExecutionResult { response: Some(execution_response) })
```
iamalwaysuncomfortable: Same here this should return its original type, we can't change the original return types on any of these methods.

wasm/src/programs/manager/execute.rs
iamalwaysuncomfortable: We don't want to change the types returned by wasm, the return signature should stay the same, so this should return the `Transaction` type.

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,572 @@
...
+impl ProgramImports {
+    /// Create a cheap clone that shares the same underlying data.
+    ///
+    /// Useful for passing to WASM execution functions which consume ownership:
+    /// the caller keeps the original, and both copies see any mutations
+    /// (e.g. synthesized keys) through the shared interior state.
+    #[wasm_bindgen(js_name = "clone")]
+    pub fn js_clone(&self) -> ProgramImports {
+        self.clone()
+    }
```
iamalwaysuncomfortable: This can just be clone() - and we can override manage the clippy with a lint override.
```suggestion
    #[allow(clippy::should_implement_trait)]
    pub fn clone(&self) -> ProgramImports {
        self.clone()
    }
```

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,572 @@
...
+    ///   and optional keys.
+    /// @returns {ProgramImports}
+    #[wasm_bindgen(js_name = "fromObject")]
+    pub fn from_object(object: Object) -> Result<ProgramImports, String> {
+        let mut imports = Self::new()?;
+        let keys = Object::keys(&object);
+        for i in 0..keys.length() {
+            let Some(name) = keys.get(i).as_string() else {
+                continue;
+            };
```
iamalwaysuncomfortable: This should probably fail if it's not a string with an error that program names should be a string.

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,572 @@
...
+    /// @param {ProvingKey} key The proving key.
+    #[wasm_bindgen(js_name = "addProvingKey")]
+    pub fn add_proving_key(&mut self, program_name: &str, identifier: &str, key: ProvingKey) -> Result<(), String> {
+        let program_id = ProgramIDNative::from_str(program_name).map_err(|e| e.to_string())?;
+        let fn_id = IdentifierNative::from_str(identifier).map_err(|e| e.to_string())?;
+        let inner = self.inner.borrow_mut();
+        if !inner.process.contains_program(&program_id) {
+            return Err(format!("Program '{program_name}' must be added via addProgram before adding keys"));
+        }
+        let has_pk = inner.process.get_stack(&program_id).map_or(false, |stack| stack.contains_proving_key(&fn_id));
```
iamalwaysuncomfortable, line 200: To note we need to ensure that in using this builder pattern externally we:

1. We check what the latest edition was on the network.
2. If we get a key, we ENSURE we query the latest edition, if we insert a previous edition's key from the keystore, it will fail, so the outer call in program manager must account for this.

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,572 @@
...
+    /// @param {VerifyingKey} key The verifying key.
+    #[wasm_bindgen(js_name = "addVerifyingKey")]
+    pub fn add_verifying_key(&mut self, program_name: &str, identifier: &str, key: VerifyingKey) -> Result<(), String> {
+        let program_id = ProgramIDNative::from_str(program_name).map_err(|e| e.to_string())?;
+        let fn_id = IdentifierNative::from_str(identifier).map_err(|e| e.to_string())?;
+        let inner = self.inner.borrow_mut();
+        if !inner.process.contains_program(&program_id) {
+            return Err(format!("Program '{program_name}' must be added via addProgram before adding keys"));
+        }
+        let has_vk = inner.process.get_stack(&program_id).map_or(false, |stack| stack.contains_verifying_key(&fn_id));
```
iamalwaysuncomfortable, line 227: Same comment here, we need to ensure the outer KeyStore cache hit enforces the edition check.

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,572 @@
...
+    /// @param {Uint8Array} bytes The proving key bytes.
+    #[wasm_bindgen(js_name = "addProvingKeyBytes")]
+    pub fn add_proving_key_bytes(&mut self, program_name: &str, identifier: &str, bytes: &[u8]) -> Result<(), String> {
+        let program_id = ProgramIDNative::from_str(program_name).map_err(|e| e.to_string())?;
+        let fn_id = IdentifierNative::from_str(identifier).map_err(|e| e.to_string())?;
+        let inner = self.inner.borrow_mut();
+        if !inner.process.contains_program(&program_id) {
+            return Err(format!("Program '{program_name}' must be added via addProgram before adding keys"));
+        }
+        let has_pk = inner.process.get_stack(&program_id).map_or(false, |stack| stack.contains_proving_key(&fn_id));
```
iamalwaysuncomfortable, line 254: Ditto here.

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,572 @@
...
+        identifier: &str,
+        bytes: &[u8],
+    ) -> Result<(), String> {
+        let program_id = ProgramIDNative::from_str(program_name).map_err(|e| e.to_string())?;
+        let fn_id = IdentifierNative::from_str(identifier).map_err(|e| e.to_string())?;
+        let inner = self.inner.borrow_mut();
+        if !inner.process.contains_program(&program_id) {
+            return Err(format!("Program '{program_name}' must be added via addProgram before adding keys"));
+        }
+        let has_vk = inner.process.get_stack(&program_id).map_or(false, |stack| stack.contains_verifying_key(&fn_id));
```
iamalwaysuncomfortable, line 283: Ditto here/

sdk/src/program-manager.ts
iamalwaysuncomfortable: Same comment, the above should just return a `Transaction`.

wasm/src/programs/manager/imports.rs
```diff
@@ -0,0 +1,572 @@
...
+
+        let mut inner = self.inner.borrow_mut();
+        if !inner.process.contains_program(&program_id) {
+            // Resolve static imports depth-first before adding this program.
+            Self::resolve_program_imports_inner(&mut inner, &program)?;
+            log(&format!("Adding {program_id} to the process"));
+            inner.process.add_program_with_edition(&program, edition).map_err(|e| e.to_string())?;
+        }
+
+        inner.sources.insert(program_id.clone(), source.to_string());
```
iamalwaysuncomfortable, line 179: Will we capture all sources from the imports the user passes in the imports object (via dynamic dispatch)? I think so right?

iamalwaysuncomfortable (commented):
Few more comments

wasm/src/programs/manager/imports.rs
marshacb: Yes, when the TS `buildProgramImports` method resolves imports, it calls `builder.addProgram(name, source, edition)` for each import. That method inserts into `inner.sources` (line 181 of imports.rs). And `prepare_for_execution` also inserts the top-level program's source.

iamalwaysuncomfortable (commented):
This is almost there, but we still need to write a few tests that test this out.

In a node.js create-leo-app we should use a KeyStore and do the following:

**Ensure resolution of keys for dynamic dispatch programs where imports must be specified** 
* Execute a dynamic dispatch program where we need to specify the inputs and see if all the resulting keys get synthesized and stored when there's a keystore present
* Ensure that keys are loaded when we do pass a KeyStore for all functions we pass in dynamic dispatch imports


**Ensure resolution of keys for programs with deeply nested static imports**
* Execute a program with a LOT of static imports and ensure the keys get synthesized and stored
* Run the execution again and ensure all keys are loaded from disk and ensure no out of memory executions happen with deep exports.

We should bench the two approaches above and post the numbers as well.

sdk/src/program-manager.ts
```diff
@@ -873,14 +1214,9 @@ class ProgramManager {
...
-            try {
-                edition = await this.networkClient.getLatestProgramEdition(programName);
-            } catch (e: any) {
-                console.warn(`Error finding edition for ${programName}. Network response: '${e.message}'. Assuming edition 0.`);
-                edition = 0;
-            }
-        }
+        const resolved = await this.resolveEditionAndAmendment(programName, edition);
+        edition = edition ?? resolved.edition;
+        const { amendment } = resolved;
```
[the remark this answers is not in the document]
marshacb: Acknowledged. Pre-existing pattern consistent across all call sites. The amendment endpoint doesn't support querying by edition, so there's no way to resolve this without an API design change. Will address separately.

sdk/src/program-manager.ts
```diff
@@ -873,14 +1220,9 @@ class ProgramManager {
...
```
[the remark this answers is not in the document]
marshacb: Intentional. `resolveEditionAndAmendment` is needed for amendment resolution (not just edition), which the old `getLatestProgramEdition` didn't provide. The amendment is required for correct key locator construction.

wasm/src/programs/manager/execute.rs
```diff
@@ -606,13 +459,28 @@ impl ProgramManager {
...
+                inner_guard = guard;
+                &mut inner_guard.process
+            }
+            None => {
+                owned_process = ProcessNative::load_web().map_err(|err| err.to_string())?;
+                program_native = ProgramNative::from_str(program).map_err(|e| e.to_string())?;
+                ProgramManager::resolve_imports(&mut owned_process, imports, Some(&program_native.id().to_string()))?;
+                &mut owned_process
+            }
+        };
```
iamalwaysuncomfortable: This is repeated in a lot of places, is there any way to encapsulate it into a convenience method?

wasm/src/programs/manager/deploy.rs
```diff
@@ -165,19 +179,35 @@ impl ProgramManager {
...
+                let (_, guard) = pi.prepare_for_execution(&program.to_string(), 1)?;
+                inner_guard = guard;
+                &mut inner_guard.process
+            }
+            None => {
+                owned_process = ProcessNative::load_web().map_err(|err| err.to_string())?;
+                ProgramManager::resolve_imports(&mut owned_process, imports, Some(program_id.as_str()))?;
+                &mut owned_process
+            }
+        };
```
iamalwaysuncomfortable: This seems repeated in a lot of places, any way we could encapsulate this in a helper method?

sdk/src/program-manager.ts
```diff
@@ -360,15 +360,14 @@ class ProgramManager {
     private async buildProgramImports(
         program: string | Program,
         imports?: ProgramImports,
-        edition?: number,
-    ): Promise<ProgramImportsBuilder> {
+    ): Promise<{ builder: ProgramImportsBuilder; importEditions: Map<string, { edition: number; amendment: number }> }> {
```
iamalwaysuncomfortable, line 408: One major thing to note is that proving and verifying keys don't get synthesized when constructing authorizations, the consequence of this is that when building the imports, we don't need to worry about retrieving or storing proving keys when using any method that builds authorizations and proving requests.

Keys are only used or synthesized when doing an actual local execution.

The only time we need to insert new keys is when when we're doing a full proof/execution. So we probably want to add a flag to this function which 
1. Has a mode that only fetches the actual program imports during proving request or authorization building.
2. Has another mode that fetches and stores the keys when doing a full execution or key synthesis run

wasm/src/programs/manager/execute.rs
marshacb: Great point! Extracted a `ResolvedProcess` enum that encapsulates both the builder-backed and fresh-process paths. Two constructors: `resolve()` for `execution/auth` (parses program from source) and `resolve_with_program()` for `deploy` (program already parsed). All 13 occurrences replaced.

sdk/src/program-manager.ts
marshacb: Good catch! `buildProgramImports` now accepts a `loadKeys` flag (defaults to true). `Auth/proving` paths pass `false` to skip key loading and don't call `persistExtractedKeys`. Execution paths (`buildExecutionTransaction`, `run`, `executeAuthorization`) keep the full key load + persist pipeline.

sdk/src/program-manager.ts
```diff
@@ -333,6 +343,350 @@ class ProgramManager {
...
+        fallbackEdition?: number,
+    ): Promise<{ edition: number; amendment: number }> {
+        try {
+            const info = await this.networkClient.getProgramAmendmentCount(programName);
+            return { edition: info.edition, amendment: info.amendment_count };
+        } catch (e: any) {
+            console.warn(`Error finding edition/amendment for ${programName}. Network response: '${e.message}'. Defaulting to edition ${fallbackEdition ?? 1}, amendment 0.`);
+            return { edition: fallbackEdition ?? 1, amendment: 0 };
+        }
+    }
```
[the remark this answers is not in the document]
marshacb: `check_valid_edition` in snarkVM (ConsensusVersion::V8+) rejects edition 0 for non-constructor programs. Default to 1 is intentional

sdk/src/program-manager.ts
```diff
@@ -333,6 +343,350 @@ class ProgramManager {
...
+
+        const importNames = programObj.getImports();
+        if (importNames.length === 0 && (!imports || Object.keys(imports).length === 0)) {
+            return { builder, importEditions: new Map() };
+        }
+
+        let resolvedImports: ProgramImports = {};
+        if (importNames.length > 0) {
+            try {
+                resolvedImports = await this.networkClient.getProgramImports(programSource);
```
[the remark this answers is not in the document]
marshacb: Not an issue. `getProgramImports()` accepts both `Program` | `string` and handles source code strings internally via `Program.fromString()`

sdk/src/program-manager.ts
```diff
@@ -1532,16 +1856,39 @@ class ProgramManager {
             );
         }
 
+        // Build ProgramImportsBuilder for import resolution (no key loading —
+        // proving requests don't synthesize keys).
+        // Note: imports is kept for the Rust-side fee estimation fallback
+        // (estimate_fee_for_authorization uses the legacy imports Object to avoid
+        // a RefCell double-borrow — see proving_request.rs).
+        const { builder } = await this.buildProgramImports(program, imports, false);
```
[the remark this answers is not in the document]
marshacb: buildAuthorizationUnchecked does call buildProgramImports (line 1713).

sdk/src/program-manager.ts
```diff
@@ -371,6 +381,350 @@ class ProgramManager {
...
+        while ((match = callPattern.exec(programSource)) !== null) {
+            const programName = match[1];
+            const functionName = match[2];
+            if (!result.has(programName)) {
+                result.set(programName, new Set());
+            }
+            result.get(programName)!.add(functionName);
+        }
+        return result;
+    }
```
iamalwaysuncomfortable, line 580: This seems like it'll resolve every function that has a call instruction, leading to proving keys from many more functions that we need potentially being loaded. What we're interested in is just the functions in the call chain, can we correct this to only find the functions in the direct call chain?

sdk/src/program-manager.ts
```diff
@@ -3511,14 +3863,8 @@ class ProgramManager {
...
-        if (edition == undefined) {
-            try {
-                edition = await this.networkClient.getLatestProgramEdition(programName);
-            } catch (e: any) {
-                console.warn(`Error finding edition for ${programName}. Network response: '${e.message}'. Assuming edition 0.`);
-                edition = 0;
-            }
-        }
+        const resolved = await this.resolveEditionAndAmendment(programName, edition);
+        edition = edition ?? resolved.edition;
```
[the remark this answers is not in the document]
marshacb: Deploy doesn't need amendment. Discarding it is intentional. No reason to add a separate edition-only resolver for one call site.

sdk/src/program-manager.ts
```diff
@@ -371,6 +381,350 @@ class ProgramManager {
...
+            if (collected.has(name)) continue;
+            collected.set(name, source);
+
+            // Discover transitive imports and collect called functions.
+            // Use regex instead of Program.fromString to avoid a WASM
+            // round-trip per import — we only need import names here.
+            try {
+                const subImports = ProgramManager.getImportNames(source);
+                if (subImports.length > 0) {
+                    const transitiveCalls = ProgramManager.getCalledFunctions(source);
```
iamalwaysuncomfortable: This should only resolve to calls directly in the call chain.

sdk/src/program-manager.ts
```diff
@@ -3331,12 +3687,8 @@ class ProgramManager {
...
         }
         const programSource = program ? program.toString() : await this.networkClient.getProgram(programName, edition);
-        const programImports = imports ? imports : await this.networkClient.getProgramImports(programSource);
-        console.log(JSON.stringify(programImports));
-        if (Object.keys(programImports)) {
-            return WasmProgramManager.estimateFeeForAuthorization(authorization, programSource, programImports, edition);
-        }
-        return WasmProgramManager.estimateFeeForAuthorization(authorization, programSource, imports, edition);
+        const programImports = imports ?? await this.networkClient.getProgramImports(programSource);
+        return WasmProgramManager.estimateFeeForAuthorization(authorization, programSource, programImports, edition);
```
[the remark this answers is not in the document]
marshacb: estimateExecutionFee internally calls getProgramImports(programSource) to resolve its own imports. It doesn't rely on the caller's partial imports variable. Not a real issue.

sdk/src/program-manager.ts
iamalwaysuncomfortable: Or there a reason for getting ALL of the functions with call instructions?

sdk/src/program-manager.ts
```diff
@@ -371,6 +381,350 @@ class ProgramManager {
...
+                    const vk = builder.getVerifyingKey(programName, fnName);
+                    if (pk && vk) {
+                        await keyStore.setKeys(pkLocator, vkLocator, [pk, vk]);
+                    }
+                } catch (e) {
+                    console.debug(`Failed to persist key for ${programName}/${fnName}: ${e}`);
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
marshacb: `prepare_for_execution` adds the top-level program to `inner.sources` (imports.rs line 614), so `programNames()` does include it. Combined with line 1302 which adds the top-level edition to `importEditions`, `persistExtractedKeys` covers the top-level program's keys. Not a real issue.

sdk/src/program-manager.ts
iamalwaysuncomfortable: Don't be afraid to use `wasm` here if need be to get the exact call chain.

iamalwaysuncomfortable: Note for dynamic dispatch:

The one thing that might need to be accounted for is functions with `call.dynamic` instructions. In this case we'd look inside the user provided imports for any functions provided by the user for functions that match the function name being called by call.dynamic and trace the call graph of both paths. However, that isn't really represented in this behavior here anyways so we can leave it as a follow on.

iamalwaysuncomfortable (commented):
A couple of comments about called functions here.

sdk/src/program-manager.ts
marshacb: Good call, that's right. This scans all call instructions across the entire program rather than just the ones reachable from the entry function. I'll scope it to trace the actual call chain from the entry point so we only load keys for functions in the execution path. Noted on the call.dynamic follow up.

sdk/src/program-manager.ts
```diff
@@ -371,6 +381,397 @@ class ProgramManager {
...
+
+            for (const fnName of fns) {
+                try {
+                    const pkLocator = provingKeyLocator(programName, fnName, edition, amendment);
+                    const vkLocator = verifyingKeyLocator(programName, fnName, edition, amendment);
+
+                    // Skip keys already in the store.
+                    if (await keyStore.has(pkLocator) && await keyStore.has(vkLocator)) {
+                        continue;
+                    }
```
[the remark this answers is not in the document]
marshacb: Low risk. This requires a crash mid-write where VK persists but PK doesn't, AND the next execution doesn't re-synthesize the PK. In practice, setKeys writes both atomically (IndexedDB transaction / file write), and the builder always has both keys after execution. Not worth adding complexity for this edge case.

sdk/src/program-manager.ts
```diff
@@ -371,6 +381,397 @@ class ProgramManager {
...
+     *
+     * @param loadKeys When true (default), loads cached proving/verifying keys
+     *   from the KeyStore into the builder. Set to false for authorization and
+     *   proving request paths where keys are not synthesized.
+     */
+    private async buildProgramImports(
+        program: string | Program,
+        imports?: ProgramImports,
+        loadKeys: boolean = true,
+        entryFunction?: string,
```
iamalwaysuncomfortable: Would the ever be a case where we don't have the top level entry function? If not we should make this a mandatory parameter so as to reduce the chances of error in usage of this function.

sdk/src/program-manager.ts
```diff
@@ -371,6 +381,397 @@ class ProgramManager {
...
+                        );
+                        for (const entry of fetched) {
+                            if (entry && !collected.has(entry[0])) {
+                                collectQueue.push(entry);
+                            }
+                        }
+                    }
+                }
+            } catch {
+                // Non-blocking — transitive resolution is best-effort
```
iamalwaysuncomfortable: We should emit a warning here at the very least.

sdk/src/program-manager.ts
```diff
@@ -371,6 +381,397 @@ class ProgramManager {
...
+                        for (const [p, fns] of reachable) {
+                            if (!calledFunctions.has(p)) {
+                                calledFunctions.set(p, new Set(fns));
+                            } else {
+                                for (const f of fns) calledFunctions.get(p)!.add(f);
+                            }
+                        }
+                    }
+                } catch {
+                    // Non-blocking — transitive call tracing is best-effort
```
iamalwaysuncomfortable: We should put a warning here.

sdk/src/program-manager.ts
```diff
@@ -371,6 +381,397 @@ class ProgramManager {
...
+                if (!functionsNeeded) continue;
+                try {
+                    const importProgram = Program.fromString(source);
+                    for (const fn of functionsNeeded) {
+                        const reachable = ProgramManager.callGraphToMap(importProgram.getCallGraph(fn));
+                        for (const [p, fns] of reachable) {
+                            if (!calledFunctions.has(p)) {
+                                calledFunctions.set(p, new Set(fns));
+                            } else {
+                                for (const f of fns) calledFunctions.get(p)!.add(f);
```
iamalwaysuncomfortable, line 520: Are we sure this avoids duplicates?

iamalwaysuncomfortable (commented):
A few more nits.

sdk/src/program-manager.ts
marshacb: Done. All callers always have it, so removed the optional + fallback path

sdk/src/program-manager.ts
marshacb: Added `console.warn`

sdk/src/program-manager.ts
marshacb: calledFunctions values are `Set<string>`, so `.add()` is idempotent. Added a test asserting dedup.

sdk/src/program-manager.ts
marshacb: Added `console.warn`

iamalwaysuncomfortable (approved):

PR #1263, opened 2026-03-25
[Feature] Add snarkVerify and snarkVerifyBatch for off-chain SNARK proof verification

marshacb:
## Motivation

Enables off-chain SNARK proof verification without requiring programs to be deployed on-chain. This allows users to verify proofs received from external sources (e.g. another party proved a computation) directly through the SDK using Varuna V2.

## Test Plan

- WASM: proof.rs unit tests for invalid string/bytes parsing
- SDK: program-manager.test.ts covers snarkVerify (valid + invalid inputs), snarkVerifyBatch (valid, invalid, length mismatch), ProgramManager.verifyProof (raw fields, type conversion, wrong inputs), ProgramManager.verifyBatchProof (valid + invalid), Proof string/bytes round-trip, and Value.toFields conversion
- [x] All 13 new tests pass locally against both mainnet and testnet builds

## Related PRs

#1216


wasm/src/synthesizer/proof.rs
```diff
@@ -0,0 +1,107 @@
...
+use crate::types::native::ProofNative;
+use snarkvm_wasm::utilities::{FromBytes, ToBytes};
+
+use std::{ops::Deref, str::FromStr};
+use wasm_bindgen::prelude::wasm_bindgen;
+
+/// SNARK proof for verification of program execution
+#[wasm_bindgen]
+#[derive(Clone, Debug)]
+pub struct Proof(ProofNative);
```
iamalwaysuncomfortable, line 26: This should be moved to the synthesizer directory as that's where it appears in SnarkVM.

sdk/src/program-manager.ts
```diff
@@ -3116,6 +3146,116 @@ class ProgramManager {
...
+        const fieldInputs = options.inputs.map(
+            circuit => circuit.map(instance => this.inputsToFields(instance))
+        );
+        const proof = Proof.fromString(options.proof);
+        try {
+            return snarkVerifyBatch(options.verifyingKeys, fieldInputs, proof);
+        } finally {
+            proof.free();
+        }
+    }
```
iamalwaysuncomfortable: Let's make these standalone functions, we can keep them in the `program-manager.ts` file for now until we find a better place for them. We can also make inputsToFields public as it's generally helpful.

wasm/src/programs/execution.rs
```diff
@@ -195,3 +196,85 @@ pub fn verify_function_execution(
...
+///
+/// This function verifies a batch proof produced by Aleo programs that may not be deployed on chain.
+/// Each verifying key is paired with one or more sets of public inputs (instances).
+///
+/// @param {Array<string>} verifying_keys Array of verifying key strings, one per circuit
+/// @param {Array<Array<Array<string>>>} inputs 3D array of field element strings [circuit_idx][instance_idx][field_idx]
+/// @param {Proof} proof The batch proof to verify
+/// @returns {boolean} True if the batch proof is valid, false otherwise
+#[wasm_bindgen(js_name = "snarkVerifyBatch")]
+pub fn snark_verify_batch(verifying_keys: Array, inputs: Array, proof: &Proof) -> Result<bool, String> {
```
iamalwaysuncomfortable, line 228: These are definitely good as standalone functions (there's %100 good reason to have them as such), but we should also export the .verify and .verifyBatch methods directly on the VerifyingKey struct that calls this standalone function internally so that we're matching interfaces with SnarkVM.

wasm/src/programs/execution.rs
```diff
@@ -195,3 +196,85 @@ pub fn verify_function_execution(
...
+            let fields = parse_field_inputs(&instance_js)?;
+            instances.push(fields);
+        }
+
+        vks_with_inputs.push((vk_native, instances));
+    }
+
+    // Verify the batch proof.
+    VerifyingKeyNative::verify_batch("snark_verify_batch", VarunaVersion::V2, vks_with_inputs, proof)
+        .map_or(Ok(false), |_| Ok(true))
```
iamalwaysuncomfortable, line 260: This method is the one I'm mentioning, it can have the same external singature as the function that encloses this.

wasm/src/programs/execution.rs
```diff
@@ -195,3 +196,85 @@ pub fn verify_function_execution(
...
+/// @param {Array<string>} inputs Array of field element strings representing public inputs (e.g. ["1field", "2field"])
+/// @param {Proof} proof The proof to verify
+/// @returns {boolean} True if the proof is valid, false otherwise
+#[wasm_bindgen(js_name = "snarkVerify")]
+pub fn snark_verify(verifying_key: &VerifyingKey, inputs: Array, proof: &Proof) -> Result<bool, String> {
+    // Parse the input field elements from strings.
+    let raw_inputs = parse_field_inputs(&inputs)?;
+
+    // Verify the proof using VarunaVersion::V2 (current standard).
+    let is_valid = verifying_key.verify("snark_verify", VarunaVersion::V2, &raw_inputs, proof);
```
iamalwaysuncomfortable: We should expose this method on the verifying key struct (using just the inputs + proof)

iamalwaysuncomfortable (commented):
Couple of organizational requests, but overall this is almost ready.

sdk/src/program-manager.ts
```diff
@@ -3757,4 +3785,111 @@ function validateTransferType(transferType: string): string {
...
+ *
+ * @param {string[]} inputs Array of Aleo value strings
+ * @returns {string[]} Array of field element strings
+ */
+function inputsToFields(inputs: string[]): string[] {
+    const fields: string[] = [];
+    for (const input of inputs) {
+        if (input.endsWith("field")) {
+            fields.push(input);
+        } else {
```
iamalwaysuncomfortable, line 3803: No need for this manual parsing anymore, Value.fromString() will parse fields as well in the most faithful way, so we should use Value even for fields.

sdk/src/program-manager.ts
```diff
@@ -3757,4 +3785,111 @@ function validateTransferType(transferType: string): string {
...
+            const value = Value.fromString(input);
+            const valueFields = value.toFields();
+            for (const f of valueFields) {
+                fields.push(f.toString());
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: This finally block is actually correct.

wasm/src/programs/execution.rs
```diff
@@ -195,3 +197,77 @@ pub fn verify_function_execution(
...
+        let vk_str =
+            verifying_keys.get(i).as_string().ok_or_else(|| format!("Expected verifying key string at index {i}"))?;
+        let vk_native = VerifyingKeyNative::from_str(&vk_str)
+            .map_err(|e| format!("Failed to parse verifying key at index {i}: {e}"))?;
+
+        let instances_js = Array::from(&inputs.get(i));
+        let mut instances = Vec::with_capacity(instances_js.length() as usize);
+
+        for j in 0..instances_js.length() {
+            let instance_js = Array::from(&instances_js.get(j));
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: This is a valid comment.

wasm/src/synthesizer/proof.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn test_proof_from_invalid_string() {
+        let result = Proof::from_string("invalid_proof_string");
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn test_proof_from_invalid_bytes() {
+        let result = Proof::from_bytes(&[0u8; 10]);
+        assert!(result.is_err());
+    }
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Also a valid comment/

sdk/src/program-manager.ts
marshacb: IIUC `Value.fromString("1field").toFields()` doesn't return the raw field (e.g. ["1field"]). It encodes type metadata (variant tag, literal type, bit length) into the output. The shortcut ensures raw fields pass through verbatim to match what Varuna expects. Should we still remove it?

iamalwaysuncomfortable (approved):

sdk/src/program-manager.ts
iamalwaysuncomfortable: You're correct!

iamalwaysuncomfortable (approved):

PR #1285, opened 2026-04-08
[Fix] SnapshotQuery failing for programs with dynamic.record inputs

marshacb:
## Motivation

Programs using `call.dynamic` with `dynamic.record` inputs failed during local proving with `proveExecution=true`, throwing `Not all commitments found in stored state paths.` 

Root cause: `SnapshotQuery::collect_commitments_from_inputs` only handled `Record` and `ExternalRecord` value types. `DynamicRecord` fell through to a catch-all `_ => None`, producing empty commitments. Additionally, the REST mode fallback used a hardcoded `DEFAULT_URL` instead of the caller's `node_url`, and `current_block_height()` (sync) returned 0 for REST-mode queries, causing `Output record must be Version 0 before Consensus V8` errors.

## Test Plan

- Deployed `test_dcall_sdk.aleo` on testnet with two functions: `dynamic_transfer_pub_to_priv` (basic dynamic dispatch) and `dynamic_transfer_private` (dynamic.record input)
- Added `template-node-dynamic-dispatch-proving-ts` integration test that builds execution transactions with `proveExecution=true` against testnet — both tests pass end-to-end
- Verified without the fix: "Not all commitments found" error reproduced
- Verified with the fix: transactions build successfully with valid transaction IDs
- Added Rust unit tests for `has_dynamic_record_inputs` detection (positive and negative cases)
- Added CI job `create-leo-app-dynamic-dispatch-proving` with 10-minute timeout
- All `SnapshotQuery.current_block_height()` call sites (5 files) updated to async variant defensively

---

> [!NOTE]
> **Medium Risk**
> Touches inclusion-proof/query logic used when building/proving transactions, so mistakes could break proving or cause network-dependent failures. Added CI coverage reduces risk, but the new testnet-based job can be flaky if the external API/node is unstable.
> 
> **Overview**
> Fixes local proving failures for programs using `call.dynamic` with `dynamic.record`/`dynamic.future` inputs by making `SnapshotQuery::try_from_inputs` detect these input types and **fall back to REST mode** (fetching state paths on-demand instead of precomputing commitments).
> 
> Updates REST-mode `SnapshotQuery` to **store and use the caller’s `node_url`** for all async queries (state root, state paths, block height), and updates multiple proving call sites to use `current_block_height_async()` plus `SnapshotQuery::rest(node_url)`.
> 
> Adds a new `create-leo-app` template (`template-node-dynamic-dispatch-proving-ts`) and a GitHub Actions job to run an integration test against testnet that builds execution transactions for both basic dynamic dispatch and dynamic-record inputs, plus unit tests for dynamic-record input detection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bed4121662c859f86fb115c158741f75d209618e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -159,6 +159,21 @@ jobs:
           yarn local
 
 
+  create-leo-app-dynamic-dispatch-proving:
+    name: "create-leo-app dynamic-dispatch-proving (testnet)"
+    runs-on: ubuntu-latest-m
+    needs: build
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-yarn
+      - uses: ./.github/actions/use-build
+
+      - name: Run dynamic dispatch proving test against testnet
+        working-directory: create-leo-app/template-node-dynamic-dispatch-proving-ts
+        timeout-minutes: 10
+        run: |
+          yarn start
+
   create-leo-app-loyalty-program-delegated:
     name: "create-leo-app loyalty-program (delegated + scanner)"
     runs-on: ubuntu-latest-m
--- a/create-leo-app/template-node-dynamic-dispatch-proving-ts/.gitignore
+++ b/create-leo-app/template-node-dynamic-dispatch-proving-ts/.gitignore
@@ -1,0 +1,11 @@
+# Logs
+logs
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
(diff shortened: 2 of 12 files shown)


iamalwaysuncomfortable (approved):
This is a really good catch, thanks for taking care of this.

PR #1293, opened 2026-04-14
[Feature] Add CJS build target to @provablehq/sdk and @provablehq/wasm

marshacb:
## Motivation

A partner team requested that we ship CommonJS-compatible builds so they can `require('@provablehq/sdk')` directly from a CJS codebase without wrapping every entry point in an `await import()` IIFE. Today both `@provablehq/sdk` and `@provablehq/wasm` are ESM-only, which fails with `ERR_REQUIRE_ESM` when a CJS consumer tries to load them.

This PR adds a CJS build target to both packages alongside the existing ESM output, with full type declarations (`.d.cts`), threaded worker support, and CI validation. CJS consumers can now `require()` the SDK with the same threaded API surface as ESM consumers.

> **Note:** The noble-sodium crypto swap and `.js` import extension fixes that were originally part of this branch have been split out and merged independently via #1295. This PR is now purely the CJS build system work.

### What's in the PR

- **CJS output for `@provablehq/wasm`**: a second build pass emits `dist/<network>/index.cjs` alongside `index.js`. The CJS entry bypasses wasm-bindgen's top-level-await initialization by importing the underlying glue module directly and calling `initSync` with `fs.readFileSync`'d wasm bytes. Both formats share the same `.wasm` binary.

- **CJS output for `@provablehq/sdk`**: a second Rollup output entry per network emits `dist/<network>/{node,browser,node-polyfill}.cjs` alongside the ESM `.js` files. Single TypeScript compile, two output formats. The dynamic variant gets a CJS loader too.

- **`exports` map with `types`/`import`/`require` conditions**: both packages now have full conditional exports. `arethetypeswrong --profile=node16` passes green for all subpaths on both `import` and `require`.

- **CI validation**: a new `cjs-smoke` job runs on every PR (attw + file-path smoke + packed-and-installed smoke including `initThreadPool`). A separate `cjs-e2e` job runs only on pushes to protected branches and exercises the full encrypted DPS round-trip against staging — gated to avoid burning testnet transactions on every PR.

### Notable callouts

- **`main` field flipped from ESM to CJS** in both packages. Modern consumers resolve through the `exports` map and aren't affected. Legacy consumers that bypass `exports` (some old bundlers, `require.resolve`) previously got ESM via `main` and now get CJS — this is the correct change for `main`'s historical CJS-only semantics.
- **Workers stay ESM** in both formats. The Node polyfill spawns them via `eval` + dynamic `import()`, which works fine from a CJS parent. No second wasm-bindgen worker target needed.
- **Known limitation (documented in code)**: the wasm CJS entry generation parses wasm-bindgen's emitted JS textually to recover short-name aliases. Stable across wasm-bindgen 0.2.x but worth knowing about. The build will throw loudly on a future wasm-bindgen output-shape change rather than silently miscompile, and CI smoke will catch it.

## Test Plan

Validated end-to-end on both supported paths.

**Local validation (run from repo root):**

```bash
yarn build:all
yarn test:sdk                              
node sdk/tests/cjs-smoke/smoke.cjs         # CJS file-path smoke - 17 checks
cd sdk && yarn test:cjs:pack               # pack + install + require() through exports.require
yarn test:attw                             # exports map validation (node16 CJS + ESM + bundler)
cd ../wasm && yarn test:attw               # same for the wasm package
```

### CJS support window

CJS support is accepted with a 12-month review window (April 2027). See [ADR-001](docs/adr/001-cjs-support.md) for constraints, removal plan, and review criteria.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes build outputs and package `exports`/`main` resolution for both `@provablehq/sdk` and `@provablehq/wasm`, which can affect downstream consumers’ module and type resolution. Runtime logic is largely unchanged, but the new CJS init path and packaging validations could surface edge-case compatibility issues.
> 
> **Overview**
> Adds **first-class CommonJS (`require`) support** alongside existing ESM outputs for both `@provablehq/sdk` and `@provablehq/wasm`, including conditional `exports` entries with `import`/`require` + `types` mappings and `.d.cts` type artifacts.
> 
> Updates the build pipeline to emit `.cjs` bundles (SDK via dual Rollup outputs + generated CJS dynamic loader; WASM via a new CJS entry that uses synchronous `initSync` with `fs.readFileSync`), and adds a Rollup post-step to duplicate `.d.ts` to `.d.cts` for CJS type resolution.
> 
> Introduces **CI gating** for CJS compatibility (`arethetypeswrong` checks plus local and packed-install CJS smoke tests), adds CJS smoke fixtures/scripts, tweaks worker polyfill concurrency detection, and documents the CJS support decision/removal plan in a new ADR.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da980f0dc92516c715cdc6489e916609729cf3f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -183,8 +183,17 @@ jobs:
         working-directory: sdk
         run: yarn test:attw
 
-      - name: "Pure-CJS smoke (local)"
+      - name: "Pure-CJS smoke (local, resolves by file path)"
         run: node sdk/tests/cjs-smoke/smoke.cjs
+
+      # Packs both workspace packages, installs them into a pristine CJS
+      # fixture, and `require('@provablehq/sdk/testnet.js')` through the
+      # actual `exports.require` condition. Catches broken `exports` maps,
+      # missing tarball files, and wrong `main`/`module` fields that the
+      # file-path smoke above cannot see.
+      - name: "Packed-and-installed CJS smoke"
+        working-directory: sdk
+        run: yarn test:cjs:pack
 
 
   # Full CJS + encrypted DPS round-trip against staging. Proves a CJS consumer
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -159,6 +159,60 @@ jobs:
           yarn local
 
 
+  # Cheap, deterministic, no secrets. Always runs on every PR and push.
+  # Exercises: CJS module load (no TLA), Account/signature WASM calls, noble
+  # encryption wire format, `initThreadPool` worker spawning from CJS, dynamic
+  # variant `require()` from CJS, and `exports`-map consistency via
+  # `arethetypeswrong`.
+  cjs-smoke:
+    name: "CJS smoke + exports validation"
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-yarn
+      - uses: ./.github/actions/use-build
+
+      # Exports-map validation runs first so it gates merging even if the
+      # network-dependent e2e job (below) is flaky or skipped.
+      - name: "arethetypeswrong exports validation (wasm)"
+        working-directory: wasm
+        run: yarn test:attw
+
+      - name: "arethetypeswrong exports validation (sdk)"
+        working-directory: sdk
+        run: yarn test:attw
+
+      - name: "Pure-CJS smoke (local)"
+        run: node sdk/tests/cjs-smoke/smoke.cjs
+
+
+  # Full CJS + encrypted DPS round-trip against staging. Proves a CJS consumer
+  # can build a proving request, encrypt with noble, submit, and confirm
+  # on-chain. Costs a real testnet transaction per run, so we only run it on
+  # pushes to protected branches (mainnet/testnet) — the parallel ESM loyalty
+  # job already covers general DPS path validation on every PR, and the
+  # `cjs-smoke` + `attw` jobs above catch CJS-specific packaging regressions
+  # without hitting the network.
+  cjs-e2e:
+    name: "CJS + encrypted DPS end-to-end"
+    runs-on: ubuntu-latest
+    needs: build
+    if: github.event_name == 'push'
+    env:
+      ALEO_CONSUMER_ID: ${{ secrets.ALEO_CONSUMER_ID }}
+      ALEO_DPS_URL: https://api.provable.com/prove/testnet
+      ALEO_DPS_API_KEY: ${{ secrets.ALEO_DPS_API_KEY }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-yarn
+      - uses: ./.github/actions/use-build
+
+      - name: "CJS + encrypted DPS full round-trip (staging)"
+        timeout-minutes: 10
+        run: node sdk/tests/cjs-smoke/loyalty-dps.cjs
+
+
   create-leo-app-dynamic-dispatch-proving:
     name: "create-leo-app dynamic-dispatch-proving (testnet)"
     runs-on: ubuntu-latest-m
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -196,32 +196,6 @@ jobs:
         run: yarn test:cjs:pack
 
 
-  # Full CJS + encrypted DPS round-trip against staging. Proves a CJS consumer
-  # can build a proving request, encrypt with noble, submit, and confirm
-  # on-chain. Costs a real testnet transaction per run, so we only run it on
-  # pushes to protected branches (mainnet/testnet) — the parallel ESM loyalty
-  # job already covers general DPS path validation on every PR, and the
-  # `cjs-smoke` + `attw` jobs above catch CJS-specific packaging regressions
-  # without hitting the network.
-  cjs-e2e:
-    name: "CJS + encrypted DPS end-to-end"
-    runs-on: ubuntu-latest
-    needs: build
-    if: github.event_name == 'push'
-    env:
-      ALEO_CONSUMER_ID: ${{ secrets.ALEO_CONSUMER_ID }}
-      ALEO_DPS_URL: https://api.provable.com/prove/testnet
-      ALEO_DPS_API_KEY: ${{ secrets.ALEO_DPS_API_KEY }}
-    steps:
-      - uses: actions/checkout@v4
-      - uses: ./.github/actions/setup-yarn
-      - uses: ./.github/actions/use-build
-
-      - name: "CJS + encrypted DPS full round-trip (staging)"
-        timeout-minutes: 10
-        run: node sdk/tests/cjs-smoke/loyalty-dps.cjs
-
-
   create-leo-app-dynamic-dispatch-proving:
     name: "create-leo-app dynamic-dispatch-proving (testnet)"
     runs-on: ubuntu-latest-m
--- a/docs/adr/001-cjs-support.md
+++ b/docs/adr/001-cjs-support.md
@@ -1,0 +1,62 @@
+# ADR-001: CommonJS Build Target
+
+**Status:** Accepted  
+**Date:** 2026-04-17  
+**Review by:** 2027-04-17
+
+## Context
+
+A partner integration required `require()` support from a CommonJS codebase. Their stack cannot use ESM imports (including dynamic `await import()`), and the alternative was maintaining an internal fork of the SDK.
+
+Shipping a CJS build target requires that we avoid top-level await (TLA) anywhere in the source tree that gets bundled into the Node CJS entry. Rollup's CJS output format does not support TLA and hard-fails the build if any is present.
+
+## Decision
+
+Ship CJS output alongside ESM for both `@provablehq/sdk` and `@provablehq/wasm`, with a 12-month review window.
+
+### Constraints accepted
+
+- **No TLA in the Node entry graph.** Any `await` at module scope in a source file reachable from `node.ts` will fail the CJS build. Enforced automatically by Rollup at build time (exit code 1).
+- **`initSync` for WASM initialization in CJS.** The CJS entry uses wasm-bindgen's `initSync` with `fs.readFileSync` instead of the default async `init()`. This is a documented wasm-bindgen API ([synchronous instantiation](https://wasm-bindgen.github.io/wasm-bindgen/examples/synchronous-instantiation.html)), but its long-term availability is not guaranteed.
+- **No TLA-dependent initialization patterns.** Any future startup initialization (metadata fetching, consensus config, etc.) must use explicit init functions (same pattern as `initThreadPool`) rather than TLA.
+
+### What is unaffected
+
+- ESM consumers get identical output. CJS files ship alongside and bundlers pick the format they need.
+- Async functions (including `initThreadPool`) work normally from CJS when called inside async contexts.
+- Browser builds are unchanged.
+- Source code itself has no CJS-specific logic. All CJS concerns live in the build/packaging layer.
+
+## Removal plan
+
+CJS support is structured for clean removal. All CJS-specific code is additive:
+
+| Component | What to remove |
+|---|---|
+| `wasm/build.js` | CJS entry template |
+| `wasm/package.json` | `require` condition in exports map |
+| `sdk/package.json` | `require`/`types.require` conditions, revert `main` field, remove `test:cjs*` scripts |
+| `sdk/rollup.config.js` | CJS output entry |
+| `sdk/tests/cjs-smoke/` | Entire directory |
+| `.github/workflows/sdk.yml` | CJS CI jobs |
+
+No source code changes required. Removing CJS is equivalent to reverting PR #1293.
+
+## Review criteria (April 2027)
+
+At the 12-month mark, evaluate:
+
+1. **Is anyone using the CJS build?** Check npm download stats for `require` condition hits if available, or survey known consumers.
+2. **Has the TLA constraint caused friction?** Review whether any feature or dependency was blocked or complicated by the no-TLA rule.
+3. **Has `initSync` been deprecated upstream?** Check wasm-bindgen's status on synchronous instantiation.
+4. **Has the Node/npm ecosystem moved further from CJS?** If major LTS versions or frameworks have dropped CJS support, the case for maintaining it weakens.
+
+If usage is low and the constraint has caused friction, drop CJS support. If usage is meaningful, extend with a new review date.
+
+## References
+
+- PR #1293: CJS build target implementation
+- PR #1295: Noble-sodium swap (removed `await sodium.ready` TLA, prerequisite for CJS)
+- wasm-bindgen synchronous instantiation: https://wasm-bindgen.github.io/wasm-bindgen/examples/synchronous-instantiation.html
+- wasm-bindgen initSync discussion: https://github.com/wasm-bindgen/wasm-bindgen/issues/1976
+- W3C Service Worker / initSync context: https://github.com/w3c/ServiceWorker/issues/1499
--- a/sdk/package.json
+++ b/sdk/package.json
@@ -7,23 +7,60 @@
   ],
   "license": "GPL-3.0",
   "type": "module",
-  "main": "./dist/testnet/node.js",
+  "main": "./dist/testnet/node.cjs",
+  "module": "./dist/testnet/node.js",
   "browser": "./dist/testnet/browser.js",
   "exports": {
     ".": {
-      "node": "./dist/testnet/node.js",
+      "node": {
+        "types": {
+          "import": "./dist/testnet/node.d.ts",
+          "require": "./dist/testnet/node.d.cts"
+        },
+        "import": "./dist/testnet/node.js",
+        "require": "./dist/testnet/node.cjs",
+        "default": "./dist/testnet/node.js"
+      },
+      "types": "./dist/testnet/browser.d.ts",
       "default": "./dist/testnet/browser.js"
     },
     "./testnet.js": {
-      "node": "./dist/testnet/node.js",
+      "node": {
+        "types": {
+          "import": "./dist/testnet/node.d.ts",
+          "require": "./dist/testnet/node.d.cts"
+        },
+        "import": "./dist/testnet/node.js",
+        "require": "./dist/testnet/node.cjs",
+        "default": "./dist/testnet/node.js"
+      },
+      "types": "./dist/testnet/browser.d.ts",
       "default": "./dist/testnet/browser.js"
     },
     "./mainnet.js": {
-      "node": "./dist/mainnet/node.js",
+      "node": {
+        "types": {
+          "import": "./dist/mainnet/node.d.ts",
+          "require": "./dist/mainnet/node.d.cts"
+        },
+        "import": "./dist/mainnet/node.js",
+        "require": "./dist/mainnet/node.cjs",
+        "default": "./dist/mainnet/node.js"
+      },
+      "types": "./dist/mainnet/browser.d.ts",
       "default": "./dist/mainnet/browser.js"
     },
     "./dynamic.js": {
-      "node": "./dist/dynamic/node.js",
+      "node": {
+        "types": {
+          "import": "./dist/dynamic/node.d.ts",
+          "require": "./dist/dynamic/node.d.cts"
+        },
+        "import": "./dist/dynamic/node.js",
+        "require": "./dist/dynamic/node.cjs",
+        "default": "./dist/dynamic/node.js"
+      },
+      "types": "./dist/dynamic/browser.d.ts",
       "default": "./dist/dynamic/browser.js"
     }
   },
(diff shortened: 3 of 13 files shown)


wasm/build.js
```diff
@@ -70,20 +72,128 @@ export async function initThreadPool(threads) {
...
+function extractReexportAliases(source) {
+    const exportMatch = source.match(/export \{([^}]+)\} from/);
+    if (!exportMatch) {
+        throw new Error("Could not locate re-export block in aleo_wasm.js");
+    }
+    return exportMatch[1]
+        .split(",")
+        .map((s) => s.trim())
+        .filter((s) => s && !/\bas (?:initSync|__wbg_init|initThreadPool)$/.test(s))
+        .join(", ");
```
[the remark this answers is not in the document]
marshacb: Won't fix. Claim is empirically incorrect. Verified: neither i`nitSync` nor `__wbg_init` is declared in `aleo_wasm.d.ts` (`wasm-bindgen` treats them as internal and omits type declarations). Since i`ndex.d.ts/index.d.cts` only re-export * from "./aleo_wasm.js", neither symbol is in either type file. A CJS consumer destructuring `initSync` would fail TypeScript type-checking before any runtime issue. Types and runtime exports are consistent.

iamalwaysuncomfortable (commented):
@Pauan could you take a look at this and give it a once over?

iamalwaysuncomfortable:
One question is what was the overall effect on the bundle size?

Pauan:
I would really like to avoid a CommonJS build if we can help it.

ES6 modules have been [available in Node for the past 9 years](https://nodejs.org/en/blog/release/v8.5.0), and they've been [enabled by default in Node for the past 7 years](https://nodejs.org/en/blog/release/v13.2.0).

ES6 modules have been heavily pushed as *the* one true standard way of writing JavaScript code, and developers have had plenty of time (7 years!) to start adopting them.

At a certain point we have to drop support for very old and deprecated technologies.

Before we commit to adding a CommonJS build (which is a quite large maintenance burden), we will need some more information on exactly why they need to use `require` and cannot use `await import` to import the SDK.

----

@marshacb Does it work correctly if a CommonJS module uses `await import` instead of `require`? Or does this force CommonJS modules to always use `require`?

If a CommonJS module uses `await import`, will they get the ES6 version or the CommonJS version?

marshacb:
> One question is what was the overall effect on the bundle size?

Measured the esbuild bundle of the testnet entry, minified + gzipped, with @provablehq/wasm and node:* external (consumers pull those separately):

| Target | Format | Baseline (`mainnet` branch) | This branch | Delta |
|---|---|---|---|---|
| Browser | ESM | 316 KB | 42 KB | **−274 KB (−87%)** |
| Node    | ESM | 334 KB | 54 KB | **−280 KB (−84%)** |
| Node    | CJS | not emitted    | 65 KB | new format        |

Much smaller. Dominated by dropping the libsodium WASM blob (it ships as a ~600 KB base64 string inside `libsodium.js`) and replacing with the noble primitives we actually call.

Tarball does grow ~+84% (451 KB → 830 KB), driven by two things: (1) we now emit `.cjs` alongside `.js` for each entry. Both ship in the package but a consumer's bundler picks the one it needs; (2) we still emit per-network variants (testnet/mainnet/dynamic). Consumers only load one at runtime. Neither hits the consumer's bundle.

marshacb:
> I would really like to avoid a CommonJS build if we can help it.
> 
> ES6 modules have been [available in Node for the past 9 years](https://nodejs.org/en/blog/release/v8.5.0), and they've been [enabled by default in Node for the past 7 years](https://nodejs.org/en/blog/release/v13.2.0).
> 
> ES6 modules have been heavily pushed as _the_ one true standard way of writing JavaScript code, and developers have had plenty of time (7 years!) to start adopting them.
> 
> At a certain point we have to drop support for very old and deprecated technologies.
> 
> Before we commit to adding a CommonJS build (which is a quite large maintenance burden), we will need some more information on exactly why they need to use `require` and cannot use `await import` to import the SDK.
> 
> @marshacb Does it work correctly if a CommonJS module uses `await import` instead of `require`? Or does this force CommonJS modules to always use `require`?
> 
> If a CommonJS module uses `await import`, will they get the ES6 version or the CommonJS version?

Hey @Pauan 

Neither path forces the other. Both work, the consumer just picks by syntax. A CJS module using `await import("@provablehq/sdk/testnet.js")` resolves the `"import"` condition and gets the ESM build. A CJS module using `require("@provablehq/sdk/testnet.js")` resolves the `"require"` condition and gets the new CJS build. Node docs: https://nodejs.org/api/packages.html#conditional-exports.

Verified empirically from a `"type": "commonjs"` package against this branch:

```
require.resolve(...)  → /.../sdk/dist/testnet/node.cjs
await import(...)     → resolved namespace, Module tag

require()'s Account === import()'s Account?  false
→ different files loaded: require() hit the "require" condition (CJS),
  await import() hit the "import" condition (ESM).
```

So CJS users today can already do `await import()` and get the ESM build (this works on `mainnet` HEAD too). What this PR adds is the option to also `require()` synchronously, without wrapping every entry point in an `async` IIFE.

The specific ask seemed to be that their stack is CJS and they didn't want to refactor to wrap every call site in `await import()` (their words: "no need to write extra code"). Perhaps their refactor would be non trivial?

Fair concern on the maintenance burden. The ongoing CJS surface is mostly build-time: no runtime divergence between formats, the same source compiles to both. The crypto change (libsodium → noble-sodium) is an independent win regardless of the CJS question since it drops the WASM blob, no top-level await, and we'd have a reduced min+gz bundle.

Pauan:
> Neither path forces the other. Both work, the consumer just picks by syntax.

That's what I figured, thanks for checking!

.github/workflows/sdk.yml
```diff
@@ -159,6 +159,69 @@ jobs:
...
+
+
+  # Full CJS + encrypted DPS round-trip against staging. Proves a CJS consumer
+  # can build a proving request, encrypt with noble, submit, and confirm
+  # on-chain. Costs a real testnet transaction per run, so we only run it on
+  # pushes to protected branches (mainnet/testnet) — the parallel ESM loyalty
+  # job already covers general DPS path validation on every PR, and the
+  # `cjs-smoke` + `attw` jobs above catch CJS-specific packaging regressions
+  # without hitting the network.
+  cjs-e2e:
```
Pauan: I don't like that this does a real transaction, is there another way?

docs/adr/001-cjs-support.md
```diff
@@ -0,0 +1,62 @@
...
+| `sdk/tests/cjs-smoke/` | Entire directory |
+| `.github/workflows/sdk.yml` | CJS CI jobs |
+
+No source code changes required. Removing CJS is equivalent to reverting PR #1293.
+
+## Review criteria (April 2027)
+
+At the 12-month mark, evaluate:
+
+1. **Is anyone using the CJS build?** Check npm download stats for `require` condition hits if available, or survey known consumers.
```
Pauan, line 49: We should list out explicitly who is using CommonJS, so we know who to contact.

sdk/package.json
```diff
@@ -7,23 +7,60 @@
...
+  "main": "./dist/testnet/node.cjs",
+  "module": "./dist/testnet/node.js",
   "browser": "./dist/testnet/browser.js",
   "exports": {
     ".": {
-      "node": "./dist/testnet/node.js",
+      "node": {
+        "types": {
+          "import": "./dist/testnet/node.d.ts",
+          "require": "./dist/testnet/node.d.cts"
```
Pauan, line 18: Is this required? Or can it reuse the same types?

sdk/rollup.config.js
```diff
@@ -14,75 +14,153 @@ const runtimes = [
...
+`);
+
+        await $fs.writeFile(`dist/dynamic/${runtime}.cjs`, `"use strict";
+
+const networks = {
+    ${cjsLeaves}
 };
 
-export function loadNetwork(name) {
+async function loadNetwork(name) {
```
Pauan: This should be `export` instead of `async`.

sdk/rollup.config.js
```diff
@@ -14,75 +14,153 @@ const runtimes = [
...
 
-        const typings = networks.map((network) => `"${network}": typeof import("../${network}/${runtime}"),`).join("\n    ");
+        const typings = networks.map((network) => `"${network}": typeof import("../${network}/${runtime}.js"),`).join("\n    ");
 
         await $fs.writeFile(`dist/dynamic/${runtime}.js`, `const networks = {
-    ${loading}
+    ${esmLeaves}
+};
+
+export async function loadNetwork(name) {
```
Pauan: This shouldn't be `async`.

sdk/rollup.config.js
```diff
@@ -14,75 +14,153 @@ const runtimes = [
...
 }
 
 export { Networks };
 
 export declare function loadNetwork<Key extends keyof Networks>(name: Key): Promise<Networks[Key]>;
-`);
+`;
+
+        await $fs.writeFile(`dist/dynamic/${runtime}.d.ts`, dtsContent);
+        await $fs.writeFile(`dist/dynamic/${runtime}.d.cts`, dtsContent);
```
Pauan, line 164: Do we really need two separate type files?

sdk/rollup.config.js
```diff
@@ -14,75 +14,153 @@ const runtimes = [
...
+// through the `require` condition (which expects `.d.cts`); without these
+// siblings `arethetypeswrong` flags "Masquerading as ESM". The declaration
+// bodies are identical — the distinct filename is all TypeScript's node16
+// module-resolution check cares about.
+async function copyDtsToCts(dir) {
+    let entries;
+    try {
+        entries = await $fs.readdir(dir, { withFileTypes: true });
+    } catch {
+        return;
```
Pauan: Why does this ignore errors?

wasm/build.js
```diff
@@ -70,20 +72,128 @@ export async function initThreadPool(threads) {
...
-        plugins: [
-            virtual({
-                "entry": js,
-            }),
-        ],
-    }, {
-        dir: `dist/${network}`,
-        format: "es",
-        sourcemap: true,
+    const bundleEsm = await rollup({
```
Pauan: Why does this not use `buildRollup`?

wasm/build.js
```diff
@@ -70,20 +72,128 @@ export async function initThreadPool(threads) {
...
+// Rollup's CJS output transforms \`import.meta.url\` into
+// \`pathToFileURL(__filename).href\`, so this resolves to the emitted
+// \`index.cjs\` sibling — from which \`aleo_wasm.wasm\` lives next to.
+const __here = dirname(fileURLToPath(import.meta.url));
+const __wasmBytes = readFileSync(join(__here, "aleo_wasm.wasm"));
+__initSync({ module: __wasmBytes });
+
+export { ${reexportAliases} } from "./dist/${network}/tmp/index.js";
+
+export async function initThreadPool(threads) {
```
Pauan: Why does this reimplement `initThreadPool`? The old implementation works fine in NodeJS.

wasm/build.js
```diff
@@ -70,20 +72,128 @@ export async function initThreadPool(threads) {
...
+    if (threads == null) {
+        threads = require('node:os').availableParallelism();
+    }
+
+    console.info(\`Spawning \${threads} threads\`);
+
+    await wasmInitThreadPool(new URL("worker.js", import.meta.url), threads);
+}`;
+
+    const bundleCjs = await rollup({
```
Pauan: Similarly, why isn't this using `buildRollup`?

wasm/build.js
```diff
@@ -70,20 +72,128 @@ export async function initThreadPool(threads) {
...
+    // disk via `fs.readFileSync`, and call `initSync` before re-exporting.
+    //
+    // The re-export alias list is derived at build time from the ESM entry
+    // that wasm-bindgen produces, so the CJS surface stays in sync with the
+    // ESM surface without hand-maintenance.
+    const aleoWasmEsm = await $fs.readFile(
+        `dist/${network}/tmp/aleo_wasm.js`,
+        "utf8",
+    );
+    const reexportAliases = extractReexportAliases(aleoWasmEsm);
```
Pauan: I don't like the way this is implemented, I'm going to submit another PR that does things differently.

Pauan (changes requested):

wasm/build.js
marshacb: The ESM version falls back to `navigator.hardwareConcurrency` for the default thread count, which makes sense for browser + Node. The CJS entry is `Node-only`, so we use `os.availableParallelism()` instead. It's cgroup-aware in containers where `navigator.hardwareConcurrency` (or the polyfill's `os.cpus().length`) can overcount available CPUs.

That said, since you're rewriting the implementation in a follow-up PR, happy to drop this.

sdk/rollup.config.js
marshacb: It's a recursive directory walker. If a subdirectory doesn't exist (readdir throws ENOENT), it returns gracefully instead of crashing the build. Could be more explicit by checking for ENOENT specifically rather than catching all errors.

sdk/rollup.config.js
marshacb: TypeScript's `node16/nodenext` resolution expects `.d.cts` alongside `.cjs` entries. Without it, `attw` flags "Masquerading as ESM." The content is identical, it's just the file extension that the resolver cares about.

sdk/rollup.config.js
marshacb: The async was to handle the case where `networks[name]` is undefined. That throws synchronously before `import()` is reached. But since you're rewriting this part, happy to leave it for the follow-up PR.

sdk/rollup.config.js
marshacb: This is inside the CJS template (exported via `module.exports`, not ES export). The `async` is for the same sync-throw reason. `require()` throws synchronously if the network name is invalid, and without async, `.catch()` on the returned Promise wouldn't catch it. But same as below, happy to leave for your follow-up.

sdk/package.json
marshacb: Yes. TypeScript's `node16/nodenext` resolution requires separate type conditions for import vs require. Without the split, `attw` flags "Masquerading as ESM" because it can't distinguish the CJS type entry from the ESM one. The `.d.cts` content is identical to `.d.ts`, it's the conditional mapping that matters.

.github/workflows/sdk.yml
marshacb: Fair point. The `cjs-smoke` tests already cover CJS packaging, exports, threading, and encryption without hitting the network. The DPS round-trip isn't testing anything CJS-specific. The ESM loyalty job covers that path. Happy to drop the `cjs-e2e job`, or leave it for the follow-up PR if you'd prefer to handle it there.

wasm/build.js
marshacb: It should! That's an oversight from when I split the ESM and CJS builds apart.

wasm/build.js
Pauan: We already polyfill `navigator.hardwareConcurrency` so it works on Node.

https://github.com/ProvableHQ/sdk/blob/bd8b75c26a8b51491282a309b351203178851e19/sdk/src/polyfill/worker.ts#L8-L10

This is our typical strategy: we polyfill the things so they work on Node, instead of writing two separate implementations. This is a much lower maintenance burden.

If you want to make changes to the polyfill, that's fine, but there's no need to re-implement `initThreadPool`.

wasm/build.js
marshacb: Same as above

sdk/rollup.config.js
Pauan: So it's ignoring the `exports.node.types.require` field in package.json?

.github/workflows/sdk.yml
Pauan: Yeah I think it's best to have it in a followup PR.

wasm/build.js
marshacb: Good point! The polyfill approach is cleaner than maintaining two implementations. The reason I used `os.availableParallelism()` directly was that the `wasm` CJS entry can be consumed without the SDK polyfill (e.g. require('@provablehq/wasm') directly), but that's something the follow-up can sort out.

sdk/rollup.config.js
marshacb: It uses the `types.require` field to find the file, but under `node16` resolution it also validates that the file extension matches the module format. If `types.require` points to a `.d.ts` file, `attw`/TypeScript flags it as "Masquerading as ESM." The `.d.cts` extension is what signals "this is a CJS declaration." So both the exports map entry and the .d.cts file are needed.

.github/workflows/sdk.yml
marshacb: Sounds good! Leaving for your follow-up PR.

wasm/build.js
marshacb: Sounds good! You know this code better than anyone. Happy to review or rebase on top of your changes.

sdk/rollup.config.js
Pauan: That's a weird way of them implementing it, but alright, thanks for the explanation.

Pauan:
@marshacb I published a new version 3.1.5 of `@wasm-tools/rollup-plugin-rust`

After updating to that version, you should be able to use this code to build the CommonJS version of Wasm:

```js
// Builds a .cjs version which initializes the Wasm synchronously so the module can be loaded with `require`
async function buildCommonJS(network) {
    const js = `import { module as url, initSync } from "./dist/${network}/tmp/aleo_wasm_custom.js";
import { readFileSync } from "node:fs";
import { fileURLToPath } from "node:url";

const wasm = initSync({
    module: readFileSync(fileURLToPath(url)),
});

// We have to make a shallow copy because ES6 exports are frozen.
const exports = { ...wasm };

const wasmInitThreadPool = wasm.initThreadPool;

exports.initThreadPool = async function initThreadPool(threads) {
    if (threads == null) {
        threads = navigator.hardwareConcurrency;
    }

    console.info(\`Spawning \${threads} threads\`);

    await wasmInitThreadPool(new URL("worker.js", import.meta.url), threads);
};

module.exports = exports;`;

    await buildRollup({
        input: {
            "index": "entry",
        },
        plugins: [
            virtual({
                "entry": js,
            }),
        ],
    }, {
        dir: `dist/${network}`,
        format: "cjs",
        external: ["node:fs", "node:url"],
        entryFileNames: "[name].cjs",
        chunkFileNames: "[name].cjs",
        sourcemap: true,
    });
}
```

A lot cleaner than using regexps, and now the synchronous Wasm support is being maintained by `@wasm-tools/rollup-plugin-rust`, so we no longer need to maintain it in the SDK.

marshacb:
> @marshacb I published a new version 3.1.5 of `@wasm-tools/rollup-plugin-rust`
> 
> After updating to that version, you should be able to use this code to build the CommonJS version of Wasm:
> 
> ```js
> // Builds a .cjs version which initializes the Wasm synchronously so the module can be loaded with `require`
> async function buildCommonJS(network) {
>     const js = `import { module as url, initSync } from "./dist/${network}/tmp/aleo_wasm_custom.js";
> import { readFileSync } from "node:fs";
> import { fileURLToPath } from "node:url";
> 
> const wasm = initSync({
>     module: readFileSync(fileURLToPath(url)),
> });
> 
> // We have to make a shallow copy because ES6 exports are frozen.
> const exports = { ...wasm };
> 
> const wasmInitThreadPool = wasm.initThreadPool;
> 
> exports.initThreadPool = async function initThreadPool(threads) {
>     if (threads == null) {
>         threads = navigator.hardwareConcurrency;
>     }
> 
>     console.info(\`Spawning \${threads} threads\`);
> 
>     await wasmInitThreadPool(new URL("worker.js", import.meta.url), threads);
> };
> 
> module.exports = exports;`;
> 
>     await buildRollup({
>         input: {
>             "index": "entry",
>         },
>         plugins: [
>             virtual({
>                 "entry": js,
>             }),
>         ],
>     }, {
>         dir: `dist/${network}`,
>         format: "cjs",
>         external: ["node:fs", "node:url"],
>         entryFileNames: "[name].cjs",
>         chunkFileNames: "[name].cjs",
>         sourcemap: true,
>     });
> }
> ```
> 
> A lot cleaner than using regexps, and now the synchronous Wasm support is being maintained by `@wasm-tools/rollup-plugin-rust`, so we no longer need to maintain it in the SDK.

This is really clean. Eliminating the regex parsing entirely is a big win. The shallow copy approach for re-exports is much more robust. Happy to update the PR with this. Thanks for publishing the plugin update.

sdk/rollup.config.js
Pauan: Yeah I'd rather it explicitly check for ENOENT.

sdk/rollup.config.js
Pauan: The only time `networks[name]` is undefined is if the user has a typo and is trying to import a module that doesn't exist, so we *want* it to throw.

sdk/rollup.config.js
Pauan: Same as above, we *want* it to throw if it's `undefined`, so I think it shouldn't be `async`.

.github/workflows/sdk.yml
Pauan: Since we're keeping this PR, could you please remove this, we can re-implement it in a future PR.

Pauan (changes requested):
Just a few small things and this should be good to go.

Pauan (approved):
LGTM, thanks!

iamalwaysuncomfortable (approved):

PR #1304, opened 2026-04-29
Route WASM proof-building network calls through configurable transport

marshacb:
## Motivation

The SDK's `transport` option (#1297) covers JS-level HTTP calls, but during proof building the WASM layer makes its own `fetch` calls via `reqwest` for state roots, block heights, and state paths. These bypass the configured transport entirely. For consumers with mTLS, custom certificates, or network policies, proving calls fail with timeout errors while regular SDK calls succeed.

### How it works

Two complementary mechanisms ensure all network calls during proof building route through the JS transport:

**CallbackQuery** (handles main execution trace):
A new `QueryTrait` implementation in Rust that holds JS callback functions. When WASM needs state data during `trace.prepare_async()`, it calls back into JS via the callbacks instead of making its own `reqwest` calls. The JS side fulfills these requests through the transport-aware `AleoNetworkClient`. This handles all programs including those with `DynamicRecord`/`DynamicFuture` inputs so no pre-computation of commitments is needed.

**Auto-constructed OfflineQuery** (handles fee execution):
For methods with separate fee execution (`buildExecutionTransaction`, `join`, `transfer`), the SDK pre-fetches state data via the JS transport and passes a populated `OfflineQuery` to WASM for the fee path. Fee execution is always `credits.aleo` (non-dynamic), so pre-fetching works reliably.

Applied to all 6 WASM execution methods: `buildExecutionTransaction`, `executeFunctionOffline`, `executeAuthorization`, `buildJoinTransaction`, `buildSplitTransaction`, `buildTransferTransaction`.

### What's in the PR

**Rust/WASM:**
- `callback_query.rs` - New `QueryTrait` implementation that delegates async methods to JS callbacks via `JsFuture`. Feature-gated behind `#[cfg(feature = "browser")]`.
- `computeQueryRequirements()` - Introspects program inputs to determine commitments needed for fee OfflineQuery construction. No network calls.
- All 6 execution methods accept `callback_query: Option<CallbackQuery>` with priority: CallbackQuery > OfflineQuery > direct fetch.

**TypeScript:**
- `buildCallbackQuery()` - Creates a `CallbackQuery` with three closures wrapping `networkClient.getLatestStateRoot()`, `getStatePaths()`, `getLatestHeight()`.
- `buildAutoOfflineQuery()` - Pre-fetches state data for fee execution via transport.
- `getLatestStateRoot()` and `getStatePaths()` - New transport-aware methods on `AleoNetworkClient`.
- `ensureInclusionKeys()` - Extracted helper replacing 6 duplicated blocks.

## Test Plan

```bash
yarn build:all
yarn test:sdk                
```

**Unit tests** (`sdk/tests/transport.test.ts`, 32 tests):

- `getLatestStateRoot` and `getStatePaths` route through transport
- CallbackQuery callbacks are invoked during `run()` with `proveExecution=true`
- Fee OfflineQuery pre-fetch skipped for dynamic record inputs
- Graceful fallback when state fetching fails (error logged, not propagated)

**E2E test** (`sdk/tests/transport-e2e.mjs`, CI job):

- Replaces `globalThis.fetch` with a throwing function (simulates mTLS)
- Calls `buildExecutionTransaction` with `transfer_public`
- TX builds successfully - zero WASM-side fetch calls
- All state data fetched via configured transport

**Manual validation** (performed locally):

- Built `transfer_private` TX with real on-chain record and blocked global fetch
- State path for record commitment fetched via transport
- TX built successfully - zero WASM bypass calls

---

> [!NOTE]
> **Medium Risk**
> Touches core proving/transaction-building paths by changing how WASM fetches state roots/heights/state paths (introducing callback-based queries and a unified query wrapper), so regressions could break proving or fee calculation across networks despite added tests.
> 
> **Overview**
> Ensures **WASM proof-building no longer bypasses the SDK’s configured `transport`** by introducing a JS-callback query path and plumbing it through all execution/transaction-building entrypoints.
> 
> On the TypeScript side, `AleoNetworkClient` tracks whether a custom transport is set and adds `getStatePaths`, while `ProgramManager` builds a `CallbackQuery` (wrapped in `QueryOption`) when a custom transport is present, and centralizes offline inclusion-key preloading via `ensureInclusionKeys`.
> 
> On the WASM/Rust side, adds `CallbackQuery` (delegates `QueryTrait` async methods to JS Promises) and a unified `QueryOption` wrapper, updates fee/height macros and all relevant manager methods to accept `Option<QueryOption>` instead of `OfflineQuery`, and exposes `computeQueryRequirements` to derive record commitments without network calls. Adds unit tests plus a new CI `transport-e2e` job that blocks global `fetch` and verifies proving still succeeds via transport.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8cf672ad1726256be49dafc406e8d94a49a7c307. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


wasm/src/programs/manager/execute.rs
```diff
@@ -56,13 +57,83 @@ use snarkvm_synthesizer::prelude::{InclusionVersion, execution_cost, execution_c
...
+    /// @param {RecordPlaintext | undefined} fee_record Optional fee record to include in commitment computation
+    /// @returns {Object} An object with { commitments: string[], hasDynamicInputs: boolean }
+    #[wasm_bindgen(js_name = "computeQueryRequirements")]
+    pub fn compute_query_requirements(
+        program: &str,
+        function_name: &str,
+        inputs: Array,
+        private_key: &PrivateKey,
+        fee_record: Option<RecordPlaintext>,
+    ) -> Result<JsValue, String> {
```
iamalwaysuncomfortable, line 88: Quick question, if we have this, why the `CallbackQuery`? Should we just contruct the `OfflineQuery` outside instead of using two QueryTypes?

iamalwaysuncomfortable (commented):
One question about this PR is, if we compute all the requirements outside, can we just construct the `OfflineQuery` outside `wasm`?

wasm/src/programs/manager/execute.rs
marshacb: For non-dynamic programs, yes. constructing the OfflineQuery outside works and we use that approach for the fee path. The issue is DynamicRecord inputs: computeQueryRequirements can't compute their commitments because the target program/record type is resolved at runtime during call.dynamic. So the OfflineQuery can't be fully populated before WASM runs.

CallbackQuery handles this because WASM discovers commitments during `trace.prepare_async()` and calls back to JS on-demand(avoiding the need for pre-computation). So OfflineQuery exists for what we can pre-fetch (fees) and CallbackQuery handles what we can't (dynamic dispatch).

wasm/src/programs/manager/transfer.rs
```diff
@@ -82,6 +83,7 @@ impl ProgramManager {
         fee_proving_key: Option<ProvingKey>,
         fee_verifying_key: Option<VerifyingKey>,
         offline_query: Option<OfflineQuery>,
+        callback_query: Option<CallbackQuery>,
```
iamalwaysuncomfortable: Can we maybe change this query to an `Option<Enum<of the currently supported Wasm query types we've made>>` so it stays one nullable parameter that can take any query type?

iamalwaysuncomfortable (approved):

iamalwaysuncomfortable (changes requested):
Let's try and modify the method signature to accept an enum of the supported queries. 

And sidenote, this particular feature likely won't need to be ported to mobile

wasm/src/programs/manager/execute.rs
```diff
@@ -224,9 +295,9 @@ impl ProgramManager {
         );
 
         log("Preparing inclusion proofs for execution");
-        let latest_height = if let Some(offline_query) = offline_query.as_ref() {
-            trace.prepare_async(offline_query).await.map_err(|err| err.to_string())?;
-            offline_query.current_block_height().map_err(|e| e.to_string())?
+        let latest_height = if let Some(ref query) = query {
+            trace.prepare_async(query).await.map_err(|err| err.to_string())?;
+            query.current_block_height_async().await.map_err(|e| e.to_string())?
```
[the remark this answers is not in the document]
marshacb: Acknowledged. The block height is fetched multiple times via callbacks (main trace + fee macros). Each is a lightweight API call, negligible vs the proving time. Could cache at the `QueryOption` level in a follow-up if the round-trips become a concern for high-latency environments.

wasm/src/programs/manager/execute.rs
```diff
@@ -56,13 +57,83 @@ use snarkvm_synthesizer::prelude::{InclusionVersion, execution_cost, execution_c
...
+        for commitment in &all_commitments {
+            js_commitments.push(&JsValue::from_str(commitment));
+        }
+        Reflect::set(&result, &"commitments".into(), &js_commitments.into())
+            .map_err(|_| "Failed to set commitments".to_string())?;
+        Reflect::set(&result, &"hasDynamicInputs".into(), &has_dynamic_inputs.into())
+            .map_err(|_| "Failed to set hasDynamicInputs".to_string())?;
+
+        Ok(result.into())
+    }
```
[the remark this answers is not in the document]
marshacb: Intentional. computeQueryRequirements is kept as public API for advanced consumers who need to inspect what commitments a program requires. The SDK no longer uses it internally but the WASM function and its tests are retained.

sdk/src/program-manager.ts
```diff
@@ -962,20 +1000,17 @@ class ProgramManager {
...
-        }
-
         // Auto-convert bare string inputs to field elements where the function expects field type.
         const preparedInputs = this.prepareInputs(program, functionName, inputs);
 
+        // Build the query: user-provided OfflineQuery for truly offline execution,
+        // or a CallbackQuery that delegates state fetching to JS transport.
+        const query = offlineQuery
+            ? QueryOption.offlineQuery(offlineQuery)
+            : QueryOption.callbackQuery(this.buildCallbackQuery());
```
iamalwaysuncomfortable: The callback query should really only be invoked when a custom transport is here.

sdk/src/program-manager.ts
```diff
@@ -1139,18 +1174,11 @@ class ProgramManager {
...
-                WasmProgramManager.loadInclusionProver(inclusionKeys[0])
-                this.inclusionKeysLoaded = true;
-                console.log("Successfully loaded inclusion key");
-            } catch {
-                logAndThrow(`Inclusion key bytes not loaded, please ensure the program manager is initialized with a KeyProvider that includes the inclusion key.`)
-            }
-        }
+        const query = offlineQuery
+            ? QueryOption.offlineQuery(offlineQuery)
+            : QueryOption.callbackQuery(this.buildCallbackQuery());
```
iamalwaysuncomfortable: Same here it should be invoked primarily when a custom transport is specified.

sdk/src/network-client.ts
```diff
@@ -992,6 +992,40 @@ class AleoNetworkClient {
...
 
+    /**
+     * Returns the latest state root of the network.
+     *
+     * @returns {Promise<string>} The latest state root as a string.
+     */
+    async getLatestStateRoot(): Promise<string> {
+        try {
+            this.ctx = { "X-ALEO-METHOD": "getLatestStateRoot" };
+            return String(await this.fetchData<string>("/stateRoot/latest"));
```
iamalwaysuncomfortable: The `String(...)` wrapper here is suspicious. If `fetchData` returns a string, it's a no-op; if it ever returns a parsed JSON value (object/number), this will silently produce something like `"[object Object]"` and the downstream `StateRoot::from_str` in WASM will fail with a confusing parse error rather than a clear network/shape error. Either remove the cast or add a comment explaining the expected response shape.

sdk/src/network-client.ts
```diff
@@ -992,6 +992,40 @@ class AleoNetworkClient {
...
+     * Returns state paths for the given record commitments.
+     *
+     * @param {string[]} commitments - Array of commitment field strings.
+     * @returns {Promise<string[]>} Array of state path strings corresponding to the commitments.
+     */
+    async getStatePaths(commitments: string[]): Promise<string[]> {
+        try {
+            this.ctx = { "X-ALEO-METHOD": "getStatePaths" };
+            const csv = commitments.join(",");
+            return await this.fetchData<string[]>(`/statePaths?commitments=${csv}`);
```
iamalwaysuncomfortable, line 1007: Commitments are joined with `,` and interpolated straight into the URL without encoding. Today commitments are field elements (URL-safe), so this works — but it's fragile to any future change in commitment shape and would fail open rather than throw. Suggest `encodeURIComponent(csv)` defensively.

iamalwaysuncomfortable (commented):
A few inline notes — none blocking on their own, but worth addressing before merge.

wasm/src/programs/callback_query.rs
```diff
@@ -0,0 +1,153 @@
...
+        1 => func.call1(&JsValue::NULL, &args[0]),
+        _ => {
+            let js_args = Array::new();
+            for arg in args {
+                js_args.push(arg);
+            }
+            func.apply(&JsValue::NULL, &js_args)
+        }
+    };
+
```
iamalwaysuncomfortable, line 75: `Promise::from(JsValue)` is an unchecked cast. If a caller hands in a synchronous JS function (e.g. `() => "value"`), this won't validate — `JsFuture` will resolve to the raw return value and downstream `as_string()` / `from_str` will produce an opaque error rather than a clear "callback must return a Promise" message. Low risk given `buildCallbackQuery()` always uses `async` arrows, but it's a footgun for anyone constructing `CallbackQuery` directly. A `result.dyn_into::<Promise>()` or a `Reflect::has(&result, "then")` check would surface a clearer error.

wasm/src/programs/manager/execute.rs
```diff
@@ -94,8 +165,8 @@ impl ProgramManager {
         proving_key: Option<ProvingKey>,
         verifying_key: Option<VerifyingKey>,
         url: Option<String>,
-        offline_query: Option<OfflineQuery>,
         edition: Option<u16>,
+        query: Option<QueryOption>,
```
iamalwaysuncomfortable, line 167: The query ordering is changed here, it should be in the position it was before so as not to break users of the direct wasm package (like shield core)

sdk/tests/transport-e2e.mjs
```diff
@@ -0,0 +1,125 @@
...
+        } else {
+            // Non-transport errors are OK (e.g., record already spent)
+            log("transfer_private — transport worked (non-transport error)", true, e.message?.slice(0, 100));
+        }
+    }
+
+    const t2StatePaths = transportCalls.filter(u => u.includes("statePaths"));
+    log("State paths fetched via transport", t2StatePaths.length > 0, `${t2StatePaths.length} call(s)`);
+} else {
+    console.log("⚠ Skipping record test — no unspent record found. Run transport-create-record.mjs first.");
```
iamalwaysuncomfortable: This silently skips the state-paths leg of the test when no unspent record is found. On CI — particularly if the funded test account is drained or the network call is rate-limited — the e2e job will pass while never exercising `getStatePaths`, defeating the test's purpose. Either fail the run when no record is available (so the regression surfaces), or seed/fund a record before this script runs.

marshacb: Good catch. Reworked to use a synthetic `RecordPlaintext` instead of scanning for real records. The record produces a valid commitment locally, WASM calls `getStatePaths` through transport (assertion passes), and the API's "not found" is caught as a non-transport error. Exercises the full path deterministically wiithout needing a funded account or skip flags.

wasm/src/programs/manager/execute.rs
marshacb: Great catch! Fixed

wasm/src/programs/callback_query.rs
marshacb: Agreed. replaced `Promise::from(promise)` with `promise.dyn_into::<Promise>()` which does a checked cast. Sync callbacks now get a clear error: "JS callback must return a Promise, got a non-thenable value".

sdk/src/network-client.ts
marshacb: Added `encodeURIComponent` on each commitment before joining. Defensive against future format changes.

sdk/src/network-client.ts
marshacb: `getLatestStateRoot` was a duplicate of the existing `getStateRoot()` with a String() wrapper tacked on. Removed it entirely and updated `buildCallbackQuery` to use `getStateRoot()`.

sdk/src/program-manager.ts
marshacb: Agreed! Added `hasCustomTransport` flag to `AleoNetworkClient`. `CallbackQuery` is now only constructed when a custom transport is configured. Without one, query is undefined and WASM falls back to its internal `SnapshotQuery` (pre-PR behavior). Added a test confirming `getStateRoot` is not invoked when no custom transport is set. (Same change applies to both call sites you flagged.)

sdk/tests/transport-e2e.mjs
```diff
@@ -0,0 +1,118 @@
...
+try {
+    const tx = await pm.buildExecutionTransaction({
+        programName: "credits.aleo",
+        functionName: "transfer_private",
+        inputs: [syntheticRecord.toString(), address, "1u64"],
+        priorityFee: 0,
+        privateFee: false,
+    });
+    log("transfer_private built with blocked global fetch", true);
+} catch (e) {
```
iamalwaysuncomfortable, line 100: One thing, can we try a dynamic dispatch function to see if it works in the case that the transport needs to be called in the case of dynamic dipsatch?

iamalwaysuncomfortable (approved):
Approved modulo testing it out with a dynamic dispatch function, thanks for the work on this!

PR #1330, opened 2026-05-28
Add test showcasing external signing for multiple requests for a single transaction

vicsn:
## Motivation

The high level workflow:
1. `sampleAuthorization` samples inputs for the full call graph
2. `computeExternalSigningInputs` is called for every request
3. Signing and gamma computation in MPC can happen. This is left out of the test.
4. `fromExternallySignedDataWithViewKey` is called for every request
5. `buildProvingRequestFromExecutionRequest` prepares a submission to the delegated prover.

### DIFF
--- a/sdk/src/program-manager.ts
+++ b/sdk/src/program-manager.ts
@@ -215,7 +215,7 @@ interface ProvingRequestOptions {
     unchecked?: boolean;
     edition?: number;
     useFeeMaster?: boolean;
-    executionRequest?: ExecutionRequest;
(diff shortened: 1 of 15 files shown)


iamalwaysuncomfortable (commented):
This deserves a discussion before this lands. There's a lot of open questions about the pipeline of sampling to --> consuming requests into a proof.

I assume some of those questions might be answered by examining the `SnarkVM rev` in this PR?

iamalwaysuncomfortable:
Closing in favor of updated work in #1355

PR #1331, opened 2026-05-29
feat: per-origin Kong session-affinity cookie jar in defaultTransport

bcgreijnlautier:
Linear: [PRO-541](https://linear.app/provable/issue/PRO-541/sdk-provablehqsdk-per-origin-cookie-jar-for-session-affinity-gateways)

## Motivation

Some Aleo backends sit behind a gateway (Kong in our case) that uses cookie-based session affinity: the server sets a `Set-Cookie: route=<UUID>` on the first response and expects it back on subsequent requests so per-session state (ephemeral pubkeys, JWT registrations) stays on the same upstream instance. Browsers and iOS NSURLSession persist cookies automatically; Node `fetch` and bare React Native do not — without a jar, those runtimes land on a random upstream per request, which has produced intermittent 400/401 failures on the delegated-proving and record-scanner paths.

This adds a small per-origin cookie jar to `defaultTransport`. Since all SDK HTTP traffic flows through `TransportFunction`, the fix covers `AleoNetworkClient` and `RecordScanner` automatically. Browser remains a no-op (`Cookie` is a forbidden request header).

## Design

- **Whitelisted cookie names only.** Only `route` is captured. Unrelated auth/session/CSRF cookies on the same origin are ignored, so other SDK clients in the same process can't inherit caller state. The whitelist is a `const ReadonlySet` at the top of the module — easy to extend if other affinity cookie names appear.
- **Module-scoped jar, per-origin keyed.** `Map<origin, Map<cookieName, value>>`. Shared across all `defaultTransport` callers in the same process — appropriate for routing cookies, and the whitelist contains the blast radius.
- **Defensive response reading.** Prefers `getSetCookie()` when the runtime exposes it (Node 22+ — handles multiple `Set-Cookie` headers correctly). Falls back to `get('set-cookie')` and splits the comma-joined string on `, ` followed by a cookie-name pattern, so multiple cookies are captured even on older runtimes; the regex deliberately skips the comma inside `Expires=Wed, 21 Oct 2015 …` date attributes. No-ops if the response has no `headers` field (existing SDK tests stub fetch with minimal response-like objects).
- **Runtime-global guards.** `Headers` and `Request` constructor checks use `typeof X !== 'undefined'`, and the Request-headers manipulation is wrapped in a try/catch so guarded-header runtimes don't fail the request.
- **`init.headers` priority (Fetch spec).** When the caller passes both a `Request` input and `init.headers`, `init.headers` REPLACES the Request's headers in the actual fetch — so the cookie must be attached there, not on the Request, or fetch would silently discard it. The transport checks `init.headers` first and falls back to merging into the Request's headers only when no init override is present.
- **Caller's Cookie wins.** `attachCookie` never overwrites an existing `Cookie` header passed by the caller in any of the three `HeadersInit` shapes (`Headers` instance, tuple array, plain object). The manual pubkey-cookie forwarding in `network-client.ts` continues to take precedence on the DPS path.
- **No mutation of caller state.** The transport never mutates the caller's `init`, `Headers` instance, or `Request` input. It shallow-clones `init` (`{ ...init, headers: … }`), clones any `Headers` via `new Headers(headersInit)` before `.set()`, and for Request input it copies the Request's existing headers into a fresh `Headers` and passes them via `init.headers`. This avoids a cross-origin leak where a caller reusing the same `Headers`/`init` would otherwise see origin A's routing cookie persist into the object and block origin B's correct cookie.

## Changes

| File | Net |
|---|---|
| `sdk/src/utils/utils.ts` | +238/-15 |
| `sdk/tests/cookie-jar.test.ts` | +431 (new) |

## Test plan

- [x] `cd sdk && rm -rf tmp && yarn exec rollup -c rollup.test.js && yarn exec mocha tmp/testnet/tests/cookie-jar.test.js --timeout 60000` — 14/14 passing:
  - captures the affinity Set-Cookie and attaches it on the next same-origin request
  - scopes cookies by origin (origin A does not leak to origin B)
  - does not overwrite a caller-provided Cookie header
  - ignores response objects without a `headers` field (back-compat with existing minimal fetch mocks)
  - uses `getSetCookie()` when the runtime exposes it (Node multi-Set-Cookie)
  - drops non-whitelisted cookies (no cross-client leakage of auth/session)
  - attaches jar cookie when the input is a `Request` object (via `init.headers`, not by mutating the Request)
  - preserves caller `Cookie` passed via a `Headers` instance
  - preserves caller `Cookie` passed via tuple-array headers
  - attaches the cookie to `init.headers` (not `input.headers`) when both a Request and `init.headers` are provided
  - parses multiple cookies from a comma-joined Set-Cookie fallback string
  - does not split on commas inside `Expires=…` date attributes (fallback path)
  - does not mutate a caller-supplied Headers instance, and does not leak cookies across origins when reused
  - does not add new properties to a caller-supplied init object
- [x] `yarn exec prettier --check sdk/src/utils/utils.ts sdk/tests/cookie-jar.test.ts` — clean
- [x] Broader `yarn test:sdk` — verified no new regressions beyond pre-existing failures (URL parsing in `new RecordScanner`, WASM `passStringToWasm0`) which reproduce on `mainnet` without this change

## Review history

| Round | Finding | Addressed in |
|---|---|---|
| Initial review | broader tests fail on `response.headers.get` over minimal mocks; prettier; runtime-global checks; cookie jar shared across clients; multi-`Set-Cookie` handling | `a4a0d23` |
| Cursor (medium) | `init.headers` overrides `Request.headers` per Fetch spec — cookie on `input.headers` would be discarded | `84aa1a1` |
| Cursor (low) | comma-joined `Set-Cookie` fallback only parses first cookie | `67fdb91` |
| Cursor (medium) | mutating caller's `init` / `Headers` / `Request` causes stale cross-origin cookies via shared-state reuse | `a2009dc` |

## Notes for reviewers

- This is a Node/RN-targeted fix; the browser code path is effectively a no-op.
- Downstream consumer (shield-mobile) currently carries this via `patches/@provablehq+sdk+0.10.2.patch`. Once this lands and is published, that patch file can be deleted.

---

> [!NOTE]
> **Medium Risk**
> Touches shared HTTP plumbing and cookie handling for proving/record-scanner paths; behavior is opt-in and whitelisted, but incorrect jar logic could still mis-route affinity-sensitive requests.
> 
> **Overview**
> Adds an **opt-in** `cookieAffinityTransport` that keeps a per-origin jar for Kong-style **`route`** affinity cookies on Node/React Native, where `fetch` does not persist cookies. Responses’ `Set-Cookie` values are stored (whitelist-only), replayed on later same-origin requests, and keyed by **`response.url`** after redirects. **`defaultTransport` is unchanged**; consumers pass `cookieAffinityTransport` via the existing `transport` option on network clients and related helpers.
> 
> The transport follows Fetch header rules (`init.headers` overrides `Request`), never overwrites caller `Cookie` headers, and avoids mutating caller `init`/`Headers`/`Request`. **`cookieAffinityTransport`** is re-exported from `browser.ts` and `utils/index.ts`. A dedicated **`cookie-jar.test.ts`** suite covers origin scoping, redirects, `getSetCookie()`, comma-joined fallbacks, and no-mutation guarantees.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1905b05f9e0b5f72a64ad9ee360b4d2e67046057. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/sdk/src/utils/utils.ts
+++ b/sdk/src/utils/utils.ts
@@ -178,13 +178,21 @@ function storeSetCookies(origin: string | null, cookies: string[]) {
     }
 }
 
+/*
+ * Returns a fresh HeadersInit with `cookie` attached when the input doesn't
+ * already carry one. NEVER mutates the caller's input — a caller reusing the
+ * same Headers/init across calls to different origins must not see origin A's
+ * jar cookie persist into the call for origin B.
+ */
 function attachCookie(
     headersInit: HeadersInit | undefined,
     cookie: string,
 ): HeadersInit {
     if (isHeadersLike(headersInit)) {
-        if (!headersInit.has("cookie")) headersInit.set("cookie", cookie);
-        return headersInit;
+        if (headersInit.has("cookie")) return headersInit;
+        const cloned = new Headers(headersInit);
+        cloned.set("cookie", cookie);
+        return cloned;
     }
     if (Array.isArray(headersInit)) {
         const hasCookie = headersInit.some(
@@ -222,23 +230,32 @@ export const defaultTransport: TransportFunction = async (input, init) => {
     const origin = originOf(input as URL | string | Request);
     const cookie = cookieHeaderFor(origin);
     if (cookie) {
-        // Per the Fetch spec, an explicit `init.headers` REPLACES the
-        // Request input's headers when fetch builds the actual request.
-        // So if init.headers is provided, that's the only place worth
-        // attaching the cookie — attaching to input.headers would be
-        // silently discarded.
+        // Always operate on shallow clones so callers reusing the same
+        // `init` / `Headers` / `Request` across origins don't end up with
+        // origin A's jar cookie persisting into the object and blocking
+        // origin B's correct cookie.
         if (init?.headers !== undefined && init.headers !== null) {
-            init.headers = attachCookie(init.headers, cookie);
+            // init.headers, when explicitly set, REPLACES the Request's
+            // headers per the Fetch spec — attach there.
+            init = { ...init, headers: attachCookie(init.headers, cookie) };
         } else if (isRequestLike(input)) {
+            // Merge the Request's existing headers with the cookie and
+            // pass them via `init.headers` instead of mutating the
+            // Request itself. (Fetch spec: an explicit `init.headers`
+            // replaces Request.headers entirely, so we copy the
+            // originals into the merged set first.)
             try {
-                if (!input.headers.has("cookie"))
-                    input.headers.set("cookie", cookie);
+                const merged = new Headers(input.headers);
+                if (!merged.has("cookie")) {
+                    merged.set("cookie", cookie);
+                    init = { ...(init ?? {}), headers: merged };
+                }
             } catch {
-                // Some runtimes treat Request.headers as immutable; skip
-                // rather than failing the request.
+                // Some runtimes restrict Headers construction from a
+                // Request; skip rather than failing the request.
             }
         } else if (init) {
-            init.headers = attachCookie(undefined, cookie);
+            init = { ...init, headers: attachCookie(undefined, cookie) };
         } else {
             init = { headers: attachCookie(undefined, cookie) };
         }
--- a/sdk/src/utils/utils.ts
+++ b/sdk/src/utils/utils.ts
@@ -3,11 +3,17 @@ import { logger } from "./logger.js";
 function detectBrowser() {
     const userAgent = navigator.userAgent;
 
-    if (/chrome|crios|crmo/i.test(userAgent) && !/edge|edg|opr/i.test(userAgent)) {
+    if (
+        /chrome|crios|crmo/i.test(userAgent) &&
+        !/edge|edg|opr/i.test(userAgent)
+    ) {
         return "chrome";
     } else if (/firefox|fxios/i.test(userAgent)) {
         return "firefox";
-    } else if (/safari/i.test(userAgent) && !/chrome|crios|crmo|android/i.test(userAgent)) {
+    } else if (
+        /safari/i.test(userAgent) &&
+        !/chrome|crios|crmo|android/i.test(userAgent)
+    ) {
         return "safari";
     } else if (/edg/i.test(userAgent)) {
         return "edge";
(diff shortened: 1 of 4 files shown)


marshacb (approved):
LGTM. One non-blocking question: It seems the jar holds cookies indefinitely and ignores Max-Age / Expires. Does Kong always issue a fresh Set-Cookie on rotation that overwrites the stored one, or is there a scenario where we'd keep replaying a stale `route` past server-side expiry without ever getting a new one? Worth a TODO, or non-issue in practice?

(Also confirmed the two CI reds reproduce on mainnet. Unrelated to this PR.)

bcgreijnlautier:
>Does Kong always issue a fresh Set-Cookie on rotation that overwrites the stored one, or is there a scenario where we'd keep replaying a stale `route` past server-side expiry without ever getting a new one? Worth a TODO, or non-issue in practice?
> 

AFAIK it's a non-issue in practice. Kong re-issues `Set-Cookie: route=<uuid>` on every response from the sticky upstream, so the jar is continuously refreshing. A stale entry lasts for one request at most before the next response overwrites it or rotates it (e.g. Kong load balances and emits a new UUID). The jar is also module-scoped, so process lifetime bounds anything stale.

sdk/src/utils/utils.ts
```diff
@@ -47,8 +54,216 @@ export function logAndThrow(message: string): never {
...
+ * jar on top. Responses' `Set-Cookie` headers are captured (filtered by
+ * `AFFINITY_COOKIE_NAMES`) and replayed as a `Cookie` header on subsequent
+ * same-origin requests, which is required for backends that use cookie-based
+ * session affinity (see `cookieJar` above).
+ *
+ * A caller-supplied `Cookie` header is never overwritten — `network-client.ts`
+ * forwards the pubkey-response cookie manually onto delegated-prove requests
+ * for Node compatibility, and that path takes precedence over the jar.
+ */
+export const defaultTransport: TransportFunction = async (input, init) => {
```
iamalwaysuncomfortable: The side effects of this aren't completely known and this may have side effects we're not fully aware of on existing callers. 

We need to restore the `defaultTransport` here to what it was and rename this. Any users of it should be able to replace things manually.


```suggestion
export const cookieAffinityTransport: TransportFunction = async (input, init) => {
```

You'll be able to set the transport yourself with this in the mobile SDK (or anywhere else). I'd say, to land this:
1. Change the name of this, and ensure it's re-exported within browser.ts so people can import it from the top level SDK imports.
2. Update the tests to this shape.

iamalwaysuncomfortable (changes requested):

iamalwaysuncomfortable (approved):

PR #1381, opened 2026-07-29
Skip fee estimation for fee-master proving requests

kpandl:
## Motivation

`buildProvingRequest` always estimated the execution fee and derived the execution ID after building the main authorization. It then discarded both results when the request used a fee master, or when the selected `credits.aleo` function intentionally had no fee authorization.

This moves fee estimation and execution-ID derivation into the branch that actually builds a fee authorization. The existing `useFeeMaster` option remains the public signal, so this introduces no new API or conflicting skip flag.

In a local benchmark of the representative testnet swap transaction `at174kyef8hwfyv7v2tk67x3truvcc4zm2ldh0rwmdqn8an48l08u9q27xwsw`, an unchecked fee-master proving request with a warm process dropped from roughly 1.41–1.48 s to 0.252–0.267 s. The removed fee estimate accounted for roughly 1.15–1.19 s. Authorization shape, transition count, absence of a fee authorization, and serialization round-trip remained unchanged.

## Test Plan

- `yarn build:wasm`
- `yarn build:sdk`
- `cargo fmt --manifest-path wasm/Cargo.toml -- --check`
- `cargo check --manifest-path wasm/Cargo.toml --target wasm32-unknown-unknown --features testnet`
- Targeted WASM test for fee-authorization requirements — 1 passed

## Related PRs

- #1382 — prepared-program/prewarm API
- ProvableHQ/shield-core#256 — request unchecked authorization for delegated proving

---

> [!NOTE]
> **Medium Risk**
> Behavior change is limited to when fee work runs, but incorrect branching would break fee authorization for paying users; logic is covered by new unit tests and existing fee-master integration tests.
> 
> **Overview**
> **`buildProvingRequest`** no longer runs **`estimate_fee_for_authorization`** or derives the execution ID on every call. Those steps run only when the request will actually attach a fee authorization.
> 
> Fee-master requests (`use_fee_master`) and the existing **`credits.aleo`** exceptions (`split`, `upgrade`, `fee_public`, `fee_private`) skip that work entirely. The public API is unchanged: `base_fee_credits` is still accepted but ignored when no fee auth is built; the base fee is estimated inside the fee-auth branch as before.
> 
> A small **`requires_fee_authorization`** helper centralizes that decision, with a WASM unit test covering fee-master, credits exceptions, and normal programs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6aa0d2ad61142a90d89b1bda30d7dab7b3707124. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/wasm/src/programs/manager/proving_request.rs
+++ b/wasm/src/programs/manager/proving_request.rs
@@ -31,6 +31,12 @@ use crate::{
 use js_sys::{Array, Object};
 use std::str::FromStr;
 
+fn requires_fee_authorization(program_id: &str, function_name: &str, use_fee_master: bool) -> bool {
+    !use_fee_master
+        && !(program_id == "credits.aleo"
+            && matches!(function_name, "split" | "upgrade" | "fee_public" | "fee_private"))
+}
+
 #[wasm_bindgen]
 impl ProgramManager {
     /// Create a `ProvingRequest` object. This object creates authorizations for the top level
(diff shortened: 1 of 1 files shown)


iamalwaysuncomfortable:
## Review

### Overview

Moves fee estimation (`estimate_fee_for_authorization`) and execution-ID derivation inside the branch of `buildProvingRequest` that actually builds a fee authorization, so fee-master requests and the `credits.aleo` exceptions (`split`/`upgrade`/`fee_public`/`fee_private`) skip both entirely. The skip decision is centralized in a new `requires_fee_authorization` helper with a unit test. Given the benchmark (~1.15–1.19 s saved per fee-master request), this is a clear win for the delegated-proving path.

### Correctness

- ✅ **The helper is an exact negation of the old condition.** The old inline expression was `(program_id == "credits.aleo") && (split || upgrade || fee_public || fee_private) || use_fee_master`, which by Rust precedence parses as `(credits && exceptions) || use_fee_master`. The new `!use_fee_master && !(credits && exceptions)` is its De Morgan complement, so branch selection is unchanged — including the slightly surprising precedence of the original.
- ✅ **The moved computations were genuinely dead on the skip paths.** In the old code, `base_fee_microcredits` and `execution_id` fed only `authorize_fee!`; nothing else in the function (or the returned `ProvingRequest`) consumed them. The removed `program_native` clone was only used for the `credits.aleo` check, which now reads `resolved.program().id()` before `process_mut()` — same value, and it drops a full `Program` clone as a bonus.
- ✅ For the paying path, the order estimate → execution ID → `authorize_fee!` is preserved, so fee-authorization output is byte-for-byte the same.
- ℹ️ **One deliberate behavior change worth being aware of:** errors from `estimate_fee_for_authorization` or `to_execution_id()` no longer surface on fee-master / exception requests. A request that previously failed at estimation (e.g. an import-resolution edge case) will now succeed and reach the prover. That's arguably the point — the estimate was irrelevant to those requests — but it does shift where such an error would first appear (client → remote prover).

### Suggestions (non-blocking)

1. **`base_fee_credits` doc is now misleading.** The `@param base_fee_credits The base fee to be paid for the authorization` JSDoc still implies the value is used, but it's fully ignored (`let _ = base_fee_credits;`). It was already effectively ignored before this PR (the old conversion result was underscore-prefixed and unused), so no behavior change — but since this PR touches the area, consider updating the doc to say the parameter is retained for signature compatibility and the base fee is always estimated.
2. Micro-nit: `requires_fee_authorization` takes `&str`, forcing a `to_string()` allocation at the call site. Comparing `ProgramID`/`Identifier` natives directly would avoid it — negligible cost either way, fine as is.

### Tests

- The new `fee_authorization_requirement_matches_request_type` unit test covers all three helper branches (fee-master, credits exceptions, normal program).
- Existing integration tests already pin both runtime paths: `test_nested_proving_request` (fee authorization present, serialization round-trip) and `test_proving_request_with_use_fee_master_has_no_fee_authorization` (no fee auth). Good coverage for a one-function change.

### Verdict

**LGTM.** Focused diff, semantics-preserving refactor with a measured perf win, and the one intentional behavior change (skipping estimation errors for fee-master requests) is consistent with the PR's motivation. Only the stale `base_fee_credits` doc is worth a follow-up touch.

iamalwaysuncomfortable (approved):

